PR #1, opened 2017-02-08
Feature/init

malaffoon:
Initial checkin ...


build.gradle
```diff
@@ -0,0 +1,69 @@
...
+//        compile 'org.slf4j:slf4j-api'
+
+        // every app is a microservice with actuator end-points
+        compile('org.springframework.boot:spring-boot-starter-actuator')
+
+        // every app is a spring cloud stream participant
+        compile('org.springframework.cloud:spring-cloud-stream')
+        compile('org.springframework.cloud:spring-cloud-starter-stream-rabbit')
+
+//      compile("org.springframework.boot:spring-boot-starter-web")
```
malaffoon: i've left these in as comments just until i get through the initial rush of functional changes

exam-service/src/main/java/org/rdw/ingest/DefaultExamService.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import rdw.model.TDSReport;
+import rdw.model.Test;
+
+import java.util.Optional;
+
+/**
+ * Default implementation of ExamService
```
malaffoon: yeah, okay, this is pretty useless. I'll buff up the class comments as more real impl is filled in

exam-service/src/main/java/org/rdw/ingest/ExamController.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import rdw.model.TDSReport;
+
+/**
+ * Controller for exam / testresults end-points
+ */
+@RestController
+@RequestMapping({ "/exams", "/testresults" })
```
malaffoon: Thoughts? Per Chris "we don't want to use Test in our java code 'cause it's confusing." But i don't think we should propagate that to our end users. So i want the end-point to be "/testresults". But the it is kind of weird that the controller is ExamController. Meh.

exam-service/src/test/java/org/rdw/ingest/DefaultExamServiceTests.java
```diff
@@ -0,0 +1,39 @@
+package org.rdw.ingest;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.verify;
+
+public class DefaultExamServiceTests {
```
malaffoon: i noticed that the project generator called the test suite FubarTests; i have traditionally not pluralized that so it would be FubarTest. Anybody care enough for me to change this?

agorina: I do not think it is worth the effort. I would leave it as is.

rcordeau: I'm traditionally wouldn't have pluralized it either, but I don't care enough.

esotrope:
Are gradle artifacts other than build.gradle really necessary for checkin? seems like something that belongs in .gitignore

rcordeau:
All of the files currently checked in are necessary for gradle except for maybe the .bat file and wrapper.jar (that will get downloaded automatically when you run gradlew)

exam-service/src/main/java/org/rdw/ingest/ExamController.java
rcordeau: /examresults?

malaffoon:
@rcordeau is that true (about gradle-wrapper.jar)? When i tried without it it didn't work. And i don't see anywhere in the script where it tries to download the jar? I thought the whole point of having the jar in the project is to guarantee the same version of gradle.

PR #4, opened 2017-02-09
Initial set up of the exam processor.

agorina:


exam-processor/src/main/resources/application.yml
```diff
@@ -0,0 +1,10 @@
+
+spring:
+  cloud:
+    stream:
+      bindings:
+        input:
+          destination: exam
+        output:
```
malaffoon: It could be that this should be a sink instead of a processor. We should discuss.
If it is a processor, the output would be something that indicates a new exam is available to migrate.

exam-processor/src/main/java/org/rdw/processor/ExamProcessorApplication.java
```diff
@@ -0,0 +1,17 @@
+package org.rdw.processor;
```
malaffoon: Should this be org.rdw.processor or org.rdw.ingest? Sure, the artifact may be exam-processor but should the package change? Open to opinion, and i don't feel too strongly about it.

exam-processor/src/main/resources/application.yml
agorina: Makes sense. Let's talk tomorrow.

exam-processor/src/main/java/org/rdw/processor/ExamProcessorApplication.java
agorina: What about: org.rdw.ingest.processor?

malaffoon: Ok.

PR #6, opened 2017-02-09
docker

malaffoon:
Add docker build task. This is WIP but i want to test some stuff with merging branches. And i'm hopeful Rob will be able help with this gradle stuff. I dragged Brett in to look at gradle as well, any comments welcome!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,35 +1,44 @@
 buildscript {
     repositories {
+        jcenter()
         mavenLocal()
         mavenCentral()
     }
     dependencies {
+        classpath("com.bmuschko:gradle-docker-plugin:3.0.5")
         classpath("io.spring.gradle:dependency-management-plugin:1.0.0.RELEASE")
         classpath("org.springframework.boot:spring-boot-gradle-plugin:1.5.1.RELEASE")
     }
 }
 
 apply plugin: 'io.spring.dependency-management'
+apply plugin: 'com.bmuschko.docker-remote-api'
+
+import com.bmuschko.gradle.docker.tasks.image.Dockerfile
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+import com.bmuschko.gradle.docker.tasks.image.DockerPushImage
+import com.bmuschko.gradle.docker.DockerRegistryCredentials
+
 
 allprojects {
     group = 'org.rdw.ingest'
     version = '0.0.1-SNAPSHOT'
 
-    apply plugin: 'java'
-    apply plugin: 'maven'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1,35 +1,44 @@
...
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+import com.bmuschko.gradle.docker.tasks.image.DockerPushImage
+import com.bmuschko.gradle.docker.DockerRegistryCredentials
+
 
 allprojects {
     group = 'org.rdw.ingest'
     version = '0.0.1-SNAPSHOT'
 
-    apply plugin: 'java'
```
malaffoon: These didn't belong at the rootProject which is just a container parent.

build.gradle
```diff
@@ -1,35 +1,44 @@
 buildscript {
     repositories {
+        jcenter()
```
malaffoon: This may not be necessary, i pulled it in for another plugin i'm no longer using.

build.gradle
```diff
@@ -1,35 +1,44 @@
...
         mavenLocal()
         mavenCentral()
```
rcordeau: You can likely remove mavenCentral since you now have jcenter. jcenter seems to be a wrapper around mavenCentral anyway.

build.gradle
```diff
@@ -61,4 +70,33 @@ subprojects {
...
+        entryPoint("java", "-jar", "${jar.archivePath.name}")
+    }
+
+    // task to build the docker image
+    task buildImage(type:DockerBuildImage, dependsOn:[build, createDockerfile]) {
+        dockerFile = createDockerfile.destFile
+        inputDir = project.file('build/docker/')
+        tag = "fwsbac/rdw-${project.name}"
+        doFirst {
+            copy {
```
malaffoon: is copy smart? Will it not copy if the file hasn't changed?

rcordeau: If which file hasn't changed?

malaffoon: If the "from" file hasn't changed. 
If i run buildImage over and over, is it copying the jar file every time even though it hasn't changed. I guess i can play with it, just wondering if anybody knew off the top of their head.

rcordeau: I am pretty sure that it won't generate a new dockerFile, but it will copy it even if it hasn't changed.

malaffoon: Okay. I experimented with having another task, copyJarToDockerContext(type:Copy), to try to avoid that. But this is cleaner to read and, honestly, copying the file is not the long pole for this task.

PR #7, opened 2017-02-10
add RdwImport and change end-points, service, etc.

malaffoon:
and some other stuff that i will call out in comments

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,14 +104,4 @@ subprojects {
     task pushImage(type:Exec, dependsOn:[buildImage]) {
         commandLine("docker", "push", "fwsbac/rdw-${project.name}")
     }
-
-    // to use this, you must set docker hub credentials in ~/.gradle/gradle_properties:
-    task pushImageByApi(type:DockerPushImage, dependsOn:[buildImage]) {
-        imageName = "fwsbac/rdw-${project.name}"
-        registryCredentials = new DockerRegistryCredentials(
-                username: dockerHubUsername,
-                password: dockerHubPassword,
-                email: dockerHubEmail
-        )
-    }
 }
--- a/exam-service/src/main/java/org/rdw/ingest/model/ImportStatus.java
+++ b/exam-service/src/main/java/org/rdw/ingest/model/ImportStatus.java
@@ -1,0 +1,27 @@
+package org.rdw.ingest.model;
+
+/**
+ * Status for imports
+ */
+public enum ImportStatus {
+    INVALID(-1),
+    ACCEPTED(0),
+    PROCESSED(1);
+
+    private final int value;
+
+    ImportStatus(final int value) {
+        this.value = value;
+    }
+
+    public int getValue() {
+        return value;
+    }
+
+    public ImportStatus fromValue(final int value) {
+        for (final ImportStatus status : ImportStatus.values()) {
+            if (value == status.value) return status;
+        }
+        throw new IllegalArgumentException("Unknown ImportStatus value.");
+    }
+}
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -104,14 +104,4 @@ subprojects {
     task pushImage(type:Exec, dependsOn:[buildImage]) {
         commandLine("docker", "push", "fwsbac/rdw-${project.name}")
     }
-
-    // to use this, you must set docker hub credentials in ~/.gradle/gradle_properties:
-    task pushImageByApi(type:DockerPushImage, dependsOn:[buildImage]) {
```
malaffoon: not really using this, just an experiment ... didn't realize gradle would evaluate the properties during compilation of the script

exam-service/src/main/java/org/rdw/ingest/model/ImportStatus.java
```diff
@@ -0,0 +1,27 @@
+package org.rdw.ingest.model;
+
+/**
+ * Status for imports
+ */
+public enum ImportStatus {
+    INVALID(-1),
```
malaffoon: these int values aren't really needed yet ... i was thinking these would map to the value stored in the database

exam-service/src/main/java/org/rdw/ingest/model/RdwImport.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public Instant getCreated() {
+        return created;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
malaffoon: i just did my own builder stuff, not using Immutables

exam-service/src/main/java/org/rdw/ingest/web/ExamController.java
```diff
@@ -30,16 +30,15 @@ public ExamController(@NotNull final ExamService service) {
         this.service = checkNotNull(service);
     }
 
-    @RequestMapping(method = RequestMethod.POST,
-                    consumes = MediaType.APPLICATION_XML_VALUE,
```
malaffoon: remove all restrictions on what this will consume

exam-service/src/main/java/org/rdw/ingest/model/ImportStatus.java
agorina: Yes, agree. We should chat about it. I have similar issue with grade and other codified fields.

exam-service/src/main/java/org/rdw/ingest/model/RdwImport.java
agorina: Well, after spending a day fighting Immutable - I am all for doing it 'old ways'

chdietz: I thought we agreed to not use any Immutable or Lomback for now?

agorina: I definitely agree now:)

 I was told that we agreed on Immutable...

exam-service/src/main/java/org/rdw/ingest/model/RdwImport.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.time.Instant;
+import java.util.UUID;
+
+import static java.time.Instant.now;
+
+/**
+ * An import request.
+ */
+public class RdwImport {
```
chdietz: Not sure I understand the name of this model object?  Is it used for all RDW imports?

malaffoon: Well, i would love to make it just Import but java doesn't like that. So, yeah, in a way, it represents any RDW import request. For example, exam, assessment package, student reg. We may want to add polymorphism and make this an abstract base class at some point but this works for now.

exam-service/src/main/java/org/rdw/ingest/model/RdwImport.java
esotrope: Let me know if you guys want help with Immutables - it's really nice once you spend the hour to figure it out.

PR #8, opened 2017-02-10
Adding initial support for MySQL access.

agorina:


exam-processor/src/main/java/org/rdw/ingest/processor/repositories/impl/AssessmentRepositoryImpl.java
```diff
@@ -0,0 +1,67 @@
...
+                .addValue("name", assessment.getName())
+                .addValue("label", assessment.getLabel())
+                .addValue("version", assessment.getVersion());
+
+        jdbcTemplate.update(sql, parameterSource, keyHolder);
+        return Assessment.builder().copy(assessment).id(keyHolder.getKey().longValue()).build();
+
+    }
+
+    private static class AssessmentRowMapper implements RowMapper<Assessment> {
```
malaffoon: ugh

rcordeau: What's the matter? Don't like RowMapper?

exam-processor/src/test/resources/application.yml
```diff
@@ -0,0 +1,9 @@
+
+spring:
+  datasource:
+    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false
```
malaffoon: shouldn't this have a different schema or something?  if it is exactly the same as non-test, why have a separate file?

exam-processor/src/main/java/org/rdw/ingest/processor/repositories/impl/AssessmentRepositoryImpl.java
agorina: sorry, I close the request. I will re-open it though. Do you have a better alternative?

malaffoon: previous life and jpa ... not sure i'm going to be happy with our decision to not use it

exam-processor/src/test/resources/application.yml
agorina: Yes, later - just wanted to keep it simple

exam-processor/src/main/java/org/rdw/ingest/processor/repositories/impl/AssessmentRepositoryImpl.java
malaffoon: no @agorina, just venting

PR #10, opened 2017-02-14
add basic auth using oauth as the auth provider

malaffoon:
To clarify this: our client (ETS) will use basic auth while we use OpenAM IDP as the authority for user credentials. This is a rudimentary implementation just to make sure it works as expected. 

After a chat with ETS we will decide whether to continue down this route or switch to full oauth2 (where they'd have to know about our OpenAM IDP and get a token).


README.md
```diff
@@ -50,7 +50,7 @@ to catch the exams and log something about them.
 ```bash
 docker run -d -p :8080:8080 --name exam-service --link rabbitmq:rabbitmq fwsbac/rdw-exam-service --spring.rabbitmq.host=rabbitmq
 docker run -d -p :8090:8080 --name exam-log --link rabbitmq:rabbitmq springcloudstream/log-sink-rabbit --spring.rabbitmq.host=rabbitmq \
-  --spring.cloud.stream.bindings.input.destination=exams \
+  --spring.cloud.stream.bindings.input.destination=exam \
```
malaffoon: changed this to match what is in the exam-processor; we need to decide on naming conventions for our message queues at some point

PR #12, opened 2017-02-14
use TenancyChain from utils

malaffoon:
fix security hole
apply oauth to only exams end-points

### DIFF
--- a/exam-service/src/main/java/org/rdw/ingest/web/OAuth2AuthenticationProvider.java
+++ b/exam-service/src/main/java/org/rdw/ingest/web/OAuth2AuthenticationProvider.java
@@ -21,7 +21,7 @@ import org.springframework.security.oauth2.client.token.grant.password.ResourceO
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
 import org.springframework.security.oauth2.config.annotation.web.configuration.EnableOAuth2Client;
 import org.springframework.stereotype.Component;
-import org.springframework.util.StringUtils;
+import rdw.utils.TenancyChain;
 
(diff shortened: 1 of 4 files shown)


exam-service/src/main/java/org/rdw/ingest/web/OAuth2AuthenticationProvider.java
```diff
@@ -59,32 +59,33 @@ public Authentication authenticate(final Authentication authentication) throws A
         if (!supports(authentication.getClass())) return authentication;
 
         final UsernamePasswordAuthenticationToken auth = (UsernamePasswordAuthenticationToken) authentication;
+        final String username = auth.getName();
+        final String password = (String) auth.getCredentials();
 
-        UserDetails user = userCache.getIfPresent(auth.getName());
+        UserDetails user = userCache.getIfPresent(username+password);
```
malaffoon: Before this change ...
If a client provided a valid username and password and was authenticated, that username could then be used without the password, at least until the cache expired the entry.

exam-service/src/main/java/org/rdw/ingest/web/SecurityConfig.java
```diff
@@ -40,8 +40,8 @@ protected void configure(final HttpSecurity http) throws Exception {
                 .disable()
             .authorizeRequests()
                 // TODO - worry about actuator end-points
-                .anyRequest()
```
malaffoon: This was eclipsing the Spring Boot access control on the actuator end-points. So you had to be authenticated to GET ./health.

PR #13, opened 2017-02-15
auth tweaks to preserve tenancy chain and content type

malaffoon:
This catches the content type and authenticated user so additional message headers can be set for exam-payload

### DIFF
--- a/exam-processor/src/main/java/org/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -17,6 +17,7 @@ public class ExamProcessorConfiguration {
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         final String payload = (String) message.getPayload();
-        logger.info(accessor.getContent() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
+        logger.info("received " + accessor.getContent() + " from " + accessor.getUserLogin());
+        logger.info(accessor.getContentType() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
     }
 }
--- a/exam-service/src/main/java/org/rdw/ingest/auth/OAuth2AuthenticationProvider.java
+++ b/exam-service/src/main/java/org/rdw/ingest/auth/OAuth2AuthenticationProvider.java
@@ -27,6 +27,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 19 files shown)


exam-processor/src/main/java/org/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -17,6 +17,7 @@
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         final String payload = (String) message.getPayload();
-        logger.info(accessor.getContent() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
+        logger.info("received " + accessor.getContent() + " from " + accessor.getUserLogin());
```
malaffoon: just more logging to prove the stuff is getting to the next step in processing

exam-service/src/main/java/org/rdw/ingest/service/ExamService.java
```diff
@@ -10,7 +10,7 @@
  */
 public interface ExamService {
 
-    RdwImport importExam(String xml, String batchId);
+    RdwImport importExam(RdwUser user, String payload, String contentType, String batchId);
```
malaffoon: i guess i owe some javadocs here

exam-service/src/main/java/org/rdw/ingest/auth/OAuth2AuthenticationProvider.java
```diff
@@ -80,11 +76,11 @@ public Authentication authenticate(final Authentication authentication) throws A
                 // extract roles from tenancy chain
                 // TODO - should this be client-specific? i.e. chain.hasRoleForClient(DataLoadRole)
                 final List<GrantedAuthority> authorities = new ArrayList<>();
-                if (chain.hasRole(DataLoadRole)) {
-                    authorities.add(new SimpleGrantedAuthority(DataLoadRole));
+                if (chain.hasRole(SecurityConfig.DataLoadRole)) {
```
esotrope: Hey Mark, I'm interested to know the design demand for creating a custom auth provider for oauth2. Is there not an out of the box auth provider for oauth2? I ask because with the other auth methods generally spring supplies the auth provider and the developer adds a custom UserDetailsService to map the user details to a user impl.

malaffoon: The problem with the UserDetailsService approach (which i have used in the past) is that it requires that a user can be loaded with just the username. In oauth2 that is not possible: you can only get authorization information when you have the username and password (along with your app's credentials). Of course, this may indicate that i am using oauth2 improperly here; i am basically using it as a directory service. Perhaps there are other interfaces that OpenAM make available that i could use instead to get the user details?

PR #14, opened 2017-02-17
A first draft of the exam processor model and flow.

agorina:
There are a lot of TODOs that I will be filling in later, including unit tests.


agorina:
Somethings does not look right here. Will create a new request.

PR #26, opened 2017-02-28
Import persistence

malaffoon:
Add mysql persistence of import.
Flesh out some features.


exam-processor/src/main/resources/application.yml
```diff
@@ -12,4 +12,5 @@ spring:
     testWhileIdle: true
     validationQuery: SELECT 1
     driverClassName: com.mysql.jdbc.Driver
+    initialize: false
```
malaffoon: The default value of `true`; it wasn't doing anything because there are no scripts in the classpath, but it still felt wrong to leave the default.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/model/RdwImport.java
```diff
@@ -1,37 +1,57 @@
...
  */
 public class RdwImport {
-    private String id;
-    private String content;
+    private Long id;
+    private ImportContent content;
+    private String digest;
     private ImportStatus status;
-    private String batchId;
+    private String batch;
```
malaffoon: For now, batch is just a string that is applied to an import. Perhaps in the future there will be a normalized `batch` entity but we don't need it yet.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/model/RdwImport.java
```diff
@@ -1,37 +1,57 @@
...
 
 /**
  * An import request.
  */
 public class RdwImport {
-    private String id;
-    private String content;
+    private Long id;
+    private ImportContent content;
+    private String digest;
```
malaffoon: `digest` is used to avoid reprocessing the exact same payload.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/DataSourceConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+import javax.sql.DataSource;
+
+@Configuration
+@EnableTransactionManagement
+public class DataSourceConfiguration {
```
malaffoon: This change set includes a few files that are identical between exam-service and exam-processor. At some point we will reach critical mass and will probably want to break out shared components.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepository.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.ingest.repository;
+
+import org.opentestsystem.rdw.ingest.model.RdwImport;
+
+/**
+ * A repository that manages {@link RdwImport}s
+ */
+public interface RdwImportRepository {
```
malaffoon: sigh ... we really don't want to use spring data jpa?

agorina: I actually learned to like the jpaTemplates. Sorry...

PR #30, opened 2017-02-28
clean up some TODO's

malaffoon:
load SQL from YAML file

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.ingest.processor;
 
+import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Import;
@@ -8,7 +9,7 @@ import org.springframework.context.annotation.Import;
  * App for exam processor
  */
 @SpringBootApplication
-@Import(ExamProcessorConfiguration.class)
+@Import({ ExamProcessorConfiguration.class, YamlPropertiesConfigurator.class })
 public class ExamProcessorApplication {
 
     public static void main(String[] args) {
--- a/exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
+++ b/exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
@@ -76,7 +76,6 @@ public class OAuth2AuthenticationProvider implements AuthenticationProvider {
                 }
 
                 // extract roles from tenancy chain
-                // TODO - should this be client-specific? i.e. chain.hasRoleForClient(DataLoadRole)
                 final List<GrantedAuthority> authorities = new ArrayList<>();
                 if (chain.hasRole(DataLoadRole)) {
                     authorities.add(new SimpleGrantedAuthority(DataLoadRole));
--- a/exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
+++ b/exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
@@ -4,6 +4,7 @@ import org.opentestsystem.rdw.ingest.model.ImportContent;
 import org.opentestsystem.rdw.ingest.model.ImportStatus;
 import org.opentestsystem.rdw.ingest.model.RdwImport;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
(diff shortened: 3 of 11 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
```diff
@@ -8,7 +9,7 @@
  * App for exam processor
  */
 @SpringBootApplication
-@Import(ExamProcessorConfiguration.class)
+@Import({ ExamProcessorConfiguration.class, YamlPropertiesConfigurator.class })
```
malaffoon: This is the correct way to have multiple configuration classes in an app.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
```diff
@@ -76,7 +76,6 @@ public Authentication authenticate(final Authentication authentication) throws A
                 }
 
                 // extract roles from tenancy chain
-                // TODO - should this be client-specific? i.e. chain.hasRoleForClient(DataLoadRole)
```
malaffoon: My thought here: because this service has no idea what the content represents, it can't do a client-specific role lookup. So, it just makes sure this user has DataLoadRole mentioned somewhere in the tenancy chain.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
```diff
@@ -23,13 +24,30 @@
 @Repository
 class RdwImportRepositoryImpl implements RdwImportRepository {
 
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Value("${sql.import.count}")
```
malaffoon: @agorina, i think this is the best way for handling the sql properties: it gets them out of the constructor since the properties wiring is much simpler than that injection wiring (needed for the jdbcTemplate). If a class has a *ton* of sql properties, i suppose they could be pulled into a separate Configuration object but i played with that approach and didn't feel like there was any real value-add -- i just ended up having to type more stuff in for the same result.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultExamService.java
```diff
@@ -43,8 +43,6 @@ public RdwImport importExam(final RdwUser user, final String payload, final Stri
             return rdwImport;
         }
 
-        // TODO - should we distinguish between 200/201?
```
malaffoon: My thoughts on this: this end-point will always just return 202, Accepted. This seems to fit philosophically: we are accepting their request to import the data.

And, if the client resubmits an identical payload, they will get a 202 back where the status may have advanced.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultExamSource.java
```diff
@@ -16,7 +16,6 @@
 
 /**
  * Default implementation of ExamSource
- * TODO - should we do MessageProducerSupport stuff instead?
```
malaffoon: Yeah, that just adds a whole lot of complexity that we don't need here.

exam-service/src/main/resources/application.sql.yml
```diff
@@ -0,0 +1,17 @@
+sql:
```
malaffoon: I (strongly) feel that the sql properties should be scoped with `sql` so they all occupy the same space in the properties. That makes the nested `sql` redundant of course. This effectively flips the property key. So the pattern is `sql.<resource|table>.<method|function>`. For example, `sql.fubar.findOne` (vs the previous approach which would've been `fubar.findOne.sql`).

Thoughts?

malaffoon: The filename, `application.sql.yml` seems more consistent than `user-service.sql.yml`. At the very least, it doesn't need to be changed when i change the name of the project.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
agorina: Thank you!

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
agorina: Makes sense. This assumes that somewhere down the chain we need to verify that the client has access to the tenant. Just stating the obvious :)

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
agorina: Ok, sounds good. I will adjust my code to use this approach.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultExamService.java
agorina: Ok

exam-service/src/main/resources/application.sql.yml
agorina: I like your idea of sql.fubar.findOne.

agorina: This works as long as we have only one sql file.
For now I do not see a need of breaking it into multiple files. 
We can discuss it later if needed.

malaffoon: That's correct. I toyed with `import.sql.yml` or `all.sql.yml` as possibilities.

PR #32, opened 2017-03-01
add csv import

malaffoon:


malaffoon:
I have no idea if this will survive first contact with real data but i figured i could copy the csv handling over from the prototype so we have a way of easily getting a bunch of records in.

PR #34, opened 2017-03-02
docker-compose

malaffoon:
Add a docker-compose.yml spec and adjust README.md to use it.

NOTE: this reverts back to a native MySQL instance.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,44 @@
+# docker-compose specs for rdw ingest processes
+#
+# Before running this spec, set the DATAWAREHOUSE_HOST; if that is the (mac) docker host:
+# $ export DATAWAREHOUSE_HOST=$(ipconfig getifaddr en0)
+# This value must be reset whenever you change networks.
+#
+# TODO - healthcheck for rabbitmq, e.g. curl -u guest:guest http://localhost:5672/api/overview  -> AMQP
+
+version: '2'
+services:
+  rabbitmq:
+    image: rabbitmq:3-management
+    hostname: rabbitmq
+    ports:
+      - "5672:5672"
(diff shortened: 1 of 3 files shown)


docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+# docker-compose specs for rdw ingest processes
+#
+# Before running this spec, set the DATAWAREHOUSE_HOST; if that is the (mac) docker host:
```
malaffoon: Please note this bit when you pull code and want to run things locally.

exam-service/src/main/resources/mapping.xml
```diff
@@ -3,14 +3,33 @@
   <stream name="exam" format="csv" mode="read">
```
malaffoon: Can probably ignore this -- it is just me trying to get CSV import working. Since i've now spent in excess of 6 hours on CSV import, i'll probably give up. The data is just too junky (we knew it was incomplete but i was really hoping i could use it to ingest bunches of data).

docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      - "8080:8080"
+    depends_on:
+      - rabbitmq
+    links:
+      - rabbitmq
+    environment:
+      - spring.rabbitmq.host=rabbitmq
+      - spring.datasource.url=jdbc:mysql://dwhost:3306/warehouse?useSSL=false
+    extra_hosts:
+      - "dwhost:$DATAWAREHOUSE_HOST"
```
malaffoon: @wtritch i added you so you could comment on this approach. Instead of using an env variable to define the host IP, i'm using a host definition. Any opinion? Probably this is temporary anyway since i'll be doing the remote config server next but ...

PR #35, opened 2017-03-02
add message to import

malaffoon:
minor tweak to start fleshing out import records, produces import objects like:
`{
    "id": 43,
    "content": "EXAM",
    "contentType": "text/csv",
    "digest": "6CECD8358506280DD938608E314F6501",
    "status": "INVALID",
    "batch": "3C70D4EC0194DCB626E8FAE8C16BF3E3",
    "creator": "dwtest@example.com",
    "created": {
      "nano": 643000000,
      "epochSecond": 1488472962
    },
    "message": "org.beanio.InvalidRecordException: Invalid 'tdsReport' record at line 8\n ==> Invalid 'academicYear':  Type conversion error: Invalid Long value 'NotAYear'"
  }`

### DIFF
--- a/exam-service/src/main/java/org/opentestsystem/rdw/ingest/model/ImportStatus.java
+++ b/exam-service/src/main/java/org/opentestsystem/rdw/ingest/model/ImportStatus.java
@@ -4,9 +4,13 @@ package org.opentestsystem.rdw.ingest.model;
  * Status for imports
  */
 public enum ImportStatus {
+    UNAUTHORIZED(-4),
+    BAD_DATA(-3),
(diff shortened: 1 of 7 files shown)


exam-service/src/main/java/org/opentestsystem/rdw/ingest/model/ImportStatus.java
```diff
@@ -4,9 +4,13 @@
  * Status for imports
  */
 public enum ImportStatus {
+    UNAUTHORIZED(-4),
```
malaffoon: Just throwing out ideas for how an import may be rejected.

PR #36, opened 2017-03-03
Cleaned up repos and added unit tests for them.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
```diff
@@ -68,6 +67,9 @@ public long create(final Exam exam) {
     }
 
     private void batchCreateExamClaims(final List<ExamClaim> examClaims, final long examId) {
+        //TODO: is it correct?
```
agorina: Should the repo assume that the service does the validation and if it is given an ICA exam with no claims - that is alright?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
```diff
@@ -100,6 +102,9 @@ private long create(final StudentExamAttributes studentExamAttributes) {
     }
 
     private void batchCreateExamItems(final List<ExamItem> examItems, final long examId) {
+        //TODO: is it correct?
+        if (examItems.isEmpty()) return;
```
agorina: Same as above but for items?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
malaffoon: I"d vote yes, but i can see both sides of it.

agorina: Ok. Then I will assume this is right.

PR #37, opened 2017-03-03
flesh out marshaling

malaffoon:
Tweak ObjectMapper to deal with java 8 time, and null values.
Add HATEOAS link to imports.

BEFORE:
```
{
  "id": 69,
  "content": "EXAM",
  "contentType": "application/xml",
  "digest": "CEE97BE54F21345CAE7D24CB33A969AF",
  "status": "ACCEPTED",
  "batch": null,
  "creator": "dwtest@example.com",
  "created": {
    "nano": 797000000,
    "epochSecond": 1488572552
  },
  "message": null
}
```

AFTER:
```
{
  "id": 69,
  "content": "EXAM",
  "contentType": "application/xml",
  "digest": "CEE97BE54F21345CAE7D24CB33A969AF",
  "status": "ACCEPTED",
  "creator": "dwtest@example.com",
  "created": "2017-03-03T20:22:32.797Z",
  "_links": {
    "self": {
      "href": "http://localhost:8080/exams/69"
    }
  }
}
```

### DIFF
--- a/exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
+++ b/exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
@@ -57,7 +57,6 @@ class RdwImportRepositoryImpl implements RdwImportRepository {
     }
 
     @Override
-    @Transactional(readOnly = false)
     public RdwImport create(final RdwImport rdwImport) {
         final KeyHolder keyHolder = new GeneratedKeyHolder();
         final SqlParameterSource parameterSource = new MapSqlParameterSource()
@@ -78,7 +77,6 @@ class RdwImportRepositoryImpl implements RdwImportRepository {
     }
 
     @Override
-    @Transactional(readOnly = true)
     public RdwImport findOne(final long id) {
         final MapSqlParameterSource parameterSource = new MapSqlParameterSource().addValue("id", id);
         try {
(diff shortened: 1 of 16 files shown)


exam-service/src/test/java/org/opentestsystem/rdw/ingest/web/ExamControllerTest.java
```diff
@@ -92,13 +93,52 @@ public void itShouldUseServiceToGetExam() throws Exception {
...
+    }
+
+    @Test
+    public void itShouldReturn4xxWhenNoCriteria() throws Exception {
+        mvc.perform(get("exams/imports"))
+                .andExpect(status().is4xxClientError());
+    }
+
+    @Test
+    public void itShouldReturn5xxWhenMultipleCriteria() throws Exception {
```
malaffoon: Oops, test name is wrong, should be `4xx`.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
```diff
@@ -53,6 +57,7 @@
     }
 
     @Override
+    @Transactional(readOnly = false)
     public RdwImport create(final RdwImport rdwImport) {
```
agorina: Just curious: since all of the seem to run just one query, why did you decide to use Transactional?

malaffoon: Probably a mistake. If there is no annotation, does it default to having a transaction or not? I was thinking that i didn't want it to default to a read-write transaction.

agorina: My understanding is that without an annotation there is no transaction.

PR #38, opened 2017-03-06
More refactoring in preparation for the different validations.

agorina:
This should support the following validation logic:
1. authorize the user with the given tenancy chain
2. verify that the assessment packages are loaded
3. parsing and codified fields validation

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -1,8 +1,9 @@
 package org.opentestsystem.rdw.ingest.processor;
 
-import org.opentestsystem.rdw.ingest.processor.service.ExamService;
+import java.io.ByteArrayInputStream;
+import java.io.UnsupportedEncodingException;
+import org.opentestsystem.rdw.ingest.processor.service.TDSReportProcessor;
 import org.opentestsystem.rdw.messaging.RdwMessageHeaderAccessor;
-import org.opentestsystem.rdw.model.TDSReport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -12,9 +13,6 @@ import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.Message;
 
-import java.io.ByteArrayInputStream;
-import java.io.UnsupportedEncodingException;
-
 import static org.opentestsystem.rdw.model.XmlUtils.tdsReportFromXml;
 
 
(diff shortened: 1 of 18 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/ParsedExaminee.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String School = "school";
+
+    /**
+     * Constructor
+     *
+     * @param school                   the {@link School}
+     * @param student                  the {@link Student}
+     * @param studentAttributesBuilder the {@link StudentExamAttributes.Builder}
+     */
+    public ParsedExaminee(final School school, final Student student, final StudentExamAttributes.Builder studentAttributesBuilder) {
```
agorina: This suggests that something is not right with the data model since I have to use a builder here.
Since I am trying to avoid doing extra data copying or lookup, I am not sure what the best way to address it.
For now it works :)
Open to refactor it later if people have better ideas.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -36,9 +34,7 @@ public void process(final Message<?> message) throws UnsupportedEncodingExceptio
         logger.info("received " + accessor.getContent() + " from " + accessor.getUserLogin());
         logger.info(accessor.getContentType() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
 
-        final TDSReport report = tdsReportFromXml(new ByteArrayInputStream(payload.getBytes("UTF-8")));
-        //TODO: need to do something with the message header
-        service.process(report);
-        //TODO: need to do post-processing: provenance and status update-?
+        //TODO: need to get a tenancy chain and import id
```
agorina: @malaffoon - will need your help here.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/ExamParser.java
```diff
@@ -0,0 +1,55 @@
...
+import org.opentestsystem.rdw.model.Opportunity;
+import org.opentestsystem.rdw.model.Opportunity.Score;
+import org.springframework.stereotype.Component;
+
+import static java.lang.Double.parseDouble;
+import static java.lang.Float.parseFloat;
+import static java.util.stream.Collectors.toMap;
+
+@Component
+class ExamParser extends AnyExamParser<Exam, Builder> {
```
malaffoon: I'd like to see more class comments that make it easier to quickly see what value this class adds.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
malaffoon: ok, once this is committed, i'll add the injection of the tenancy chain and import id

agorina:
@malaffoon - would like to have a discussion with you on how we would like to handle various validation errors. Options I see:
1. define a new status for every possible error
2. combine errors into more generic ones and then a descriptive message. 

Let's chat when you have time.

PR #41, opened 2017-03-07
Cache - code review feedback.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
```diff
@@ -10,6 +10,7 @@
      *
      * @param id an id of an assessment
      * @return a code, null if not found
+     * @throws IllegalArgumentException is the given code is not found
```
agorina: @malaffoon - I am debating if I should define a specific exception for it and add it to the signature as _throwable_. What do you think?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentTypeRepositoryImpl.java
```diff
@@ -22,11 +23,12 @@ public AssessmentTypeRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplat
...
 
     @Override
-    public String toCode(final int id) {
+    @Cacheable(value = "assessmentType")
+    public String findCodeById(final int id) {
         try {
             return jdbcTemplate.queryForObject(sqlToCode, new MapSqlParameterSource("id", id), String.class);
         } catch (final EmptyResultDataAccessException ignore) {
-            return null;
+            throw new IllegalArgumentException("unknown assessment type id = [" + id + "]");
```
agorina: I want to throw so that it does not get cached. This way a value can be added in the DB without requiring to restart the app.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentTypeRepositoryIT.java
```diff
@@ -14,20 +16,50 @@
...
+    public void itShouldNotCacheUnknownResults() {
+        final Cache codes = this.cacheManager.getCache("assessmentType");
+        assertThat(codes.get(77)).isNull();
+
+        try {
+            repository.findCodeById(77);
+        } catch (final IllegalArgumentException ignore) {
+
+        }
+        assertThat(codes.get(77)).isNull();
```
agorina: Is this test redundant?
I am testing that the cache is not populated when the exception is thrown. 
Should this be a default behaviour of the cache?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
malaffoon: I don't like adding runtime exceptions that don't provide real value (i.e. extra data).

agorina: Ok, then you made my decision.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentTypeRepositoryIT.java
malaffoon: You should only be testing your code. So really it should just be testing that it throws for unknown values.

agorina: Ok, then I will remove this test.

PR #42, opened 2017-03-07
minimalist cache config

malaffoon:
@agorina for your consideration ...


exam-processor/src/test/resources/application.yml
```diff
@@ -1,9 +0,0 @@
-
```
malaffoon: Kind of unrelated -- not sure why we had this since all the values are the same.

PR #49, opened 2017-03-09
Added import exception and repo to support status update.

agorina:


common/src/main/java/org/opentestsystem/rdw/ingest/model/ImportException.java
```diff
@@ -0,0 +1,27 @@
+package org.opentestsystem.rdw.ingest.model;
+
+/**
+ * Exception that represents a processing error
+ */
+public class ImportException extends RuntimeException {
+
```
agorina: Should this be in the common?

common/src/main/java/org/opentestsystem/rdw/ingest/model/ImportStatus.java
```diff
@@ -10,6 +10,7 @@
  * </p>
  */
 public enum ImportStatus {
+    UNKNOWN_ASMT(-5),
```
agorina: @malaffoon , you will need to update the RDW_Schema.

common/src/main/java/org/opentestsystem/rdw/ingest/model/ImportException.java
malaffoon: Meh, eventually. @rcordeau's app will need something similar.

PR #51, opened 2017-03-09
Better error handling.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
```diff
@@ -37,8 +42,14 @@ public AnyExamProcessor(final R repository, final AnyExamParser<E, B> parser) {
...
      * Process the exam
      */
-    public void process(final Test report, final Opportunity opportunity, final StudentExamAttributes studentExamAttributes, final Assessment assessment) throws ParseException {
-        repository.create(parser.parse(report, opportunity, studentExamAttributes, assessment));
+    public void process(final Test report, final Opportunity opportunity, final StudentExamAttributes studentExamAttributes, final Assessment assessment) throws ImportException {
+        //todo: hanlde validation here; Ex. number of claims, score values (min/mx) and so on
+        try {
+            repository.create(parser.parse(report, opportunity, studentExamAttributes, assessment));
+        } catch (final ParseException e) {
+            logger.error("parsed exception while processing exam, " + e.getMessage());
```
malaffoon: This error is being logged more than once. I would suggest an hour-long debate with the team as to where the logging should happen. Personally, i would remove the logging here because you know it will be logged further up. But ... uhmm, have to debate it.

agorina: Ok. 
Would you like me to set up a meeting? 
Or should we discuss it at the next parking lot?

PR #53, opened 2017-03-11
Feature/config

malaffoon:
First attempt at using config server.

The gitlab repo has the actual config files: https://gitlab.com/fairwaytech/sbac-config-repo/tree/master

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,31 @@
 
 This document is targeted at developers contributing to the RDW_Ingest project.
 
+### Coding Conventions
+
+#### Application Configuration
+The ingest applications rely on a centralized configuration server that pulls property files from a git repo. Every
+application therefore has two property files: the embedded application.yml file and the config-repo application.yml
+file (of course properties may be put in separate yml files, there will be two instances of each). Properties fall 
+into three broad categories: code-like, configuration, and secrets. Code-like properties should be specified only in
+the embedded file. Configuration properties should have a reasonable (developer) default in the embedded file and
+profile-specific variations in the config-repo file. Secrets should be specified only in the config-repo file, and
+they should be encrypted using the config server.
+
+### Version Control Conventions
+Repo: https://github.com/SmarterApp/RDW_Ingest
+Config Repo: https://gitlab.com/fairwaytech/sbac-config-repo
+
+This project follows the common convention of having two main branches with infinite lifetime: `master` is the main
+branch where HEAD contains the production-ready state, while `develop` is the main branch where HEAD contains the 
+latest changes for the next release.
+ 
+Use feature branches off of `develop` for all new features. Use a prefix of `feature/` to highlight those branches.
+For example, the new shoesize feature work would be in `feature/shoesize`. Create pull requests from the feature
+branch to `develop` to solicit code reviews and feedback. Once approved use `squash and merge` into `develop`.
+
+### Running
+
 #### Running From IDE
 The README outlines how to run the applications using docker. As a developer you will want to run an application 
 locally from the IDE. This works fine; after using docker-compose to launch all the other applications, stop the
@@ -25,13 +50,3 @@ java -jar exam-processor/build/libs/rdw-ingest-exam-processor-0.0.1-SNAPSHOT.jar
 You will need valid credentials and connectivity to our SSO OAuth2 server. 
 
 
-### Coding Conventions
-
-### Version Control Conventions
-This project follows the common convention of having two main branches with infinite lifetime: `master` is the main
-branch where HEAD contains the production-ready state, while `develop` is the main branch where HEAD contains the 
-latest changes for the next release.
- 
-Use feature branches off of `develop` for all new features. Use a prefix of `feature/` to highlight those branches.
-For example, the new shoesize feature work would be in `feature/shoesize`. Create pull requests from the feature
-branch to `develop` to solicit code reviews and feedback. Once approved use `squash and merge` into `develop`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,10 +8,12 @@ This document is targeted at developers contributing to the RDW_Ingest project.
 The ingest applications rely on a centralized configuration server that pulls property files from a git repo. Every
 application therefore has two property files: the embedded application.yml file and the config-repo application.yml
 file (of course properties may be put in separate yml files, there will be two instances of each). Properties fall 
-into three broad categories: code-like, configuration, and secrets. Code-like properties should be specified only in
-the embedded file. Configuration properties should have a reasonable (developer) default in the embedded file and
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
+## RDW_Ingest for Developers
```
malaffoon: Trying to simplify the README by pulling out developer-specific notes.

exam-service/src/main/resources/application.yml
```diff
@@ -1,7 +1,14 @@
...
+# or we need to enable security and create an internal user that has the ACTUATOR role
+management:
+  security:
+    enabled: false
+
 oauth2:
   access-token-uri: https://sso-deployment.sbtds.org:443/auth/oauth2/access_token?realm=/sbac
   client-id: sbacdw
-  client-secret: sbacdw123
+  client-secret:
```
malaffoon: The whole point. This secret should never have been checked in this way.

docker/docker-compose.yml
```diff
@@ -1,42 +1,78 @@
...
 
 version: '2'
 services:
+  config-server:
+    image: fwsbac/configuration-service:latest
+    container_name: config-server
+    ports:
+      - 8888:8888
+    volumes:
+      - $SBAC_CONFIG_REPO:/sbac-config-repo
```
malaffoon: @wtritch i'm not thrilled with this but it gets around every developer having to have the same folder organization for their code (i.e. ${user.home}/smarter_app/sbac-config-repo doesn't have to be where they pulled down the config repo). Thoughts?

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+### Running
+
+#### Running From IDE
+The README outlines how to run the applications using docker. As a developer you will want to run an application 
+locally from the IDE. This works fine; after using docker-compose to launch all the other applications, stop the
+application you're working on, then launch it from the IDE. Note that this defaults to running without a 
+configuration server; since the embedded property files do not contain any secrets they must be specified on the
+command line. For example, the exam-service needs the oauth2 client secret so it must be specified in the program 
+arguments, i.e. `--oauth2.client-secret=mysecret`. The server port may also have to be specified to avoid conflict 
+with other running applications, e.g. `--server.port=8081`.
```
wtritch: I've been checking in a "local-development" profile in the source-code application.yml with reasonable standalone-defaults.  The point being that a new developer should just be able to activate a profile and start the project without have to go figure out what they're missing.

malaffoon: I agree. Except i think that profile should be `default` and not have to be called out with a special name.

My opinion is not necessarily that of my employer.

docker/docker-compose.yml
wtritch: Interesting.  I never thought of mounting a host machine volume into the docker-compose container in order to use file system properties... I'll think about this.

malaffoon: I'm guessing that, over time, i will give up on mounting a local file system and just assume connectivity to gitlab with credentials and all that.

CONTRIBUTING.md
wtritch: It can't always be 'default' because once a property is defined, it can't be overridden to 'non-existent'  I ran into an issue with configuring spring data mongodb where if `spring.data.mongodb.username` is defined it completely ignores `spring.data.mongodb.uri` which was needed to configure a replica set.
I now prefer to leave the `default` properties generally empty except for relatively immutable configuration values. (At least, ones that will only ever be changed, never removed.)

As for secrets... I haven't run into that issue, yet, where a service is completely un-runnable without auth access to an external service. =)

exam-service/src/main/resources/bootstrap.yml
```diff
@@ -3,4 +3,12 @@
...
-    name: exam-service
\ No newline at end of file
+    name: rdw-ingest-exam-service
+
+  cloud:
+    config:
+      enabled: ${CONFIG_SERVICE_ENABLED:false}
+      fail-fast: ${CONFIG_SERVICE_ENABLED:false}
+      label: ${CONFIG_SERVICE_LABEL:develop,master}
+      profile: ${CONFIG_SERVICE_PROFILE:default}
```
wtritch: Interesting, why double-define this here?  Wouldn't you want to pass the currently-active profile(s) automatically?

exam-service/src/main/resources/bootstrap.yml
```diff
@@ -3,4 +3,12 @@
...
   application:
-    name: exam-service
\ No newline at end of file
+    name: rdw-ingest-exam-service
+
+  cloud:
+    config:
+      enabled: ${CONFIG_SERVICE_ENABLED:false}
+      fail-fast: ${CONFIG_SERVICE_ENABLED:false}
+      label: ${CONFIG_SERVICE_LABEL:develop,master}
```
wtritch: I like this...

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     dependencies {
         classpath("com.bmuschko:gradle-docker-plugin:3.0.5")
         classpath("io.spring.gradle:dependency-management-plugin:1.0.0.RELEASE")
-        classpath("org.springframework.boot:spring-boot-gradle-plugin:1.5.1.RELEASE")
+        classpath("org.springframework.boot:spring-boot-gradle-plugin:1.5.2.RELEASE")
```
wtritch: I don't know if you've hit this yet, but we recently ran into some version conflicts where `spring boot` 1.5.2 has a different core spring version dependency than `spring cloud 1.2.3` I ended up sticking with spring boot 1.4.2 since it has the same core spring version as spring cloud 1.2.3

docker/docker-compose.yml
wtritch: I hope using a file-system for externalized configurations is a relatively rare occurrence.  This'll do the job, but I don't think Spring Cloud Configuration was designed to use a file-system-backed configuration repository.  Which means any solution is likely to be a little messy.

exam-service/src/main/resources/bootstrap.yml
malaffoon: Double define? I'm not sure where else the active profile is specified. Bear with me ...

The code itself has no active profile defined in it. 
You can set the profile by passing in a command-line argument.
But that would make one of these things not like the other. 
What am i missing?

CONTRIBUTING.md
malaffoon: The one the currently bites us is the oauth2 server. I suppose we could jack in a no-op shim that returns a hard-coded tenancy chain so that standalone doesn't need the secret. But that's like work and stuff.

docker/docker-compose.yml
malaffoon: Yeah, i think i'll write myself a TODO to just document setting the gitlab credentials in environment variables and getting rid of this.

PR #54, opened 2017-03-12
add local-auth option

malaffoon:
I'm not 100% sure of this change and would like thoughts.

What it does is introduce a local auth provider that is instantiated if no other auth provider is configured. This auth provider reads a file to get user credentials. This will allow us to more easily test with arbitrary tenancy chains without having to create users in ART. The obvious downside is that, should this escape into the wild, it would allow somebody to configure a user without having to use ART. We can easily protect against that but it is still a possible security concern ...

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -46,7 +46,6 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
-      - CONFIG_SERVICE_LABEL=rdwauth,master
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
 
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -46,6 +46,7 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
+      - CONFIG_SERVICE_LABEL=rdwauth,master
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
 
--- a/exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthConfig.java
+++ b/exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthConfig.java
@@ -1,0 +1,38 @@
+package org.opentestsystem.rdw.ingest.auth;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+/**
+ * Configuration for auth provider
+ */
+@Configuration
+@EnableConfigurationProperties({OAuth2Properties.class})
+@EnableWebSecurity
+public class AuthConfig {
+
+    @Autowired
+    private OAuth2Properties oauth2Properties;
+
+    @Bean
+    @AuthProviderType("oauth2")
+    public AuthenticationProvider oauth2AuthenticationProvider() {
+        return new OAuth2AuthenticationProvider(oauth2Properties);
+    }
+
+    @Bean
+    @AuthProviderType("local")
+    public AuthenticationProvider localProvider() {
+        return new LocalAuthenticationProvider();
+    }
+
+    @Bean
+    public WebSecurityConfigurerAdapter securityConfigurer() {
+        return new SecurityConfigurer();
+    }
+}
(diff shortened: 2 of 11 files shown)


docker/docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
+      - CONFIG_SERVICE_LABEL=rdwauth,master
```
malaffoon: i won't check this in, it is just how you get local-docker to look at a branch in the config-repo

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+/**
+ * Configuration for auth provider
+ */
+@Configuration
+@EnableConfigurationProperties({OAuth2Properties.class})
+@EnableWebSecurity
+public class AuthConfig {
```
malaffoon: This is the configuration for instantiating the auth provider.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthProviderType.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for conditionally selecting auth provider
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Conditional(AuthProviderTypeCondition.class)
+public @interface AuthProviderType {
```
malaffoon: necessary wiring, nothing to see

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthProviderTypeCondition.java
```diff
@@ -0,0 +1,31 @@
...
+public class AuthProviderTypeCondition implements Condition {
+
+    @Override
+    public boolean matches(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
+        final Map<String, Object> attributes = metadata.getAnnotationAttributes(AuthProviderType.class.getName());
+        final String type = (String) attributes.get("value");
+
+        final Environment env = context.getEnvironment();
+
+        if ("oauth2".equalsIgnoreCase(type)) {
```
malaffoon: This is the place that decides which auth provider is allowed to be instantiated. Currently it just checks to see if oauth2 appears to be configured. We could add checks against profiles, other environment variables, etc.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/LocalAuthenticationProvider.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private List<RdwUser> userInfo = newArrayList();
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        try {
+            // format of file lines is expected to be:
+            // username, password, tenancyChain
+            final File userInfoFile = ResourceUtils.getFile(userInfoUri);
+            try (final BufferedReader reader = new BufferedReader(new FileReader(userInfoFile))) {
```
malaffoon: Reads in file to get list of users.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
```diff
@@ -5,46 +5,34 @@
...
 import java.util.concurrent.TimeUnit;
 
-import static org.opentestsystem.rdw.ingest.auth.SecurityConfig.DataLoadRole;
-
 /**
  * This auth provider acts as an adapter between basic auth from the clients of our RESTful API and the
  * credentials stored in our IDP which are accessed via oauth2. If things change so our clients are aware
  * of oauth2 and use it fully, this can go away and we can use magick Spring wiring (i think).
  */
-@Component
```
malaffoon: Had to switch this to being handled directly as a bean instead of auto-instantiating as a component.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/OAuth2AuthenticationProvider.java
```diff
@@ -71,17 +59,8 @@ public Authentication authenticate(final Authentication authentication) throws A
                 final Map tokenInfo = template.getForObject(oauth2Properties.getTokenInfoUri(), Map.class,
                         Collections.singletonMap("access_token", token.getValue()));
                 final TenancyChain chain = TenancyChain.fromString((String) tokenInfo.get("sbacTenancyChain"));
-                if (chain.isEmpty()) {
```
malaffoon: Introduced a helper method for this.

agorina:
I like this idea.
 It will make the testing way easier. 
If we are concerned about the 'escape into the wild', we can have a TODO to remove it before we release.

PR #61, opened 2017-03-15
caching test helper

malaffoon:
@agorina, submitted for your review. This allows an easy annotation for tests that use implementations that use caching; it clears cache before every test.

I couldn't find a built-in spring thing that does this and this is more surgical than @DirtiesContext.

If you're okay with it, i'll change all the existing ITs and merge it.

### DIFF
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/CachingTestExecutionListener.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/CachingTestExecutionListener.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.ingest.processor;
 
+import org.opentestsystem.rdw.ingest.processor.repository.impl.CachingTest;
 import org.springframework.cache.CacheManager;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.core.annotation.AnnotatedElementUtils;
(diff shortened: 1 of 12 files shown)


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/GenderRepositoryIT.java
```diff
@@ -19,8 +17,7 @@
 @RunWith(SpringRunner.class)
 @SpringBootTest
 @Transactional
-@EnableCaching
-@AutoConfigureCache(cacheProvider = CacheType.SIMPLE)
+@CachingTest
```
malaffoon: This is the annotation for the IT's.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/GenderRepositoryIT.java
```diff
@@ -54,4 +51,16 @@ public void itShouldCacheReturnedId() {
 
         assertThat(codes.get("test55", Integer.class)).isEqualTo(55);
     }
+
+    @Test
```
malaffoon: One of these two tests fails when i don't use the new annotation (order dependent of course). With the new annotation, both tests always pass.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/CachingTestExecutionListener.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public void beforeTestMethod(final TestContext testContext) throws Exception {
+        if (!(Boolean)testContext.getAttribute(EnabledAttribute)) return;
+
+        final CacheManager cacheManager = testContext.getApplicationContext().getAutowireCapableBeanFactory().getBean(CacheManager.class);
+        for (final String name : cacheManager.getCacheNames()) {
+            cacheManager.getCache(name).clear();
+        }
+    }
+}
```
agorina: I like it, I was thinking of doing something similar, but you already did it. 
So now I can use it :)

PR #64, opened 2017-03-17
Added support for multiple ethnicity for a student.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Student.java
```diff
@@ -46,8 +48,8 @@ public int getGenderId() {
         return genderId;
     }
 
-    public int getEthnicityId() {
-        return ethnicityId;
+    public List<Integer> getEthnicityIds() {
+        return ethnicityIds == null ? newArrayList() : ethnicityIds;
```
malaffoon: Null handling seems a bit inconsistent. If the builder doesn't ever set it to null then why check it here? Also, i don't know how robust you want this to be, but you're returning a reference to the internal list allowing it to be modified by the caller. Personally, i would make it an Immutable list when the builder creates it (as opposed to copying the list in the getter).

agorina: Right...will change to the Immutable list. 

As far as the Null handling - I would like to defer it to the next milestone. I think all the builders in this module needs to be revisited. I will add comments to the JIRA that defines this task.

PR #65, opened 2017-03-17
add database status indicator, which required a little refactoring

malaffoon:
A sample of the response:
```
  "database": {
    "statusText": "Ideal",
    "statusRating": 4,
    "databaseOperations": [
      {
        "statusText": "Ideal",
        "statusRating": 4,
        "schema": "import",
        "responseTime": "2ms",
        "type": "READ"
      },
      {
        "statusText": "Ideal",
        "statusRating": 4,
        "schema": "import",
        "responseTime": "7ms",
        "type": "WRITE"
      }
    ]
  },
```

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/status/AbstractStatusIndicator.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/status/AbstractStatusIndicator.java
@@ -1,26 +1,60 @@
 package org.opentestsystem.rdw.ingest.status;
 
+import java.util.function.Supplier;
+
 /**
  * Helper to handle general exceptions for {@link StatusIndicator}s.
  */
 public abstract class AbstractStatusIndicator implements StatusIndicator {
 
     @Override
-    public Status status() {
-        Status.Builder builder = Status.builder();
+    public Status status(final int level) {
+        if (!doLevelCheck(level)) return null;
+
+        final Status.Builder builder = Status.builder();
         try {
-            doStatusCheck(builder);
+            doStatusCheck(builder, level);
         } catch (final Exception e) {
             builder.rating(Rating.Failed);
+            builder.detail("error", e.getMessage());
         }
         return builder.build();
     }
 
     /**
      * Perform the actual status check.
+     * This method will only be called with a level that passed {@link AbstractStatusIndicator#doLevelCheck(int)}.
      *
      * @param builder status builder
-     * @throws Exception if any problem, results in status of Failed
+     * @param level diagnostic level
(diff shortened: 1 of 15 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/status/AbstractStatusIndicator.java
```diff
@@ -1,26 +1,60 @@
 package org.opentestsystem.rdw.ingest.status;
 
+import java.util.function.Supplier;
+
 /**
  * Helper to handle general exceptions for {@link StatusIndicator}s.
  */
 public abstract class AbstractStatusIndicator implements StatusIndicator {
 
     @Override
```
malaffoon: A lot of the changes are due to this: instead of asking the StatusIndicator what level it is for, pass the level in to the status generation method. This is because the database indicator varies its behavior depending on the level. That is, it doesn't have a single level but does work for level=3 or level=4.

common/src/main/java/org/opentestsystem/rdw/ingest/status/ConfigurationStatusIndicator.java
```diff
@@ -0,0 +1,35 @@
...
+        return "configuration";
+    }
+
+    @Override
+    protected boolean doLevelCheck(final int level) {
+        return level >= 2;
+    }
+
+    @Override
+    protected void doStatusCheck(final Status.Builder builder, final int level) {
```
malaffoon: Let spring actuators do all our work for us.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/status/DatabaseStatusIndicator.java
```diff
@@ -0,0 +1,76 @@
...
+
+        // import, WRITE
+        if (level >= 4) {
+            final Status.Builder opBuilder = Status.builder().detail("schema", "import").detail("type", "WRITE");
+            final RdwImport rdwImport = responseTime(opBuilder, 200, () ->
+                repository.create(RdwImport.builder()
+                    .content(ImportContent.PROBE)
+                    .contentType("text/plain+probe")
+                    .digest("probe")
+                    .status(ImportStatus.INVALID)
```
agorina: Just curious: why INVALID status here?

malaffoon: Well, if something goes wrong and my probe entry stays in the database, i want it be as invalid as it can possibly be; that's why i chose crazy values for it.

PR #69, opened 2017-03-21
ldap

malaffoon:
Some of this is simpler, some of it is bizarrely more complex. I'll comment inline.
I'm very open to suggestions but it took me in excess of 8 hours to figure all this crap out.


docker/docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
       - spring.profiles.active=local-docker
       - CONFIG_SERVICE_ENABLED=true
       - CONFIG_SERVICE_URL=http://config-server:8888
+      - CONFIG_SERVICE_LABEL=ldap,master
```
malaffoon: I will probably leave this in, then commit the changes to the config-repo, then revert this.
Or i might just sneak everything in at once. This remote config thing is annoying sometimes.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/RdwUserContextMapper.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.util.Assert;
+
+import java.util.Collection;
+
+import static org.opentestsystem.rdw.ingest.auth.SecurityConfigurer.DataLoadAuthority;
+
+/**
+ * A {@link ContextMapper} to map to/from RdwUser, which adds TenancyChain
+ */
+public class RdwUserContextMapper implements UserDetailsContextMapper {
```
malaffoon: This class and the changes to RdwUser deal with creating user details with ldap info and our tenancy chain.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
```diff
@@ -1,31 +1,75 @@
 package org.opentestsystem.rdw.ingest.auth;
```
malaffoon: This class has all the magick.

rcordeau: Lots of magick

malaffoon: The depressing part is that it is only a few lines of code ... yet it took me a long time adding and removing hundreds of lines of code before it was as simple as it could be.

PR #73, opened 2017-03-22
make import repo generic

malaffoon:
I will have three (?) PRs to make what is currently `exam-service` into a more generic `import-service`. This first set of changes is going from the bottom-up making the repo and service layers more generic. @rcordeau, this is so you can use the same service for package import.


common/src/main/java/org/opentestsystem/rdw/ingest/model/ImportContent.java
```diff
@@ -5,7 +5,8 @@
  */
 public enum ImportContent {
     PROBE(0),
-    EXAM(1);
+    EXAM(1),
+    PACKAGE(2);
```
malaffoon: There is a change in RDW_Schema to go with this.

common/src/main/java/org/opentestsystem/rdw/ingest/model/RdwImportQuery.java
```diff
@@ -0,0 +1,68 @@
+package org.opentestsystem.rdw.ingest.model;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+/**
+ * Container for query parameters.
+ */
```
malaffoon: Again, this isn't strictly speaking necessary for this change; i could've left all the findBy() parameters as individual values. But using a query object is a bit cleaner.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
```diff
@@ -140,6 +121,32 @@ public void delete(final long id) {
     }
 
     @Override
+    public List<RdwImport> findBy(final RdwImportQuery query) {
+        // TODO - better way to do this using application.sql.yml?
```
malaffoon: @agorina thoughts?

agorina: Options I see (in a semi-random order):
1. write SQL like this 
```sql
SELECT * FROM import
  WHERE
         (:content is null or conent = :content)
    AND  (:status is null or status=:status)
    AND  (:batch is null or batch=:batch)
```

2. Introduce something similar to what we had at SRI with 'optional'. We may need it for the Paging/Sorting anyway.

3. Keep it as is. Admit I like it the least. Defeats the whole idea of sql.yml file.

malaffoon: I'll experiment with #1, thanks.

PR #74, opened 2017-03-22
Feature/reporting package loader

rcordeau:
Putting this in. Doesn't do much as of yet other than create a reporting package service&processor, and spits back out a parsed list of fields from a tabulator output csv file.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/74)


exam-processor/src/main/resources/application.yml
```diff
@@ -11,7 +11,7 @@ spring:
     stream:
       bindings:
         input:
-          destination: exams-payload
+          destination: exam-payload
```
malaffoon: Please note that this is now `EXAM` and directly correlates to ImportContent enum values.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultReportingPackageService.java
```diff
@@ -0,0 +1,83 @@
+package org.opentestsystem.rdw.ingest.service;
```
malaffoon: Are you going to merge with develop one more time before committing? That would get rid of this project. Or will you get rid of it as a subsequent change?

rcordeau: This pull request is closed. I'll do another one after I merge from develop.

exam-service/src/main/java/org/opentestsystem/rdw/ingest/service/ReportingPackageSource.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Abstraction for injecting an exam payload into the message queue.
+ */
+interface ReportingPackageSource {
+
+    /**
+     * Submit the given text to the "exam" queue
+     *
+     *  @param user user credentials
```
malaffoon: javadocs aren't correct and there is an unused import; but this will all go away so i'm not too worried.

PR #76, opened 2017-03-22
cleanup from my mess

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
```diff
@@ -10,9 +10,9 @@
 @SpringBootApplication
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
 @Import({YamlPropertiesConfigurator.class, StatusConfiguration.class})
-public class ExamServiceApplication {
+public class ImportServiceApplication {
```
malaffoon: missed a rename

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/WebMvcContext.java
```diff
@@ -2,13 +2,11 @@
 
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
-import org.springframework.web.servlet.config.annotation.EnableWebMvc;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
 import java.util.List;
 
 @Configuration
-@EnableWebMvc
```
malaffoon: This is a Bad Thing since it disables all default web mvc behavior.

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/ExamControllerTest.java
```diff
@@ -56,6 +57,7 @@ public void itShouldUseServiceToImportExam() throws Exception {
         mvc.perform(post("/exams/imports").contentType(MediaType.APPLICATION_XML).content(body))
                 .andExpect(status().is2xxSuccessful())
                 .andExpect(content().string(containsString("123")))
+                .andExpect(content().string(not(containsString("null"))))
```
malaffoon: Because of the `@EnableWebMvc` thing, marshaling got messed up. This new assert failed because it was sticking `'message': null` in the payload which we don't want.

PR #79, opened 2017-03-24
add archive service

malaffoon:
The archive service implementation is selected based on the configured archive.root property. This allows tests, devs, and local-docker to point to a local folder while remote/production env can use S3.


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/ArchiveConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+public class ArchiveConfiguration {
+
+    /**
+     * Create {@link ArchiveService} based on configuration setting of "archive.root".
+     *
+     * @param archiveRoot configured archive root
+     * @return new created {@link ArchiveService}
+     */
+    @Bean
+    public ArchiveService archiveService(@Value("${archive.root:file:///tmp/}") final String archiveRoot) {
```
malaffoon: Select impl based on property.

exam-processor/src/main/resources/bootstrap.yml
```diff
@@ -10,4 +10,8 @@ spring:
       enabled: ${CONFIG_SERVICE_ENABLED:false}
       fail-fast: ${CONFIG_SERVICE_ENABLED:false}
       label: ${CONFIG_SERVICE_LABEL:develop,master}
-      uri: ${CONFIG_SERVICE_URL:http://localhost:8888}
\ No newline at end of file
+      uri: ${CONFIG_SERVICE_URL:http://localhost:8888}
+      # defaults of 1000/2000 are a bit too needy
```
malaffoon: did this to all the apps since they were all making 3-5 requests on startup

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -40,11 +44,12 @@ public RdwImport importContent(final RdwUser user, final String payload, final I
 
         RdwImport rdwImport = repository.findOneByDigest(digest);
         if (rdwImport != null) {
-            logger.info("Ignoring exam payload with existing digest {}", digest);
+            logger.info("Ignoring content payload with existing digest {}", digest);
```
malaffoon: fixed some typos talking about `exam` instead of generic `content`

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -40,11 +44,12 @@ public RdwImport importContent(final RdwUser user, final String payload, final I
...
             return rdwImport;
         }
 
-        // TODO - save payload
```
malaffoon: finally

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/LocationStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+        @Override
+        public String location(final String id) {
+            return id;
+        }
+    }
+
+    /**
+     * For import content the id is a hex digest so it can be used to partition the files.
+     */
+    public static class ImportContentLocationStrategy implements LocationStrategy {
```
malaffoon: What this does is partition files based on the md5 digest. For example, a payload might hash to `D5A0852B21BB0DC2A98B5CF78D45F14F` so the payload will be stored in a file at `root/EXAM/D5/A0/D5A0852B21BB0DC2A98B5CF78D45F14F`.

import-service/src/test/java/org/opentestsystem/rdw/ingest/service/S3ArchiveServiceTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@TestPropertySource(properties = {
+        "archive.root=s3://rdw-archive2/",
+        "cloud.aws.credentials.accessKey: yourAccessKeyHere",
```
malaffoon: To run these tests you have to supply the accessKey and secretKey, so i had to @Ignore them. Not really sure how else to handle this since we can't checkin a secret.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -25,12 +25,16 @@
...
     private final ImportSource source;
 
     @Autowired
-    DefaultImportService(@NotNull RdwImportRepository repository, @NotNull final ImportSource source) {
+    DefaultImportService(@NotNull final RdwImportRepository repository,
+                         @NotNull final ArchiveService archiveService,
+                         @NotNull final ImportSource source) {
         this.repository = checkNotNull(repository);
+        this.archiveService = checkNotNull(archiveService);
         this.source = checkNotNull(source);
```
agorina: I am too lazy to google it: is the triple NotNull protection is needed ( Autowire, NoNull and checkNotNull) ? I was under impression that 'Autowired' will complain if the value is null?

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/ArchiveConfiguration.java
agorina: Neat idea!

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
malaffoon: Apparently, the way i'm doing it, the bean can be null. Even though it is required. I think that might be because of my ArchiveConfiguration class; i'll play with it.

agorina: My question was: what is the 'right' way of doing it? I wanted to research it, but did not get to it. Do you have an opinion?

malaffoon: I have an opinion ... and i'll remove most of this.

agorina: I think was not clear -  I would like to know your opinion :)

 Instead of sharing your opinion you want me to go and look at what you have changed?

malaffoon: I agree with your point: autowiring just works, it throws if there is a problem supplying a bean. So all the specification of NotNull and checking is pointless.

PR #81, opened 2017-03-27
add properties to ArchiveService

malaffoon:
I had a dream last night that we'd want properties and contentType was just the tip of the iceberg. So i added this, mostly so properties would be saved.


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -46,7 +47,13 @@ public RdwImport importContent(final RdwUser user, final String payload, final I
         }
 
         final String location = new LocationStrategy.ImportContentLocationStrategy(content).location(digest);
-        archiveService.writeResource(location, payload, contentType);
+        final Properties properties = new Properties();
```
malaffoon: Now, even if we lose the import record (or if it never got created properly), we still have metadata needed to resubmit.

PR #83, opened 2017-03-29
Added trait scores support.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
```diff
@@ -1,26 +1,28 @@
...
+    private final static int OrganizationAndPurpose = 2;
+    private final static int Conventions = 3;
+
+    private String naturalId;
+    private Double score;
+    private String scoreStatus;
+    private String response;
+    private Double maxScore;
+    private Long position;
+    private Map<Integer, ExamItemSubScore> subScoresBytraitId;
```
malaffoon: A map seems like overkill here. At most there are 3 of these right? A list would be more efficient; yeah, you'd want to write a little helper to replace the containsKey() call but that is easy.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private final Map<String, Integer> itemTraitIds;
+
+    /**
+     * Constructor
+     *
+     * @param repository the {@link ItemTraitRepository}
+     */
+    @Autowired
+    public DefaultExamItemProcessor(final ItemTraitRepository repository) {
```
malaffoon: in general should these autowired constructors be package-private?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
agorina: Really? I would think that the map implementation should be smart enough. You made me curious. I will check it out.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
agorina: You are right. Will change. Thank you.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
malaffoon: Smart? Sure, maps are smart. But they are big and slow compared to a list, when dealing with less than a couple dozen things. They also complicate marshaling, etc. although that might not be an issue here.

PR #84, opened 2017-03-29
More package loader additions

rcordeau:
Needs more unit and integrations tests. Needs queries and database insertion next as well.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/84)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -9,13 +10,13 @@
     private final String itemKey;
     private final String claimCode;
     private final String target;
-    private final float difficulty;
-    private final int dok;
-    private int mathPractice;
+    private final Float difficulty;
+    private final Integer dok;
+    private Integer mathPractice;
```
rcordeau: Using objects instead of primitives because values might be blank in the csv, which could be incorrect to assume blank == 0.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public void process(String assessment) throws ImportException {
+        try {
+            Iterable<CSVRecord> records = CSVFormat.RFC4180.withFirstRecordAsHeader().parse(new StringReader(assessment));
+
+            // consider pulling these out to separate processing modules - one for assessments and one for items
+            for(CSVRecord record : records) {
+                assessments.add(Assessment.builder()
```
rcordeau: This mess needs to be broken out in to separate assessment and item parsers that return representative objects.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -0,0 +1,175 @@
...
+        private String name;
+        private String label;
+        private String version;
+
+        public Assessment build() {
+            final Assessment assessment = new Assessment(naturalId, name, grade, label, type, subject, academicYear, version);
+            return assessment;
+        }
+
+        public Builder withNaturalId(final String naturalId) {
```
agorina: Just FYI: @malaffoon  talked me into dropping *with* from the builders. I am going to _silently_ modify all the builders. It really does not make a difference code-wise, but makes Mark happy :)

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+                        .withVersion(record.get("AssessmentVersion"))
+                        .build());
+
+               items.add(Item.builder()
+                        .withItemKey(record.get("ItemId"))
+                        .withClaimCode(record.get("Claim"))
+                        .withTarget(record.get("Target"))
+                        .withAllowCalculator(record.get("AllowCalculator").equals("Y"))
+                        .withDifficulty(floatValidator.validate(record.get("avg_b")))
+                        .withDOK(intValidator.validate(record.get("DOK")))
```
agorina: I know it is not final, just curious if you plan on validating the data against the _translation_ tables. Things like: claim, target, dok and so on.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorConfiguration.java
```diff
@@ -43,18 +48,15 @@ public void process(final Message<?> message) {
...
             importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
+        } catch (final ImportException ie) {
+            logger.error("failed with an import error: " + ie.getMessage());
+            importRepository.updateStatusAndMessageById(importId, ie.getStatus(), ie.getMessage());
         } catch (final Exception e) {
-            logger.error("failed with an unexpected import error: " + e.getMessage());
+            logger.error("failed with an unexpected error: " + e.getMessage());
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
         }
         //TODO: actually process the incoming record (convert to classes as necessary)
```
malaffoon: TODone?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -0,0 +1,175 @@
...
+import com.mysql.jdbc.StringUtils;
+
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * An assessment definition
+ */
```
malaffoon: nit: include a purpose for this, i.e. is it used just to marshal to database?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -0,0 +1,175 @@
...
+    private final String grade;
+    private final String type;
+    private final String subject;
+    private final int academicYear;
+    private final String naturalId;
+    private final String name;
+    private final String label;
+    private final String version;
+
+    private Assessment(final String naturalId, final String name, final String grade, final String label,
```
malaffoon: Kind of ugly and a minor annoyance to maintain; since it is all hidden, i'd suggest getting rid of giant constructor and just setting values directly in builder.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public String getSubject() {
+        return subject;
+    }
+
+    public int getAcademicYear() {
+        return academicYear;
+    }
+
+    public static Builder builder() {
```
malaffoon: minor nit: static methods shouldn't really be mixed in with non-static methods, perhaps move getNaturalId() up a bit?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -9,13 +10,13 @@
...
     private boolean allowCalculator;
 
-    public Item(@NotNull final String itemKey, @NotNull final String claimCode, @NotNull final String target,
-                @NotNull final float difficulty, @NotNull final int dok, int mathPractice, boolean allowCalculator) {
+    private Item(@NotNull final String itemKey, @NotNull final String claimCode, @NotNull final String target,
```
malaffoon: params should all be final; although perhaps moot because i think we should ditch giant constructor (see comment above)

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Repository for dealing with the claims
+ */
+public interface ClaimRepository {
+    /**
+     * Finds an id of the subject by its code
+     *
+     * @param code the code of the subject
+     * @return the id
+     * @throws IllegalArgumentException is the given name is not found
```
malaffoon: Just confirming, is this the behavior? So it will never return null? 

We should be consistent; i favor repositories returning null (not throw exception) for not found, while the higher layers (i.e. service) can introduce exceptions or optional or whatever.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+        this.assessments = new HashSet<Assessment>();
+        this.items = new ArrayList<Item>();
+        this.intValidator = IntegerValidator.getInstance();
+        this.floatValidator = FloatValidator.getInstance();
+    }
+
+    @Override
+    public void process(String assessment) throws ImportException {
+        try {
+            Iterable<CSVRecord> records = CSVFormat.RFC4180.withFirstRecordAsHeader().parse(new StringReader(assessment));
```
malaffoon: i know we haven't really talked about it as a team, but i find `final` helps more than it hurts and suggest using it by default. (Actually, in this case i'd probably just inline the iterable.)

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PackageProessorConfigurationTest.java
```diff
@@ -21,35 +25,54 @@
...
+                .process(any(String.class));
+
+        processor.process(message);
+
+        verify(importRepository, times(1)).updateStatusAndMessageById(1, ImportStatus.UNKNOWN_ASMT, "message");
+    }
+
+    @Test
+    public void itShouldhandleAnyException() throws UnsupportedEncodingException {
+        doThrow(new RuntimeException("any message")).when(packageProcessor).process(any(String.class));
```
malaffoon: minor nit: this is only demonstrating that it handles RuntimeExceptions, should there be another test for Exception or just change this to an Exception?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Ok. On the next iteration I'll clean them up.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
rcordeau: Didn't even think about it actually. We should discuss it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorConfiguration.java
rcordeau: Whoops.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Will fix on next iteration.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Yeah, I thought about that as well.  I was actually inclined to not even have a builder for these models because I don't really plan on building them up over time. But, I'll stick with the team motif.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: I'll reorganize. Nit picker.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
rcordeau: Will do.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -37,19 +38,96 @@ public String getTarget() {
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * This instance is equal to all instances of {@code Assessment} that have equal attribute values.
+     * @return {@code true} if {@code this} is equal to {@code another} instance
+     */
+    @Override
+    public boolean equals(Object another) {
```
rcordeau: No, they aren't boiler plate.  I considered using this object in a set, and I still might.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
rcordeau: I was trying to be consistent with the exam processor. I would agree that if it isn't found, null is appropriate return value.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
rcordeau: Final it is.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PackageProessorConfigurationTest.java
rcordeau: Which would you prefer? Perhaps another test for Exception.

PR #86, opened 2017-04-06
Feature/reporting package loader

rcordeau:
Adds data validation, more integration tests, more unit tests, repository functionality.  Batch inserting isn't working yet, so that code wasn't included.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/86)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -106,70 +118,89 @@ public int hashCode() {
...
+        //Not checked against database, but required to exist
+        private List<Integer> cutPoints;
+        private Integer minScore;
+        private Integer maxScore;
+
         public Assessment build() {
-            final Assessment assessment = new Assessment(naturalId, name, grade, label, type, subject, academicYear, version);
+            checkArgument(!StringUtils.isNullOrEmpty(naturalId), "Assessment naturalId cannot be null");
+            try {
+                checkNotNull(cutPoints, "Assessment cut points cannot be null");
```
rcordeau: Validate in the builder as well as when parsing begins? Seems like having double validation is overkill, but I'm looking for opinions. Since the package loader is the only one using these classes, and data validation will occur before setting, is this necessary here?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentRepositoryImpl.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+    /**
+     * Finds an {@link Assessment} by the natural id.
+     *
+     * @param id the natural id
+     * @return the found {@link Assessment}; null if not found
+     */
+    @Override
+    public Assessment findOneByNaturalId(String id) {
+        return null;
```
rcordeau: Not using this yet.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ItemRepositoryImpl.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Finds an id of the item by its item_key
+     *
+     * @param item_key the item of the subject
+     * @return the id
+     * @throws IllegalArgumentException is the given name is not found
+     */
+    @Override
+    public Integer findIdByKey(String item_key) {
+        return null;
```
rcordeau: Not using this yet.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -0,0 +1,95 @@
...
+    @Autowired
+    public DefaultAssessmentParser(final GradeRepository gradeRepository,
+                                   final SubjectRepository subjectRepository,
+                                   final AssessmentTypeRepository assessmentTypeRepository) {
+        this.gradeRepository = gradeRepository;
+        this.subjectRepository = subjectRepository;
+        this.typeRepository = assessmentTypeRepository;
+    }
+
+    public Assessment parse(CSVRecord record) throws IllegalArgumentException {
```
rcordeau: This parsing function will be cleaned up on the next iteration to include better refactoring, more aligned with how the item parser validates, and error gathering versus failing on the first error.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultItemParser.java
```diff
@@ -0,0 +1,77 @@
...
+                             final MathPracticeRepository mathPracticeRepository) {
+        this.claimRepository = claimRepository;
+        this.targetRepository = targetRepository;
+        this.depthOfKnowledgeRepository = depthOfKnowledgeRepository;
+        this.mathPracticeRepository = mathPracticeRepository;
+    }
+
+    //TODO: Change the way parsing is done to collect all errors, similar to exam service
+    @Override
+    public Item parse(CSVRecord record) throws IllegalArgumentException{
```
rcordeau: Will be gathering errors on the next iteration instead of failing on the first error.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -1,36 +1,44 @@
 package org.opentestsystem.rdw.ingest.processor.model;
 
+import com.google.common.base.Strings;
 import com.mysql.jdbc.StringUtils;
```
malaffoon: seems odd to pull in jdbc StringUtils for a data model class

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -1,36 +1,44 @@
...
-        this.subject = subject;
-        this.academicYear = academicYear;
-        this.label = label;
-        this.version = version;
+    private final Integer minScore;
+    private final Integer maxScore;
+    private final List<Integer> cutPoints;
+
+
+    private Assessment(Builder asmtBuilder) {
```
malaffoon: I don't like this pattern, it completely couples the class and the builder. Sure they are logically coupled anyway, but what is the advantage of doing it this way vs. the builder setting the values in the build() method?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -82,13 +90,16 @@ public boolean equals(Object another) {
 
     private boolean equalTo(Assessment another) {
```
malaffoon: Uhmm, why is this `equalTo()` and not `equals()`?

I'm still curious as to whether fleshing these methods out is necessary in any of our data model objects. You only need them if you expect to put independent objects in a hashing collection and you want de-duping. I really don't see that in any of our code. I guess the methods don't hurt anything and they might come in handy some day but i want to make sure we're implementing them for the right reason.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
malaffoon: I'd suggest that we shouldn't use jdbc StringUtils in a simple data model class. Guava and Apache Commons have tons of string utils we can use instead.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
malaffoon: I'd suggest this would be more readable to use checkArgument for all of these, e.g.
```java
    checkArgument(!StringUtils.isNullOrEmpty(naturalId), "Assessment naturalId cannot be null");
    checkArgument(cutPoints != null, "Assessment cut points cannot be null");
    checkArgument(minScore != null, "Assessment min score cannot be null");
    checkArgument(maxScore != null, "Assessment max score cannot be null");
```

malaffoon: If the calling code knows the builder is doing these basic validations, it doesn't have to. In this case, the caller would merrily read in the data and set values without worrying if it found, for example, minScore. Then it just calls build() and knows that the basics are okay if it doesn't throw.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -106,70 +118,89 @@ public int hashCode() {
...
-            withNaturalId(assessment.getNaturalId());
-            withGrade(assessment.getGrade());
-            withType(assessment.getType());
-            withSubject(assessment.getSubject());
-            withAcademicYear(assessment.getAcademicYear());
-            withLabel(assessment.getLabel());
-            withName(assessment.getName());
-            withVersion(assessment.getVersion());
+        public Builder cutPoints(final List<Integer> cutPoints) {
+            this.cutPoints = cutPoints;
```
malaffoon: This should be copying the collection into a new list. Once you do that you can loosen up the parameter to be just Collection or Iterable.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
```diff
@@ -0,0 +1,15 @@
+package org.opentestsystem.rdw.ingest.processor.repository;
+
+/**
+ * Repository for dealing with assessment types
+ */
+public interface AssessmentTypeRepository {
+    /**
+     * Finds an id of the type by its code
+     *
+     * @param code the code of the target
```
malaffoon: This could be better worded, and it's nice to give an example, e.g.
```java
     * @param code the assessment type code, e.g. "ica"
```

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
```diff
@@ -0,0 +1,17 @@
...
+public interface DepthOfKnowledgeRepository {
+
+    /**
+     * Finds an level of the target by its code
+     *
+     * @param level the level of the target
+     * @return the level if it exists
+     * @throws IllegalArgumentException is the given level is not found
+     */
+    int findByLevel(int level);
```
malaffoon: So this is returning the value you pass in, if that value is in the database? If so, i suggest you make it:
```java
    boolean exists(int level);
```

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentRepositoryImpl.java
malaffoon: nit: i usually throw an exception in methods like this, so it doesn't accidentally stay this way forever, e.g. 
```java
    // TODO - implement this
    throw new UnsupportedOperationException();
```
But, if you are immediately working on this stuff i guess it isn't necessary.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -0,0 +1,95 @@
...
+    private String toTypeCode(final String value) throws IllegalArgumentException {
+
+        if ("summative".equalsIgnoreCase(value)) return "sum";
+        if ("ica".equalsIgnoreCase(value) || "iab".equalsIgnoreCase(value)) return value.toLowerCase();
+
+        throw new IllegalArgumentException("invalid typeCode [" + value + "]");
+    }
+
+    private String toGradeCode(final String value) {
+        if (value.length() == 1) return "0" + value;
```
malaffoon: return value.length() == 1 ? ("0" + value) : value;

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -1,36 +1,44 @@
...
-    private final String type;
-    private final String subject;
-    private final int academicYear;
     private final String naturalId;
     private final String name;
+    private final int gradeId;
     private final String label;
+    private final int typeId;
+    private final int subjectId;
+    private final int academicYear;
```
agorina: This came up yesterday: TRT returns the last number in the school year. For Ex.: 2016-2017, the year is presented as 2017.  I think we need to align the values and be consistent. Could you please make sure that this value works the same way as TRT?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: So I am using this object in a HashSet. You could argue that I could de-dup a different way, yes, but is there harm in having these methods?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Ok, that's an interesting idea. After writing it, I was considering changing it to a varargs

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
rcordeau: yeah - bad comment..

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
rcordeau: I thought about that as well, but it makes using the call inline easier, without having to check the boolean.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
rcordeau: fair

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+            return cutPointArray;
+        } catch (NumberFormatException | NullPointerException ex) {
+            throw new IllegalArgumentException("invalid cut points " + Arrays.toString(cutPoints));
+        }
+    }
+
+    private int parseScaleScore(String value) {
+        try {
+             return (int) Double.parseDouble(value);
```
rcordeau: Well, I wasn't sure about preserving the score - the examples I have seen have all been really 4 digit numbers represented in scientific notation.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: I know you don't. I think we should discuss it as a team to decide what our common pattern will be.  What I have seen so far is the object being created accepts parameters, which you don't like either. Another option is using something like Immutables to auto generate the code...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: I am doing validation in the service layer against the database to validate if certain incoming values are legitimate or not.  Since I am doing that there, perhaps it isn't needed at all in the model.

PR #87, opened 2017-04-10
Happyish Path (minus s3 archive)

rcordeau:
Data validation done. Error collection mechanism implemented. Inserting valid file to database. Need to save to s3 still.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/87)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -120,27 +127,19 @@ public int hashCode() {
...
-        private int subjectId;
+        private Integer gradeId;
+        private Integer typeId;
+        private Integer subjectId;
         //Not checked against database, but required to exist
-        private List<Integer> cutPoints;
-        private Integer minScore;
-        private Integer maxScore;
+        private List<Double> cutPoints;
+        private Double minScore;
```
malaffoon: It doesn't make sense to me that the builder uses `Double` while the class uses `double`. I can think of one justification but wouldn't it be easier to make the builder use `double`?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -1,22 +1,17 @@
...
  * This is a representation of metadata for an item from the Tabulator processing of an assessment package
  */
 public class Item {
     private final String naturalId;
     private final int claimId;
     private final int targetId;
-    private final float difficulty;
-    private final int dokId;
+    private final double difficulty;
+    private final int dokLevel;
```
malaffoon: I guess it doesn't really matter but SB has stated *many* times that the DoK values are *not* levels, they are more like categories. If you want an explicit distinction from dokId perhaps `dokValue`?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
```diff
@@ -7,9 +7,9 @@
...
      * Finds an id of the type by its code
      *
-     * @param code the code of the target
+     * @param code the assessment type code, e.g. "ica"
      * @return the id
-     * @throws IllegalArgumentException is the given code is not found
+     * @throws IllegalArgumentException if the given code is not found
      */
-    int findIdByCode(String code);
+    Integer findIdByCode(String code);
```
malaffoon: This change seems inconsistent with the javadocs. If this method is going to through if the code is not found, then the return type should be `int` since it can never be null. If this method is going to return null if a code is not found, then the javadocs should be changed.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Method for bulk insertion of an Assessment Package
+     *
+     * @param assessments array of {@link Assessment} objects
+     * @param items       array of {@link Item} objects
+     */
+    @Override
+    @Transactional
+    public void createPackage(final Assessment[] assessments, final Item[] items) {
```
malaffoon: I would prefer to see java collections used here instead of arrays. Just good practice. Depending on how it is used in the helper methods it could probably be an `Iterable`.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
```diff
@@ -0,0 +1,107 @@
...
+       bulkCreateAssessments(assessments);
+       bulkCreateScores(assessments);
+       bulkCreateItems(items);
+    }
+
+    private void bulkCreateAssessments(final Assessment[] assessments) {
+        if (assessments.length == 0) return;
+
+        final MapSqlParameterSource[] batchParameters = new MapSqlParameterSource[assessments.length];
+        for (int i = 0; i < assessments.length; i++) {
```
malaffoon: If you find yourself writing an indexed for loop in java, something is probably wrong.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentTypeRepositoryImpl.java
```diff
@@ -24,13 +24,21 @@
...
+     */
     @Override
     @Cacheable(value = "asmtType")
-    public int findIdByCode(final String code) {
+    public Integer findIdByCode(final String code) {
         try {
             return jdbcTemplate.queryForObject(findIdByCodeSql, new MapSqlParameterSource("code", code), Integer.class);
-        } catch (final EmptyResultDataAccessException ignore) {
-            throw new IllegalArgumentException("unknown type code [" + code + "]");
+        } catch(EmptyResultDataAccessException ignore) {
```
malaffoon: So this answers my question above and the java docs need to be updated.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -35,35 +37,57 @@
...
         try {
-            Iterable<CSVRecord> records = CSVFormat.RFC4180.withFirstRecordAsHeader().parse(new StringReader(assessment));
-            //Pass in a container for errors - keeping track of what line the errors are on.
-            //If there are errors in the container, don't insert anything into the database
+            errorCollector.reset();
+            Iterable<CSVRecord> records = CSVFormat.RFC4180.withFirstRecordAsHeader().parse(new StringReader(assessmentPackage));
+
+
             //Throw bad data import exception
             for(CSVRecord record : records) {
```
malaffoon: please make whitespace consistent; i think we generally are doing `for (...`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -20,76 +22,102 @@
...
+    public Assessment parse(CSVRecord record) {
+
         Assessment.Builder assessmentBuilder = new Assessment.Builder();
         assessmentBuilder
-                .naturalId(record.get("AssessmentId"))
-                .name(record.get("AssessmentName"))
-                .academicYear(Integer.parseInt(record.get("AcademicYear").split("-")[0]))
-                .label(record.get("AssessmentLabel"))
-                .version(record.get("AssessmentVersion"));
+                .naturalId(parserHelper.basicValidation(ASSESSMENT_ID, record.get(ASSESSMENT_ID).trim(),
```
malaffoon: static import of `basicValidation` would make this slightly more readable

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -20,76 +22,102 @@
...
 
-    public Assessment parse(CSVRecord record) throws IllegalArgumentException {
+    //This assumes that the column names actually exist. If they don't then an exception will be thrown.
+
+    /**
+     * Parse function to validate and extract an {@link Assessment}. Any errors found are gathered in a central error collector.
+     * @param record - The {@link CSVRecord} to parse
+     * @return {@link Assessment} if successfully validated and parsed, null otherwise
+     */
+    public Assessment parse(CSVRecord record) {
```
malaffoon: `final`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -20,76 +22,102 @@
...
-                record.get("ScaledHigh2"),
-                record.get("ScaledHigh3")));
+        assessmentBuilder.minScore(parserHelper.basicValidation(MIN_SCORE, record.get(MIN_SCORE).trim(), Doubles::tryParse));
 
-        assessmentBuilder.minScore(parseScaleScore(record.get("ScaledLow1")));
-        assessmentBuilder.maxScore(parseScaleScore(record.get("ScaledHigh4")));
+        assessmentBuilder.maxScore(parserHelper.basicValidation(MAX_SCORE, record.get(MAX_SCORE).trim(), Doubles::tryParse));
 
-        return assessmentBuilder.build();
+        if(errorCollector.isEmpty()) return assessmentBuilder.build();
```
malaffoon: Shouldn't this have been noticed earlier (sorry, i should pull the code down to see the whole flow)? If it really is correct, perhaps a ternary operator would make this more readable, `return errorCollector.isEmpty() ? assessmentBuilder.build() : null;`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -20,76 +22,102 @@
...
 
-    private int parseScaleScore(String value) {
-        try {
-             return (int) Double.parseDouble(value);
-        } catch (NumberFormatException | NullPointerException ex) {
-            throw new IllegalArgumentException("invalid scale score  [" + value + "]" );
-        }
+    private String toAcademicYear(final String value) {
+        String[] split = value.split("-");
+        if(split.length == 1) return split[0]; //nothing to split, so return the value
```
malaffoon: whitespace

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultItemParser.java
```diff
@@ -22,56 +24,65 @@
     private final TargetRepository targetRepository;
     private final DepthOfKnowledgeRepository depthOfKnowledgeRepository;
     private final MathPracticeRepository mathPracticeRepository;
+    private final ParserHelper parserHelper;
+    private final DataElementErrorCollector errorCollector;
+    private static final String ITEM_CLAIM = "Claim";
```
malaffoon: I don't suppose i'll ever win this argument but, since these are locally used constants (i.e. not part of an interface or anything), can we use camel-case so they aren't so glaring and annoying to type? i.e. `ItemClaim` instead of `ITEM_CLAIM`.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/utils/ParserHelper.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private final DataElementErrorCollector elementErrorCollector;
+
+    @Autowired
+    public ParserHelper(DataElementErrorCollector elementErrorCollector) {
+        this.elementErrorCollector = elementErrorCollector;
+    }
+
+    //Basic validation for types (integer, float, double, empty strings, etc), that don't require a query against the database
+    public  <T> T basicValidation (String name, String value, Function<String, T> fx) {
```
malaffoon: `final`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/utils/ParserHelper.java
```diff
@@ -0,0 +1,62 @@
...
+        T validatedT = fx.apply(value);
+        if(validatedT == null) elementErrorCollector.addDataElementError(new DataElementError(name, value, "Invalid value"));
```
malaffoon: whitespace

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -22,9 +17,9 @@
     private final int subjectId;
     private final int academicYear;
     private final String version;
-    private final Integer minScore;
-    private final Integer maxScore;
-    private final List<Integer> cutPoints;
+    private final double minScore;
```
agorina: Any reason this is a lower case double?  Will this convert Nulls to zeros silently?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentPackageRepository.java
```diff
@@ -0,0 +1,18 @@
...
+ * AssessmentPackageRepository - Interface for bulk inserting an assessment "package", which is a
+ * collection of Assessments and corresponding Items
+ */
+public interface AssessmentPackageRepository {
+    /**
+     * Method for bulk insertion of an Assessment Package
+     * @param assessments array of {@link Assessment} objects
+     * @param items array of {@link Item} objects
+     */
+    void createPackage(final Assessment[] assessments, final Item[] items);
```
agorina: Shouldn't *items* belong to *assessment*s?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+        }
+        jdbcTemplate.batchUpdate(sqlAssessmentCreate, batchParameters);
+    }
+
+    private void bulkCreateItems(final Item[] items) {
+        if (items.length == 0) return;
+
+        final MapSqlParameterSource[] batchParameters = new MapSqlParameterSource[items.length];
+        for (int i = 0; i < items.length; i++) {
```
agorina: Is there a corresponding RDW_Schema change request?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Well, what I didn't want was a default value of 0 for the object. The idea was that if there is nothing ("") in the value for the minScore, then it would be null, and the actual object wouldn't be allowed to be constructed - it would throw a NullPointerException when trying to build.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
rcordeau: Yeah I can rename it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentTypeRepository.java
rcordeau: Yeah, sorry. The Javadoc needs to be changed.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+            batchParameters[i] = new MapSqlParameterSource()
+                    .addValue("claimId", items[i].getClaimCodeId())
+                    .addValue("targetId", items[i].getTargetId())
+                    .addValue("itemId", items[i].getNaturalId())
+                    .addValue("asmtId", items[i].getAssessmentNaturalId())
```
agorina: Does this assume that we will have to relate items to an assessment by a natural ID? 
If this is the case, then I really do not agree. 
When a TRT is ingested, one of the requirement is to validate the items against the package items. We would not want to join by a natural id at that time.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
rcordeau: Yeah, I wasn't happy about passing arrays either. I'll modify.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
rcordeau: Yeah, I agree. But the problem stems from the fact that the batch parameters array is an array. I suppose I can wrap it and create an iterable.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentTypeRepositoryImpl.java
rcordeau: yep

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
rcordeau: sigh

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
rcordeau: dammit

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
rcordeau: I shall run the IntelliJ reformatter on all code.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
rcordeau: Yes, this is really correct, and I'll use the ternary operator.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
rcordeau: changed to return split.length == 0 ? value: split[split.length -1];

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultItemParser.java
rcordeau: Sorry, I don't think you'll win this argument. :-) They are meant to be glaring.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/utils/ParserHelper.java
rcordeau: fixed. I'll go through and make sure I have final in all places where it is necessary.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AssessmentPackageRepository.java
rcordeau: I thought about that as well as I went through this. I originally had Items as a collection in the Assessment object. Would have made this code easier. I might revisit and change it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
rcordeau: Yes, and I completely forgot to submit it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryImpl.java
rcordeau: We relate to the database generated id, not the natural_id. I have it as a parameter to the query because the query has a nested select statement to get the correct id. Now, had I had Items as part of Assessment, then I could have created the assessment, and then bulk inserted the items with the returned id, as I saw you did in the exam processor. I might revisit, but the sql works, and MySQL will cache the sub-select results for getting the asmtId from the asmt.natural_id

PR #88, opened 2017-04-13
pull CachingTest util into common

malaffoon:
This eliminates the code duplication.


common/build.gradle
```diff
@@ -23,4 +23,18 @@ dependencies {
     compile "com.google.code.gson:gson:2.8.0"
 
     testCompile 'org.springframework:spring-test'
+    testCompile 'org.springframework.boot:spring-boot-starter-test'
+}
+
+task testsJar(type: Jar, dependsOn: testClasses) {
+    classifier = 'tests'
```
malaffoon: Is this right? I have this sniggling thought that it should be `test` not `tests` but i can't remember?

PR #89, opened 2017-04-13
adding in missed max_points for Item

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/89)

### DIFF
--- a/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
+++ b/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
@@ -15,7 +15,7 @@ public class Item {
     private Integer mathPractice;
     private Boolean allowCalculator;
     private String assessmentNaturalId;
-    private final int maxPoints;
+    private final double maxPoints;
 
     private Item(@NotNull Builder builder) {
         this.naturalId = builder.naturalId;
@@ -65,7 +65,7 @@ public class Item {
         return assessmentNaturalId;
     }
 
-    public int getMaxPoints() {
+    public double getMaxPoints() {
         return maxPoints;
     }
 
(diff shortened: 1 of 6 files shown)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -81,6 +87,7 @@ private boolean equalTo(Item another) {
                 targetId == another.targetId &&
                 difficulty == another.difficulty &&
                 dok == another.dok &&
+                maxPoints == another.maxPoints &&
                 assessmentNaturalId.equals(another.assessmentNaturalId) &&
                 mathPractice.equals(another.mathPractice) &&
```
malaffoon: unrelated ... won't this throw if mathPractice is null? Wouldn't it be better to use Objects.equals()?

rcordeau: mathPractice can't be null for a fully constructed Item object since it is using a native type.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -25,6 +26,7 @@ private Item(@NotNull Builder builder) {
         this.mathPractice = builder.mathPractice;
         this.allowCalculator = builder.allowCalculator;
         this.assessmentNaturalId = builder.assessmentNaturalId;
+        this.maxPoints = builder.maxPoints;
```
malaffoon: I don't see a test that demonstrates the NPE if maxPoints isn't set in the builder.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
agorina: According to email from Brandt,  Mathematical Practice standards are optional. I just resent you an email.

agorina:
I am surprised to see that there is not changes to the Repo integration tests. I would expect it needed.

rcordeau:
That one was missing.

PR #93, opened 2017-04-14
Added school_year to exams.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
```diff
@@ -15,7 +15,7 @@
     private Date completedAt;
     private long assessmentId;
     private String asmtVersion;
-    private Integer schoolYear;
+    private Long schoolYear;
```
malaffoon: integer wasn't big enough??

agorina: Apparently for TRT integer is not good enough. It returns it as *long*. I got tired of fighting the conversion/casting errors.

PR #96, opened 2017-04-20
Adding a configuration for cloud settings in import-service to fix running integration tests

rcordeau:
Due to an automatic configuration of spring cloud when running on an ec2 instance, we have to disable the auto stack otherwise we get a 'stack does not exist (Service: AmazonCloudFormation; Status Code: 400;...' error

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/96)

### DIFF
--- a/import-service/src/main/resources/application.yml
+++ b/import-service/src/main/resources/application.yml
@@ -11,6 +11,8 @@ cloud:
     region:
       auto: false
       static: us-west-2
+    stack:
+      auto: false
 
 spring:
   datasource:


import-service/src/main/resources/application.yml
```diff
@@ -11,6 +11,8 @@ cloud:
     region:
       auto: false
       static: us-west-2
+    stack:
+      auto: false
```
rcordeau: Without this, we would get Spring Bean exceptions caused by  com.amazonaws.AmazonServiceException

malaffoon: Hmm. So i have that locally because of my kops work. So you are using ec2 for builds?

rcordeau: I am reusing the agent and VM that we are using to do legacy builds. Instead of using an ec2 VM to do builds, what else do you propose that we use?

PR #100, opened 2017-04-22
Fixed RepositoryIT failure when DB is not empty.

agorina:
While I am not sure I agree that it is a good idea to run IT tests on the Db with the data, since it is not hard to fix, I am okay with this approach for now.

The reason I am not fully buying into this approach: if you do not know what is in the DB, you can run into an unexpected behaviour just because of the data conflict. For example, your test may expect to create a new entry while it may be already there.  I think the better approach would be to have a separate Db for the Integration testing from the one that is used for the functional testing.

### DIFF
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AdminConditionRepositoryIT.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AdminConditionRepositoryIT.java
@@ -19,9 +19,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 @Transactional
 @CachingTest
 @Sql(statements = {
-        "INSERT INTO administration_condition (id, name) VALUES (99, 'test1');",
-        "INSERT INTO administration_condition (id, name) VALUES (88, 'test2');",
-        "INSERT INTO administration_condition (id, name) VALUES (55, 'test55');"
+        "INSERT INTO administration_condition (id, name) VALUES (-99, 'test1');",
+        "INSERT INTO administration_condition (id, name) VALUES (-88, 'test2');",
+        "INSERT INTO administration_condition (id, name) VALUES (-55, 'test55');"
 })
 public class AdminConditionRepositoryIT {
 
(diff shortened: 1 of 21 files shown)


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AdminConditionRepositoryIT.java
```diff
@@ -19,9 +19,9 @@
 @Transactional
 @CachingTest
 @Sql(statements = {
-        "INSERT INTO administration_condition (id, name) VALUES (99, 'test1');",
-        "INSERT INTO administration_condition (id, name) VALUES (88, 'test2');",
-        "INSERT INTO administration_condition (id, name) VALUES (55, 'test55');"
+        "INSERT INTO administration_condition (id, name) VALUES (-99, 'test1');",
+        "INSERT INTO administration_condition (id, name) VALUES (-88, 'test2');",
+        "INSERT INTO administration_condition (id, name) VALUES (-55, 'test55');"
```
agorina: I assume that it is safer to test with the negative ids.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AssessmentPackageRepositoryIT.java
```diff
@@ -51,10 +46,6 @@
     @Autowired
     private JdbcTemplate jdbcTemplate;
 
-    private final String ASMT_WHERE = "";
-    private final String ITEM_WHERE = "";
-    private final String ASMT_SCORE_WHERE = "";
-
```
agorina: This was not used.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/AdminConditionRepositoryIT.java
malaffoon: clever

PR #102, opened 2017-04-25
Feature/dwr 206 csv report

bgsavage:
Endpoint: 
http://localhost:8080/exams/imports/reconciliation

I have tested this where it returns in 138ms and 451.81 KB for 6417 records.  I wouldn't expect this response to ever return more than 5mb, and would usually be around 3mb.

I loaded more data after the data generator was fixed.

I've now tested where it returns in 386ms and 1.61MB for 19417 records.


common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
```diff
@@ -0,0 +1,56 @@
+package org.opentestsystem.rdw.ingest.common.model;
+
+import java.util.Date;
+
```
bgsavage: I committed this class yesterday before today's discussion about only creating a Builder when it's more than 3 params.  Should I revert the builder now that I already have it?  I suspect we that we may want to add more fields to this object later anyway.

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
```diff
@@ -118,15 +127,27 @@ public void delete(final long id) {
...
-                .addValue("status", query.getStatus() == null ? null : query.getStatus().getValue())
-                .addValue("batch", query.getBatch())
-                .addValue("creator", query.getCreator())
-                .addValue("before", query.getBefore() == null ? null : Timestamp.from(query.getBefore()))
-                .addValue("after", query.getAfter() == null ? null : Timestamp.from(query.getAfter()));
+        final MapSqlParameterSource parameterSource = mapQueryToSqlParameterSource(query);
         try {
-            return jdbcTemplate.query(sqlImportFindBy, parameterSource, new RdwImportRowMapper());
+            String sqlQuery = sqlImportFindAll + " " + sqlImportWhereClause;
+            return jdbcTemplate.query(sqlQuery, parameterSource, new RdwImportRowMapper());
```
bgsavage: I wanted to reuse the RdwImportQuery functionality so I created a where clause that can be used by all 3 queries (the 2 exam queries and the import query)

import-service/src/main/resources/application.sql.yml
```diff
@@ -13,21 +13,40 @@ sql:
...
       SELECT (CASE WHEN EXISTS (SELECT 1 FROM import WHERE id=:id) THEN true ELSE false END) exist
 
-    findBy: >-
-      SELECT * FROM import WHERE (:content IS NULL OR content=:content)
-          AND (:status IS NULL OR status = :status)
-          AND (:batch IS NULL OR batch = :batch)
-          AND (:creator IS NULL OR creator = :creator)
-          AND (:before IS NULL OR created < :before)
-          AND (:after IS NULL OR created > :after)
-          LIMIT 100
```
bgsavage: This got broken up to a findAll and a whereClause

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
```diff
@@ -0,0 +1,56 @@
...
+    public String getStudentId() {
+        return studentId;
+    }
+
+    public String getAssessmentNaturalId() {
+        return assessmentNaturalId;
+    }
+
+    public Date getCompletedAt() {
+        return completedAt;
```
esotrope: probably need defensive copy here since date is mutable

bgsavage: Ok, should I make a copy and return that?  And I should do the same when I'm building, yeah?

esotrope: yeah both would be safest

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
agorina: I really do not see what you are buying by doing this. Yeah, you have to repeat the code in the SQL, but this is the nature of the SQL. 

Instead you ended up concatenating three times in this code.

Additionally it defeats one of the main purpose I listed today why I wanted to have SQL in yml: a DBA can look at see at the query to determine what indexes may help. But breaking it this way, it is not clear where the 'WHERE" is being used.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
```diff
@@ -0,0 +1,56 @@
+package org.opentestsystem.rdw.ingest.common.model;
+
+import java.util.Date;
+
+public class RdwImportContent {
```
agorina: Is this a subset of the RwdImport class? Why a different class?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -69,18 +72,26 @@ public RdwImportResource getImport(@PathVariable Long id) {
...
+    public RdwImportReconciliation getImportReconciliation(final RdwImportQuery inputQuery) {
+        final RdwImportQuery query = inputQuery.isEmpty()
+                ? RdwImportQuery
+                    .builder()
+                    .after(Instant.now().minus(1, ChronoUnit.DAYS))
+                    .status(ImportStatus.PROCESSED)
+                    .content(ImportContent.EXAM)
+                    .build()
+                : inputQuery
+                    .copy()
```
agorina: Do we want to consider any restrictions here? I am afraid if somebody asks for the entire data set?

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
bgsavage: I was thinking of having to maintain this same where clause in 3 places instead of just 1 should it ever need to change or be added to.  Also, it keeps the code cleaner in that I don't have to write the functionality to create the where clause 3 different times for the 3 different queries.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
bgsavage: I can always set a timespan when one isn't set, and not allow a timespan larger than that same default timespan I could set.  Does this work for you?

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
bgsavage: It's different properties based on the content of the import xml.

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
agorina: I do not see how it keeps the code cleaner? If you have a common code that builds MapSqlParameterSource, then why would you have it 3 times. What am I missing?

You are still not addressing my main concern of having queries in yml for the index analysis.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
agorina: As long as we limit the data set, I am fine. I do not want somebody to ask for 300,000,000 records.

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
bgsavage: I could use the same method that builds it, but if there's a chance for them to diverge (since they would be composed in the query themselves), then should I?  I would be introducing coupling without making it clear that there is coupling.

I could put comments in the .yml indicating which queries will use the where clause, does this work?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
bgsavage: I can set a limit on the SQL, what do you recommend?  Maybe 100,000 ?

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
agorina: I still strongly believe that they need to be changed in the yml file, but want to see what other people have to say.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
agorina: I do not think setting an arbitrary number as a limit is a solution.

This may require some design:
1. what is the goal? just a day worth of data? - is yes, then I would make it a required parameter.
2. what is the max number of records per day (assuming that this is the goal)? If this is something we can support then I would  add a count first to verified that the response is not larger than we expect. And if it is larger, return an error.
3. If we expect a large size, then I would change the approach: we can either do a paging on the back end, read data in chunk and stream them as we read; or write to the file first, and then return the file. The second will requires an asynch communication with the client.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
malaffoon: i'd vote for leaving the builder in

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
jtreuting: We decided to go the yml file appraoch and I do see a benefit of having the complete queries in a single statement.  I personally don't like them split up and building them back up to save duplication.  I don't think the added complexity outweighs not being able to easily see the entire query in 1 place.

malaffoon: I want the best of both worlds.

And a pony.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
malaffoon: Can we use not-a-Date? Perhaps an Instant.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
malaffoon: For immediate safety, a LIMIT 100000 would help.

Longer term i think we'll need to a pre-check on the count, and reject any request that is too big. Can discuss in parking lot if you'd like.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -69,18 +72,26 @@ public RdwImportResource getImport(@PathVariable Long id) {
...
 
-    @GetMapping(value="/imports", headers="Accept=text/csv")
+    @GetMapping(value="/imports/reconciliation", headers="Accept=text/csv")
     @ResponseStatus(HttpStatus.OK)
     @ResponseBody
-    public RdwImportCsv getImportsAsCsv(final RdwImportQuery query) {
-        if (query.isEmpty()) {
-            throw new IllegalArgumentException("at least one query parameter must be specified");
-        }
+    public RdwImportReconciliation getImportReconciliation(final RdwImportQuery inputQuery) {
```
malaffoon: Does the converter/mapping turn this into a stream? It's always a good idea to return big files using streams.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -69,18 +72,26 @@ public RdwImportResource getImport(@PathVariable Long id) {
...
+        final RdwImportQuery query = inputQuery.isEmpty()
+                ? RdwImportQuery
+                    .builder()
+                    .after(Instant.now().minus(1, ChronoUnit.DAYS))
```
jtreuting: Have we thought through if we are storing the Instant as UTC or as whatever they pass us and then are they querying for a days worth of results by UTC or whatever they pass us?  We also will need to see if we auto-generate any dates and if they will align with the dates we receive in TRTs.

malaffoon: We are storing the instant as UTC. 
TRTs either specify a timezone or we assume UTC.
But these timestamps are generated by the server. And the query logically support "last hour", "last day" so it shouldn't be a huge issue. But, yeah, we'll have to be sure to document *exactly* how clients can specify time ranges if they need them.

jtreuting: I thought `Instant.now()` does not have a timezone, so will be the time in the current server timezone (without the timezone included).

Wouldn't we need to do something like: 

```
ZonedDateTime utc = ZonedDateTime.now(ZoneOffset.UTC);
utc.toInstant().minus(1, ChronoUnit.DAYS);
```

malaffoon: No need to get a ZonedDateTime involved for this. RdwImport has a field, created, that is an Instant. It is set by calling now() on the server. So if the client wants to get stuff from the last day it will work down to:
```java
Instant.now().minus(1, ChronoUnit.DAYS)
```
Of course, if you then want to display the created dates to the user you'll need to convert to their timezone somehow.

jtreuting: Got it. I was asking because you said it was stored as UTC and I don't believe Instant.now() is UTC unless the server timezone is UTC.  Which in this case is fine if the server generates it and then is the one providing the range when querying.

The only dependency it creates is that all servers need to be in the same timezone.  If each server did a true UTC now when storing and querying it wouldn't matter what timezone they were set in.  It also wouldn't matter if we had the database create the timestamps at times using UTC_TIMESTAMP.

malaffoon: Yeah, that's what i get for rushing a review before i leave. My point is that, as long as the server does the time manipulation, we're relatively safe.

Your point is valid though that we should replace Instant.now() with a utility method that produces UTC now and use that everywhere.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
bgsavage: I believe this is what I'm doing on the message converter unless I'm misunderstanding how a message converter works.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
bgsavage: An Instant.utcNow sure would be nice :) But yes, I agree this should be able run on any env and produce consistent times.  I'll create a utility function and replace where we use Instant.now().  What do you think, maybe a InstantUtility.java with a static utcNow function?

malaffoon: Hmm. I did a little test and i'm not sure i can explain the results. I verified that my system has pacific time as the default timezone and then i did this (at 8:25 in the morning):
```
        System.out.println(Instant.now());
        System.out.println(ZonedDateTime.now(ZoneOffset.UTC).toInstant());
        System.out.println(ZonedDateTime.now().toInstant());
        System.out.println(ZonedDateTime.now(ZoneId.systemDefault()).toInstant());
        System.out.println(LocalDateTime.now().toInstant(ZoneOffset.UTC));
```
and it produced this ...
```
2017-04-26T15:25:17.138Z
2017-04-26T15:25:17.194Z
2017-04-26T15:25:17.204Z
2017-04-26T15:25:17.206Z
2017-04-26T08:25:17.207Z
```
On the upside, Instant.now() appears to produce a UTC now, and the documentation actually supports that. On the downside how can ZDT.now produce the same instant for different time zones?

bgsavage: Is this a spring or jre trick here setting the systemDefault to utc (regardless of what you set on your machine)?

malaffoon: @bgsavage, based on my experiment there is no need for a utility function. However, i would suggest taking advantage of the RdwImportQuery builder and passing in an offset instead of an actual time, i.e.:
```
.builder().after("-PT24H")
```

jtreuting: Ok figured out the issue we had with `Instant.now()` ... you're right it is UTC.  We had a problem when we took the Instant and constructed a new `java.sql.TimeStamp` and it would then convert it to whatever timezone the JVM is running in.  So we had to work around that issue.  It caused a lot of confusion at the time (and obviously still does for me).

jtreuting: What we had to do was map an Instant to a Timestamp when doing an insert:

`new Timestamp(time.getMillis())`

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
bgsavage: Can you review the latest commit?  Just want to make sure I'm converting the SQL date to an Instant correctly.  (It works, but not sure what best practice is for this.)

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
```diff
@@ -0,0 +1,54 @@
+package org.opentestsystem.rdw.ingest.common.model;
+
+import java.time.Instant;
+
+public class RdwImportContent {
+    private String studentId;
+    private String assessmentNaturalId;
+    private Instant completedAt;
```
malaffoon: Sigh, i'm sorry @bgsavage i was confused about this. I thought this was the import created field, this is really the complete date (time?) for the test. Let me poke around and get back to you on this. If you need/want to check-in as is and we can revisit that's fine too.

bgsavage: This is the sql iab_exam.completed_at and exam.completed_at, which is a mysql date in the db.

And no worries, I might go ahead and merge this and update this later with what you recommend.

malaffoon: Yeah, we don't have a good answer for this yet, i don't think anybody has looked at it closely. @agorina you haven't vetted the date handling yet have you? 

In the TRT model, this date is stored as an Instant. We parse the value from the XML, looking for a timezone indicator and assuming UTC if we don't find one. Therefore the date value is an Instant (with implied UTC). 

During processing that Instant is converted directly to a java.util.Date. That Date is then written to SQL using direct parameter setting. So it is possible (likely?) that SQL is shifting the date based on the timezone of the client. Which means you should really just be pulling it directly out without any manipulation. You know, like your initial commit had it (sorry again).

We should dig around and see if SQL is shifting the value.

PR #103, opened 2017-04-25
reduce number of connections in pool

malaffoon:
After my Aurora experiments i felt that these services do not need a big pool of connections lying around. They aren't really web services (yeah, they have API end-points but their primary use case is not serving thousands of users). They don't spawn tons of worker threads; parallelization will happen via multiple instances of the apps, not internal threading.

The default values for pool are 10/100 (initial:10, max:100). To get the absolute optimum settings we'll need to monitor the apps under production-like pressure.

Thoughts?

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -22,3 +22,6 @@ spring:
     validationQuery: SELECT 1
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
+    maxActive: 20
+    initialSize: 4
+    maxWait: 10000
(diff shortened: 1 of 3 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -21,4 +21,7 @@ spring:
     testWhileIdle: true
     validationQuery: SELECT 1
     driverClassName: com.mysql.jdbc.Driver
-    initialize: false
\ No newline at end of file
+    initialize: false
+    maxActive: 20
+    initialSize: 4
+    maxWait: 10000
```
malaffoon: the default wait is 30000 (30s) which i've always disliked; if things are so bad that a connection doesn't free up in 10s, please throw an exception

PR #105, opened 2017-04-26
add resubmit hook

malaffoon:
This is just a quick change so i can easily resubmit imports that get stuck in ACCEPTED state. That will (currently) happen under two conditions (both will be fixed but for now...):
1. There are no processors running so the rabbit queue has not been created
2. A processor was brought online with items in the queue; it seems to grab a couple and do nothing with them.

This hook can also be used to resubmit in the future when exams are "parked" because the assessment packages aren't available yet.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
@@ -7,14 +7,18 @@ import org.opentestsystem.rdw.ingest.common.model.ImportStatus;
 import org.opentestsystem.rdw.ingest.common.model.RdwImport;
 import org.opentestsystem.rdw.ingest.common.model.RdwImportQuery;
 import org.opentestsystem.rdw.ingest.repository.RdwImportRepository;
+import org.opentestsystem.rdw.utils.TenancyChain;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+
+import static com.google.common.collect.Maps.newHashMap;
 
 /**
  * Default implementation of ImportService
@@ -22,6 +26,10 @@ import java.util.Properties;
 @Service
 class DefaultImportService implements ImportService {
     private static final Logger logger = LoggerFactory.getLogger(DefaultImportService.class);
+
+    private static final String ContentTypeProperty = "content-type";
+    private static final String UsernameProperty = "username";
+    private static final String TenancyChainProperty = "tenancy-chain";
 
     private final RdwImportRepository repository;
     private final ArchiveService archiveService;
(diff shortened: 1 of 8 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -81,4 +87,36 @@ public RdwImport importContent(final RdwUser user, final String payload, final I
...
+        if (query == null || query.isEmpty()) {
+            query = RdwImportQuery.builder()
+                    .status(ImportStatus.ACCEPTED)
+                    .after("-PT24H")
+                    .before("-PT1H")
+                    .build();
+        }
+        final Map<String, TenancyChain> userTenancyChains = newHashMap();
+        long count = 0;
+        for (final RdwImport rdwImport : repository.findBy(query)) {
```
malaffoon: In the final version we may want to make all this asynchronous and handle the query in chunks or something.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -81,4 +87,36 @@ public RdwImport importContent(final RdwUser user, final String payload, final I
...
+            final Properties properties = archiveService.readProperties(location);
+            final String contentType = properties.getProperty(ContentTypeProperty);
+            final String username = properties.getProperty(UsernameProperty);
+            // TODO - fetch current tenancy chain for user from authz service
+            // TODO   skipping for now because authz is changing Real Soon Now ...
+            final TenancyChain tenancyChain = userTenancyChains.computeIfAbsent(username, k ->
+                    TenancyChain.fromString(properties.getProperty(TenancyChainProperty)));
+
+            source.submitContent(username, tenancyChain, payload, content, contentType, rdwImport.getId());
+            ++count;
```
bgsavage: do we need to keep count, or can we just return the length of the list coming from repository.findby ?

malaffoon: You're right, for now we could return the length of the list. The count is left over from some chunking experimentation i was doing.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -89,4 +90,17 @@ public RdwImportCsv getImportsAsCsv(final RdwImportQuery query) {
...
+    @PostMapping("/imports/resubmit")
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public ResponseEntity<Long> resubmitImports(final RdwImportQuery query) {
+        // if query is empty, force to be ACCEPTED
+        return ResponseEntity.ok(service.resubmitImports(query.isEmpty() ?
+                RdwImportQuery.builder().content(ImportContent.EXAM).status(ImportStatus.ACCEPTED).build() :
+                ensureExamQuery(query)));
+    }
+
+    private static RdwImportQuery ensureExamQuery(final RdwImportQuery query) {
```
bgsavage: I think the naming of this confused me.  I asked "ensure Exam query what?".  Maybe ensureQueryIncludesExam is clearer?

malaffoon: will change

PR #106, opened 2017-04-26
where clause example

malaffoon:
With absolutely no changes to the code this works right now ... weigh in!

To be clear, this snippet is not an endorsement of the approach, merely a demonstration. I will not be merging this PR: at best it will cause conflicts, at worst we all hate it  ;)


agorina:
What Jeff said :)

PR #107, opened 2017-04-26
timestamp cleanup

malaffoon:
DWR-329

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
@@ -1,6 +1,6 @@
 package org.opentestsystem.rdw.ingest.processor.model;
 
-import java.util.Date;
+import java.time.Instant;
 import java.util.List;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -12,7 +12,7 @@ import static com.google.common.collect.Lists.newArrayList;
  */
 public abstract class AnyExam {
     //TODO: add support for accessibility accommodations
-    private Date completedAt;
+    private Instant completedAt;
     private long importId;
     private long assessmentId;
     private String asmtVersion;
@@ -35,7 +35,7 @@ public abstract class AnyExam {
         return scaleScoreStdErr;
     }
 
-    public Date getCompletedAt() {
+    public Instant getCompletedAt() {
         return completedAt;
     }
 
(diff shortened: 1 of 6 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
```diff
@@ -12,7 +12,7 @@
  */
 public abstract class AnyExam {
     //TODO: add support for accessibility accommodations
-    private Date completedAt;
+    private Instant completedAt;
```
malaffoon: This value needs to be a full datetime. I am recommending that all such fields be Instant's.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
```diff
@@ -60,7 +61,7 @@ public long create(final Exam exam) {
                 .addValue("scale_score", exam.getScaleScore())
                 .addValue("scale_score_std_err", exam.getScaleScoreStdErr())
                 .addValue("achievement_level", exam.getAchievementLevel())
-                .addValue("completed_at", exam.getCompletedAt());
+                .addValue("completed_at", Timestamp.from(exam.getCompletedAt()));
```
malaffoon: Putting an Instant into a SQL TIMESTAMP field is easy.

bgsavage:
Does the ImportRepositoryImpl need to be updated as well?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
jtreuting: oops, commented before seeing this comment.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -73,7 +73,7 @@
             "administration_condition_id = 1 and " +
             "asmt_version = '345' and " +
             "asmt_id = -1 and " +
-            "completed_at = '2007-01-02' and " +
+            "completed_at = convert_tz(timestamp('2007-01-02T14:30:00Z'), '+00:00', @@session.time_zone) and " +
```
malaffoon: And here is where it gets fun. When doing native mysql client queries you have to convert timezones. Sucks. Probably means Jeff's warning will be an issue.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryImpl.java
jtreuting: If it is an issue, then the solution that we found was to do this:

`new Timestamp(exam.getCompletedAt().getMillis())`

PR #108, opened 2017-04-27
ongoing timestamp cleanup

malaffoon:
This started with verifying that the completed_at timestamp was being pulled out properly and consistently. But i tweaked a couple other things. @bgsavage please vet changes, i'll put comments inline.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
@@ -187,7 +187,7 @@ class RdwImportRepositoryImpl implements RdwImportRepository {
             return RdwImportContent.builder()
                     .studentId(rs.getString("ssid"))
                     .assessmentNaturalId(rs.getString("natural_id"))
-                    .completedAt(Instant.ofEpochMilli(rs.getDate("completed_at").getTime()))
+                    .completedAt(rs.getTimestamp("completed_at").toInstant())
                     .build();
         }
     }
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
@@ -74,6 +74,7 @@ class ExamController {
     @ResponseStatus(HttpStatus.OK)
     @ResponseBody
     public RdwImportReconciliation getImportReconciliation(final RdwImportQuery inputQuery) {
+        // if query is empty use a reasonable default of all PROCESSED EXAMS from the last 24 hours
         final RdwImportQuery query = inputQuery.isEmpty()
                 ? RdwImportQuery
                     .builder()
@@ -81,31 +82,29 @@ class ExamController {
                     .status(ImportStatus.PROCESSED)
                     .content(ImportContent.EXAM)
                     .build()
-                : inputQuery
-                    .copy()
-                    .content(ImportContent.EXAM)
-                    .build();
+                : ensureQueryForExam(inputQuery);
 
-        RdwImportReconciliation report = new RdwImportReconciliation();
-        report.setRdwImports(service.getImportContents(query));
-
-        return report;
+        return new RdwImportReconciliation(service.getImportContents(query));
     }
 
     @GetMapping("/imports/by/{group}")
     @ResponseStatus(HttpStatus.OK)
-    public ResponseEntity<String> getImportsReport(@PathVariable final String group,
-                                                   final RdwImportQuery query) {
+    public ResponseEntity<String> getImportsReport(@PathVariable final String group, final RdwImportQuery query) {
         return ResponseEntity.ok("report exam imports by " + group + " with query " + query.asParamString());
     }
 
     @PostMapping("/imports/resubmit")
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public ResponseEntity<Long> resubmitImports(final RdwImportQuery query) {
-        // if query is empty, force to be ACCEPTED
-        return ResponseEntity.ok(service.resubmitImports(query.isEmpty() ?
-                RdwImportQuery.builder().content(ImportContent.EXAM).status(ImportStatus.ACCEPTED).build() :
-                ensureQueryForExam(query)));
+    public ResponseEntity<Long> resubmitImports(final RdwImportQuery inputQuery) {
+        // if query is empty, use a reasonable default of ACCEPTED EXAMs
+        final RdwImportQuery query = inputQuery.isEmpty()
+                ? RdwImportQuery.builder()
+                        .content(ImportContent.EXAM)
+                        .status(ImportStatus.ACCEPTED)
+                        .build()
+                : ensureQueryForExam(inputQuery);
+
+        return ResponseEntity.ok(service.resubmitImports(query));
     }
 
     private static RdwImportQuery ensureQueryForExam(final RdwImportQuery query) {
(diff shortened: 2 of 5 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
```diff
@@ -187,7 +187,7 @@ public RdwImportContent mapRow(final ResultSet rs, final int rowNum) throws SQLE
             return RdwImportContent.builder()
                     .studentId(rs.getString("ssid"))
                     .assessmentNaturalId(rs.getString("natural_id"))
-                    .completedAt(Instant.ofEpochMilli(rs.getDate("completed_at").getTime()))
+                    .completedAt(rs.getTimestamp("completed_at").toInstant())
```
malaffoon: This was the main thing. The old way was truncating the time and then applying the timezone to it so `2017-02-03 12:34:00` was turning into `2017-02-03 07:00:00`. I verified the new way is properly round-tripping UTC timestamps.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -74,38 +74,37 @@ public RdwImportResource getImport(@PathVariable Long id) {
...
     @PostMapping("/imports/resubmit")
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public ResponseEntity<Long> resubmitImports(final RdwImportQuery query) {
-        // if query is empty, force to be ACCEPTED
-        return ResponseEntity.ok(service.resubmitImports(query.isEmpty() ?
-                RdwImportQuery.builder().content(ImportContent.EXAM).status(ImportStatus.ACCEPTED).build() :
-                ensureQueryForExam(query)));
+    public ResponseEntity<Long> resubmitImports(final RdwImportQuery inputQuery) {
+        // if query is empty, use a reasonable default of ACCEPTED EXAMs
+        final RdwImportQuery query = inputQuery.isEmpty()
```
malaffoon: Make the two methods look similar in their query handling.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -74,38 +74,37 @@ public RdwImportResource getImport(@PathVariable Long id) {
...
                     .content(ImportContent.EXAM)
                     .build()
-                : inputQuery
-                    .copy()
-                    .content(ImportContent.EXAM)
-                    .build();
-
-        RdwImportReconciliation report = new RdwImportReconciliation();
-        report.setRdwImports(service.getImportContents(query));
+                : ensureQueryForExam(inputQuery);
```
malaffoon: Use the new helper method.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
```diff
@@ -59,7 +59,7 @@ protected void writeInternal(RdwImportReconciliation rdwImportCsv, HttpOutputMes
 
         OutputStream bodyStream = outputMessage.getBody();
 
-        try (ICsvBeanWriter csvWriter = new CsvBeanWriter(new OutputStreamWriter(bodyStream), CsvPreference.STANDARD_PREFERENCE)) {
+        try (ICsvBeanWriter csvWriter = new CsvBeanWriter(new OutputStreamWriter(bodyStream), CsvPreference.EXCEL_PREFERENCE)) {
```
malaffoon: Question: `STANDARD_PREFERENCE` uses CRLF as the line delimiter while `EXCEL_PREFERENCE` uses LF. I'd guess that we would prefer the latter which is why i changed this. Thoughts?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
```diff
@@ -82,26 +82,12 @@ protected void writeInternal(RdwImportReconciliation rdwImportCsv, HttpOutputMes
...
     private static class ParseInstant extends CellProcessorAdaptor {
-
-        public ParseInstant() {
-            super();
-        }
-
-        public ParseInstant(CellProcessor next) {
-            super(next);
-        }
+        private final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyyMMdd").withZone(ZoneOffset.UTC);
```
malaffoon: I did this to avoid creating the formatter for every row.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportReconciliation.java
```diff
@@ -4,14 +4,17 @@
...
-        return rdwImports;
-    }
+/**
+ * A wrapper around a list of RdwImportContent; allows custom view conversion.
+ */
+class RdwImportReconciliation {
+    private final List<RdwImportContent> rdwImports;
 
-    public void setRdwImports(List<RdwImportContent> rdwImports) {
+    public RdwImportReconciliation(final List<RdwImportContent> rdwImports) {
```
malaffoon: This simplifies the calling code (basically allows inlining instead of declaring a variable and calling the setter)

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
bgsavage: Yeah this what I was very unsure of in my previous PR, but then you mentioned the column type in the data was wrong so thought to just wait until that change was complete.  This also brings us consistency with what the import dates are doing.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
bgsavage: I'm not sure which is more of what the client will want.  Do you know what their line delimiter is for their report today?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
bgsavage: That's a good idea.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
```diff
@@ -82,26 +82,12 @@ protected void writeInternal(RdwImportReconciliation rdwImportCsv, HttpOutputMes
...
 
     private static class ParseInstant extends CellProcessorAdaptor {
-
-        public ParseInstant() {
-            super();
-        }
-
-        public ParseInstant(CellProcessor next) {
-            super(next);
-        }
```
bgsavage: These constructors are recommended by the Super CSV documentation so that the calls can be chained.

https://super-csv.github.io/super-csv/examples_new_cell_processor.html
> it has a no-args constructor (for when this is the last or only processor in the chain)
> it has constructor that allows another processor to be chained afterwards (it must call super(next))

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
malaffoon: The one report i have appears to have just LF.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/RdwImportCsvMessageConverter.java
malaffoon: Oh okay. Since we aren't using the constructor, i was getting warnings and just removed them to make my file green.

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/RdwImportRepositoryImpl.java
malaffoon: Yup. I didn't know enough in the previous PR to fix it then.

PR #109, opened 2017-04-27
producer

malaffoon:
This change will force the (rabbit) durable queues to be created when messages are posted instead of waiting for a consumer to come online. Upshot is no more lost messages due to a lack of running consumers.

### DIFF
--- a/import-service/src/main/resources/application.yml
+++ b/import-service/src/main/resources/application.yml
@@ -18,6 +18,21 @@ cloud:
       instance-profile: false
 
 spring:
+  # Force the (rabbit) queues to be created when posting messages so, even if
+  # there are no consumers running, the messages will not get lost. Routing is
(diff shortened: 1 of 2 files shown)


import-service/src/main/resources/application.yml
```diff
@@ -18,6 +18,21 @@ cloud:
...
 spring:
+  # Force the (rabbit) queues to be created when posting messages so, even if
+  # there are no consumers running, the messages will not get lost. Routing is
+  # dynamic based on content but there is no way to do that here so we must
+  # explicitly declare all possible output channels here (EXAM, PACKAGE, etc.).
+  # The group names must be coordinated with the consumer application configs.
+  cloud:
+    stream:
+      bindings:
+        EXAM:
```
malaffoon: It kind of sucks that the (java code) routing is dynamic but the configuration cannot be. Oh well.

PR #110, opened 2017-04-28
POC for scheduled API calls to import-service.

bgsavage:
This is just a prototype, I don't plan to merge this at at this time, but looking for feedback on my approach to use Eureka as a means for client discovery and facilitation of cross service communication.  I'm not sure if I need Feign in there at this time, I'm still testing it out.  

Added a new task service which holds a task scheduler.  This also doubles as a Eureka server to enable cross service communication via feign.

Right now just hitting a test endpoint in the import-service


import-service/build.gradle
```diff
@@ -44,6 +44,10 @@ dependencies {
     compile 'org.opentestsystem.rdw.common:rdw-common-utils'
     compile 'mysql:mysql-connector-java'
 
+    // enables to client discovery through Eureka
```
chdietz: your thinking about eureka for angular client to discover API server or internal service to find other services?  This is new to me sounds like it is not for edge services.

chdietz: ok, I see further down the discoveryClient

chdietz: @malaffoon does k8s provide discovery similar to this?

task-service/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+apply plugin: 'org.springframework.boot'
+
+dependencies {
+    // handy for creating metadata with @ConfigurationProperties
+    optional "org.springframework.boot:spring-boot-configuration-processor"
+
+    compile 'org.springframework.boot:spring-boot-starter-actuator'
+
+    compile 'org.springframework.cloud:spring-cloud-starter-config'
+    compile 'org.springframework.cloud:spring-cloud-stream'
```
malaffoon: doubt you need cloud-stream

bgsavage: probably not, I just used another project in ingest to form a gradle build template and removed what I know I didn't need.

malaffoon:
I don't see a huge value-add with Eureka here. Near as i can tell all it is doing is making `rdw-ingest-import-service` resolve to something. But, in an orchestrated environment (either docker-compose or k8s) that is something that can be made available. For example, look at the configuration service or rabbitmq. We could either have a property (e.g. `import-service.host`) or even "hard-code" a well-known alias for the host knowing that orchestration can make it available.

And the one thing i was hoping for isn't being provided: a wrapped rest template that has methods mapped to end-points. So, instead of specifying "http://rdw-ingest-import-service/exams/imports/reconciliation" with params and all that, you would just use something like `template.getReconciliation(...)`.

bgsavage:
The repository layer is serving as my wrapped template here.  Does this work for you?

malaffoon:
I understand that we can abstract the functionality behind a repository, and what you've done is exactly that and good. But i was hoping that you wouldn't need to have any (visible) http code.

bgsavage:
Oh similar to how we used to generate code from a wsdl for soap?

I'll go ahead and remove Eureka.  I just thought it was really cool and easy to stand up, but I can just use configured endpoints just as easy if this is preferred.  I guess I had a different understanding what we were trying to do here.

Thanks for the feedback!

PR #115, opened 2017-05-02
Feature/dwr 206 csv report

bgsavage:
Iterative task-service and scheduled job of emailing the reconciliation report.  The cron expression is configurable.  Right now the default implementation is has set host: local which means it will just write the email out to the logs.

TODO: 
Unit / integration tests
Health endpoint
Dockerization


task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/TaskServiceApplication.java
```diff
@@ -0,0 +1,15 @@
+package org.opentestsystem.rdw.ingest.tasking;
```
malaffoon: Hmm, something wrong with the file path. It is `service/src/main/java/org.opentestsystem.rdw.ingest.tasking` and it should be `service/src/main/java/org/opentestsystem/rdw/ingest/tasking`. This happened to me once a while ago because IntelliJ lets  you use dots when creating a folder (vs using dots when creating a package). Might check your file system to see if the folders all actually exist.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/ImportRepository.java
```diff
@@ -0,0 +1,7 @@
+package org.opentestsystem.rdw.ingest.tasking.repository;
+
+import org.opentestsystem.rdw.ingest.tasking.model.ReconciliationReport;
+
+public interface ImportRepository {
+    ReconciliationReport getReconciliation();
```
malaffoon: javadocs would be nice; as a naive user i have no idea what the report would look like with no arguments.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/service/impl/LoggingJavaMailSender.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String MessagePrefix = "**** DEV ONLY Fake Mail Sender **** >> ";
+
+    @Override
+    public MimeMessage createMimeMessage() {
+        log("Creating Mime Message...");
+        return new MimeMessage(Session.getInstance(new Properties()));
+    }
+
+    @Override
+    public MimeMessage createMimeMessage(InputStream contentStream) throws MailException {
```
malaffoon: nit: param (here and other methods) could be `final`

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/service/impl/ReconciliationServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Component
+public class ReconciliationServiceImpl implements ReconciliationService {
+    private static final Logger logger = LoggerFactory.getLogger(ReconciliationServiceImpl.class);
+    final ImportRepository importRepository;
+    final JavaMailSender mailSender;
+
+    @Value("${task.sendReconciliationReport.email.to}")
+    private String sendReportTo;
+
```
malaffoon: Are you going to add `cc` and `bcc`? Of course, that will be a lot of params and you should probably add a properties class instead of using the @Value's.

task-service/src/main/resources/application.yml
```diff
@@ -0,0 +1,22 @@
...
+    auth:
+      token:
+      username: dwtest@example.com
+      password: password
+
+task:
+  sendReconciliationReport:
+    cron: 0 16 15 * * *
+    email:
+      to: bsavage@fairwaytech.com
```
malaffoon: might change to something less personal for testing, e.g. to: client@example.com, from: ops@sbac.com

task-service/src/main/resources/application.yml
```diff
@@ -0,0 +1,22 @@
...
+    host: local
+    port: 465
+    protocol: smtps
+
+repository:
+  import:
+    host: http://localhost:8080
+    auth:
+      token:
+      username: dwtest@example.com
```
malaffoon: Hmm, this shouldn't be checked in. Have to think about how to get around that.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/RepositoryConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.converter.ByteArrayHttpMessageConverter;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.web.client.RestTemplate;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Configuration
+public class RepositoryConfiguration {
```
malaffoon: Rename would propagate to this as well.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/ImportRepository.java
malaffoon: I wonder if this should take query params. I know we have defaults in the exam controller but it seems like we'd want to configure it explicitly in the ReconciliationService or task definition.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/TaskServiceApplication.java
chdietz: Just did a refactor like this yesterday, did it wrong first then did a refactor rename on the package in the project view and it worked fine.  But the old package folder was left out there and just had to delete it.

import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
```diff
@@ -7,19 +7,23 @@
 import org.springframework.cloud.aws.autoconfigure.cache.ElastiCacheAutoConfiguration;
 import org.springframework.cloud.aws.autoconfigure.context.ContextStackAutoConfiguration;
 import org.springframework.cloud.aws.autoconfigure.mail.MailSenderAutoConfiguration;
+import org.springframework.cloud.client.discovery.EnableDiscoveryClient;
```
chdietz: So I understand you removed Eureka and this is providing you with discovery?

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/TaskServiceApplication.java
```diff
@@ -0,0 +1,15 @@
...
+
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+@SpringBootApplication
+@EnableScheduling
+public class TaskServiceApplication {
```
chdietz: What does the task service do?
I was going to mention we may want to discus what level of commenting, javadoc, we want in the open source project we are building. Maybe have a standard?

chdietz: Another comment on this.  Is this intended to be a generic ingest scheduled service?
The spring batch migrate I am working on will need a schedule, do we want to say all scheduled jobs run from task service?

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/ImportRepository.java
bgsavage: I think from task-service's perspective, it's a repo.  Sure you and I know that it's going to connect to a micro-service, but that's just one implementation.  Another implementation could be getting the data from S3, or a database somewhere else.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/service/impl/ReconciliationServiceImpl.java
bgsavage: I moved to a properties class, though I'm not adding cc and bcc at this time.  This can easily be added later if required.

import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
bgsavage: Oops, this can be removed.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/ImportRepository.java
malaffoon: Hmm, have to agree to disagree i guess. To me a repository = CRUD operations on a resource. This is asking for a report to be generated, not manipulating the storage of said report.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/TaskServiceApplication.java
bgsavage: Yeah I think all scheduled jobs should run from the task service.

task-service/src/main/java/org.opentestsystem.rdw.ingest.tasking/repository/ImportRepository.java
esotrope: I think spring defines the repo abstraction as an interface with any data storage/persistence system - so it could be a filesystem or database

bgsavage: I agree with you both conceptually with what a repository is.  This is why I believe at this layer, we are only "reading" reconciliation data, and shouldn't have knowledge under the hood that a report is generated on the fly.  This allows for a more flexible design should we pivot in the future on where we get this data.  I believe that putting this in the service layer smears that abstraction a bit.  I always thought anytime you have to grab data external to our current system, you should do it through a repo layer so that there becomes a clear distinction on where all of your external calls are made.

task-service/src/main/resources/application.yml
bgsavage: Do you think I should write a local impl that returns seeded data?

malaffoon: Perhaps we should talk about this. Under normal operations, the reconciliation report be generated on behalf of a certain user, correct? And we won't have the user credentials. I think this may tie into the auth stuff i am struggling with right now. The alternative would be a special user that doesn't have data restrictions, etc.. but that will be annoying to set up.

bgsavage: We've solved this in the past by standing up two instances of the same service, a private instance behind a firewall which doesn't require authentication which we can easily call internally, and a public instance which requires authentication.  Not saying that's something we should do, but just a thought.

PR #116, opened 2017-05-03
Adjusted Student id to Integer.

agorina:
Due to changes in https://github.com/SmarterApp/RDW_Schema/pull/49

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/StudentRepositoryImpl.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/StudentRepositoryImpl.java
@@ -41,9 +41,9 @@ class StudentRepositoryImpl implements StudentRepository {
 
     @Override
     @Transactional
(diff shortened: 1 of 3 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/StudentRepositoryImpl.java
```diff
@@ -41,9 +41,9 @@
 
     @Override
     @Transactional
-    public long upsert(final Student student, final long importId) {
+    public int upsert(final Student student, final long importId) {
 
-        final Long studentId = (long) new SimpleJdbcCall(jdbcTemplate)
+        final Integer studentId = (int) new SimpleJdbcCall(jdbcTemplate)
```
malaffoon: Seems very odd to make this an `Integer` and not an `int`, since it is autoboxing and then immediately undoing it.

agorina: Will change.

PR #117, opened 2017-05-03
Feature/dwr 206 csv report

bgsavage:
Creating a new pull request to make it easier to see new code changes.

TODO:
 - finish dockerization
 - health endpoint


docker/docker-compose.yml
```diff
@@ -100,4 +100,18 @@ services:
...
-      - "dwhost:$DATAWAREHOUSE_HOST"
\ No newline at end of file
+      - "dwhost:$DATAWAREHOUSE_HOST"
+
+  task-service:
+    image: fwsbac/rdw-ingest-task-service
+    ports:
+      - 8084:8080
+    depends_on:
+      - import-service
```
malaffoon: need `config-server` in here too

task-service/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+apply plugin: 'org.springframework.boot'
+
+dependencies {
+    // handy for creating metadata with @ConfigurationProperties
+    optional "org.springframework.boot:spring-boot-configuration-processor"
+
+    compile 'org.springframework.boot:spring-boot-starter-actuator'
+    compile 'org.springframework.cloud:spring-cloud-starter-config'
+
+    compile 'org.opentestsystem.rdw.common:rdw-common-model'
```
malaffoon: I don't think you need this (?)

task-service/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+
+    // enables email
+    compile 'org.springframework.boot:spring-boot-starter-mail'
+
+    testCompile 'org.springframework.boot:spring-boot-starter-test'
+    testCompile 'org.springframework.cloud:spring-cloud-stream-test-support'
```
malaffoon: I don't think you need this.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/TaskServiceApplication.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * This application is intended to handle all task scheduling
+ * required by RDW_Ingest
+ */
+@SpringBootApplication
+@EnableScheduling
+public class TaskServiceApplication {
+
+    public static void main(String[] args) throws Exception {
```
malaffoon: `final`

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.nio.charset.Charset;
+import java.util.Base64;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@Component
+public class ImportRepositoryImpl implements ImportRepository {
+    private static final Pattern filenameRegexPattern = Pattern.compile("filename=[\"'](.*)[\"']");
+    private static final String reconciliationUri = "/exams/imports/reconciliation";
```
malaffoon: I would think this would be combined with the host in a single property, importReportUrl or some such.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    private String password;
+
+    @Autowired
+    public ImportRepositoryImpl(RestTemplate restTemplate) {
+        this.restTemplate = restTemplate;
+    }
+
+    @Override
+    public ReconciliationReport getReconciliation(final String query) {
+        String queryString = StringUtils.isEmpty(query)
```
malaffoon: There are utilities for constructing a well-formed URL from the path and params. I'd hate to reinvent this stuff and get it wrong.

malaffoon:
As noted in conversation, you need a bootstrap.yml for the app.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -0,0 +1,81 @@
...
+@Component
+public class ImportRepositoryImpl implements ImportRepository {
+    private static final Pattern filenameRegexPattern = Pattern.compile("filename=[\"'](.*)[\"']");
+    private static final String reconciliationUri = "/exams/imports/reconciliation";
+    private final RestTemplate restTemplate;
+
+    @Value("${repository.import.host}")
+    private String importHost;
+
+    @Value("${repository.import.auth.username}")
```
malaffoon: I think username and password are going to end up being per request, not a fixed configuration setting. A followup.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -0,0 +1,81 @@
...
+        return new HttpHeaders() {
+            {
+                String auth = username + ":" + password;
+                byte[] encodedAuth = Base64.getEncoder().encode(auth.getBytes(Charset.forName("US-ASCII")));
+                String authHeader = "Basic " + new String(encodedAuth);
+
+                set("Authorization", authHeader);
+            }
+
+            {
```
malaffoon: i have to say this is very odd syntax; i usually reserve this for statically initialized class variables

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -0,0 +1,81 @@
...
+                ? matcher.group(1)
+                : "reconciliation_report.csv";
+
+        return new ReconciliationReport(filename, response.getBody());
+    }
+
+    private HttpHeaders createHeaders() {
+        return new HttpHeaders() {
+            {
+                String auth = username + ":" + password;
```
malaffoon: Hmm, i'd be surprised if there isn't a helper for this. On the other hand it will probably change to be a Bearer token or something.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/ServiceConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+import org.opentestsystem.rdw.ingest.tasking.service.impl.LoggingJavaMailSender;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.mail.javamail.JavaMailSender;
+
+@Configuration
+public class ServiceConfiguration {
```
malaffoon: I'd suggest some javadocs explaining how this is overriding the mail sender and, if `spring.mail.host` is set, the system will auto-configure an Amazon mail sender.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
```diff
@@ -0,0 +1,67 @@
...
+@ConfigurationProperties(prefix = "task.sendReconciliationReport")
+public class ReconciliationReportProperties {
+    private String query;
+    private Email email;
+
+    public String getQuery() {
+        return query;
+    }
+
+    public void setQuery(String query) {
```
malaffoon: All the setters in here can probably be package-private

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ScheduledTasks.java
```diff
@@ -0,0 +1,28 @@
...
+    public ScheduledTasks(final ReconciliationService service, ReconciliationReportProperties properties) {
+        this.service = service;
+        this.properties = properties;
+    }
+
+    @Scheduled(cron = "${task.sendReconciliationReport.cron}")
+    public void sendReconciliationReport() {
+        logger.info("Scheduled task triggered: Send Reconciliation Report");
+        service.sendReport(properties);
+        logger.info("Scheduled task completed: Send Reconciliation Report");
```
malaffoon: personal preference i'm sure but i think one of these log statements doesn't need to be INFO, perhaps DEBUG or TRACE.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImplTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    public void itShouldAllowForLeadingQuestionInQuery() {
+        ImportRepository repository = getRepository();
+        String query = "?status=PROCESSED";
+
+        ReconciliationReport actual = repository.getReconciliation(query);
+
+        assertThat(actual).isNotNull();
+        assertThat(urlSpy).isEqualTo("null/exams/imports/reconciliation?status=PROCESSED");
```
malaffoon: Might look into ArgumentCaptor instead of this approach.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImplTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertThat(actual).isNotNull();
+        assertThat(actual.getFilename()).isEqualTo(filename);
+    }
+
+    private ImportRepository getRepository() {
+        return getRepository("rdw_import_2017-05-03.csv");
+    }
+
+    private ImportRepository getRepository(String filename) {
+        restTemplate = mock(RestTemplate.class);
```
malaffoon: You could use a `@Before` for this where you create the mock rest template and the repository, with a default behavior. Then you could override that behavior in the one test that needs something different. It's just a bit more conventional than having a helper method like this.
```java
    private RestTemplate restTemplate;
    private ImportRepository repository;

    @Before
    public void createRepository() {
        restTemplate = mock(RestTemplate.class);
        ...
        repository = new ImportRepositoryImpl(restTemplate);
    } 
```

malaffoon: Hmm, but since you did that in the next test i guess you already know that.

task-service/build.gradle
bgsavage: Nope I was thinking at one point to have a common model between import-service and task-service but then I opted not to.  Good catch.

task-service/build.gradle
bgsavage: nope, you're right I don't.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: I thought about that, but then what if we end up having different calls to the import-service?  We would have to maintain the host in multiple configurations.  I see the host and the endpoint uri as separate concerns.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: I'll look into this.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: I know this auth stuff we will have to revisit in a later story.  I can stick todos in there.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: Alright I'll try this.  Initially I was trying to make the headers static but had to go back once I made username/password configurable.  I'm not happy with how the formatter spaces these out anyway.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/ServiceConfiguration.java
bgsavage: will do.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
bgsavage: Would my unit tests still have access?

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ScheduledTasks.java
bgsavage: I can set the second one to debug.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
malaffoon: Well, we can refactor when/if we get there. However, if you'd like to keep them separate, please introduce a property for the path and then use that combined with the host.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImplTest.java
bgsavage: Yeah, because I had a different scenario I wanted to tests.  When I started this approach I didn't realize only one of my tests would have a different.  I can move it to a `@Before` do reduce the code duplication.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
malaffoon: yup, unit tests are in the same package

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: I'm wondering also if path even should be configurable?  If the path changes, then it would also require a code change to handle the new model.  Since the path can only ever be one path, should we make this configurable to allow ourselves to shoot ourselves in the foot?

malaffoon: I think an ops guy would disagree. Imagine for instance that the import-service needs to be run behind a proxy that handles routing based on an introduced root. So, instead of hitting https://import-service/exams/imports/reconciliation, it really needs to hit http://rdw-proxy:8819/import/exams/imports/reconciliation. Sure, we could probably get it to work using what we have but it would be a simpler if it was configurable.

bgsavage: Maybe I'm missing something, but wouldn't the host just then be http://rdw-proxy:8819/import ?  I can rename the parameter to just baseUri since that's how I've actually implemented it, if that helps clarify.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
bgsavage: I think what you're suggesting here is just to remove the "public" access modifier on my setters and it will default to package-private, correct?

bgsavage: This doesn't compile as I have the properties in a different package than the service, since I'm allowing the caller to pass in the properties to my service.  My thinking here is that we should be able call sendReport and pass in some other properties without having to explicitly set them in a config file, if we wanted.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
bgsavage: So I tried using this one as described: http://www.baeldung.com/spring-uricomponentsbuilder , but it breaks my itShouldAllowForLeadingQuestionInQuery unit test.  I was going to allow for the query configuration to be set like this : 
`?status=PROCESSED&before=P2D` so that the whole configuration looks like a query string.  If I use the UriComponents builder, then having a leading `?` won't work.  Which is fine, I was always going to support without the leading `?`.  Just wanted to let you know I'm losing functionality / flexibility, but perhaps the functionality I was trying to support isn't standard so that's fine.

PR #118, opened 2017-05-03
Feature/ci improvements

rcordeau:
This is moving forward on the CI improvements.  RDW_Schema is a child project, pointing to the most current tip of Schema develop branch. Setting the version via the command line is here, and the CI system will take advantage of that and introduce a build number (or name in the case of RELEASE on the master branch) to the version when making .jar files. This needs to be done to other projects as well, but will start with Ingest.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/118)


build.gradle
```diff
@@ -18,12 +18,14 @@ apply plugin: 'io.spring.dependency-management'
 
 allprojects {
     group = 'org.opentestsystem.rdw.ingest'
-    version = '0.0.1-SNAPSHOT'
+    version = '0.0.1'
```
malaffoon: I'm asking this before i look at the whole PR: if we aren't creating release artifacts and pushing them to a repo then we can't make this change because a local build (the only way to create these artifacts) produces the -SNAPSHOT version.

import-service/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,3 @@
+spring:
```
malaffoon: This should be under src/test/resources

migrate-service/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,3 @@
+spring:
```
malaffoon: ditto (src/test/resources) and i'll stop commenting on each one

rcordeau:
Ok, changes introduced per review comments.

build.gradle
```diff
@@ -87,4 +88,38 @@ subprojects {
...
 }
+
+task cleanAllTest(type: GradleBuild) {
+    dir = "RDW_Schema"
+    tasks = ["cleanAll-test"]
+}
+
+if (project.hasProperty("buildNumber")) {
+    project.setVersion(project.version.split("-")[0] + "." + buildNumber)
+    println "I will be using build version: " + version
```
malaffoon: did you want to remove this before checking in?

rcordeau:
whoops

rcordeau:
Ok - version information gets in to the resultant build.version file as well as the top-level jar file.

PR #119, opened 2017-05-04
Use ArchiveService from RDW_Common.

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
```diff
@@ -16,10 +17,14 @@
 		ContextStackAutoConfiguration.class
 })
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
-@Import({YamlPropertiesConfigurator.class, StatusConfiguration.class})
+@Import({
+		YamlPropertiesConfigurator.class,
```
malaffoon: hmm, i'll check the whitespace on this

import-service/src/test/java/org/opentestsystem/rdw/ingest/status/DatabaseStatusIndicatorTest.java
```diff
@@ -36,7 +36,7 @@ public void createStatusIndicator() {
         });
 
         archiveService = mock(ArchiveService.class);
-        when(archiveService.readResource("status")).thenReturn("status check");
+        when(archiveService.readResource("status")).thenReturn("status check".getBytes());
```
malaffoon: in a test like this the encoding doesn't matter

build.gradle
```diff
@@ -42,6 +42,7 @@ subprojects {
 
     dependencyManagement {
         dependencies {
+            dependency 'org.opentestsystem.rdw.common:rdw-common-archive:0.0.1-SNAPSHOT'
```
rcordeau: When we move to appending build numbers to the versions, do you want to remove SNAPSHOT?

malaffoon: Absolutely.

PR #123, opened 2017-05-04
Feature/readme changes

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/123)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 RDW ingest applications:
 1. Import Service - RESTful API for submitting test results (exams), packages, etc.
 1. Exam Processor - Spring Cloud Stream application for processing test results.
+1. Package Processor - Spring Cloud Stream application for processing Assessment Packages.
 
 RDW Ingest uses other processes:
 1. MySQL - warehouse and reporting databases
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +79,13 @@ java                                    8-jre-alpine        d85b17c6762e
 ```
 
 ### Running
+The applications depend on the database being configured properly. This project has the RDW_Schema submodule pinned at a
+version that is compatible with the current version of this project, so there is no need to separately clone the RDW_Schema.
+```bash
+cd RDW_Schema
+./gradlew cleanAll migrateAll
```
bgsavage: Does ./gradlew still work at this level?  Also might be nice to have a disclaimer on what cleanAll and migrateAll does in case someone isn't ready to wipe their data.  At some point we should only be using a migrateAll script.

rcordeau: Yes, ./gradlew still works that this level because the RDW_Schema project is really a completely separate project, and the recursive clone pulls the entire thing down as if you cloned it separately. To get a description of what the tasks do, run "./gradlew tasks".  All of the tasks are grouped and have descriptions.

rcordeau: And yes, at some point we will be using migrateAll mostly instead of clean and migrate.

PR #124, opened 2017-05-05
Implement authorization using OAuth2 token to fetch user tenancy chain.

malaffoon:
A few things got swept up into this:
1a. Use OAuth2 bearer token to get tenancy chain for authorization
1b. Remove any LDAP auth, including handy local implementation.
2. Eliminate any authorization further down the pipeline; i.e. don't need to know the tenancy chain when processing payloads.
3. Add caching of auth credentials; fetching from OpenAM take about 40-50ms so this significantly improves performance.

Note a couple issues that i will pursue ASAP:
1. There is no local-only implementation. Even for local dev it is hitting OpenAM. 
2. The data fetch for the reconciliation report is broken since it is using Basic auth.
3. Integration tests for the controllers are circumventing security instead of testing it.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -35,28 +33,28 @@ public ExamProcessorConfiguration(final TDSReportProcessor reportProcessor,
         this.importRepository = importRepository;
     }
 
-    @ServiceActivator(inputChannel = Processor.INPUT)
+    @ServiceActivator(inputChannel = Sink.INPUT)
```
malaffoon: Unrelated.

import-service/src/main/resources/application.yml
```diff
@@ -17,6 +17,11 @@ cloud:
     credentials:
       instance-profile: false
 
+# TODO - can we jam in a local-only solution?
```
malaffoon: yeah ... this

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/TestAppConfig.java
```diff
@@ -1,17 +1,72 @@
...
-@Import(EmbeddedLdapAutoConfiguration.class)
+//@Import(SecurityConfigurer.class)
 @EnableWebMvc
 class TestAppConfig {
+
+//    @Autowired
+//    private ForgeRockTokenServices tokenServices;
+//
+//    @PostConstruct
+//    public void mockRestOperations() {
```
malaffoon: I've intentionally left this in to preserve the work. I will clean it up in subsequent checkins.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -35,28 +33,28 @@ public ExamProcessorConfiguration(final TDSReportProcessor reportProcessor,
...
 
-        long importId = accessor.getImportId();
+        final long importId = accessor.getImportId();
 
         try {
             final TDSReport report = toTDSReport(new ByteArrayInputStream(payload.getBytes("UTF-8")), importId);
             if (report == null) return;
 
-            process(toTDSReport(new ByteArrayInputStream(payload.getBytes("UTF-8")), importId), TenancyChain.fromString(accessor.getUserTenancyChain()), importId);
+            process(toTDSReport(new ByteArrayInputStream(payload.getBytes("UTF-8")), importId), importId);
```
jtreuting: Should this be passing in `report` as the first param?

Also, does `TDSReport` have the importId already, if so does it need to be passed to the `process` method?

bgsavage:
Let me know if you want me to jump in on  

> The data fetch for the reconciliation report is broken since it is using Basic auth.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacUser.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * A user with SBAC details.
+ */
+public class SbacUser extends User implements SbacUserDetails {
```
jtreuting: Is any of this duplicate work from the Permission tenancy chain work that @esotrope did?

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
```diff
@@ -55,21 +43,9 @@ public AuthenticationManager authenticationManager(final AuthenticationManagerBu
...
-
-        // This code has been ripped from LdapAutoConfiguration.
-        final LdapContextSource source = new LdapContextSource();
-        source.setUserDn(properties.getUsername());
-        source.setPassword(properties.getPassword());
-        source.setBase(properties.getBase());
-        source.setUrls(properties.determineUrls(environment));
-        source.setBaseEnvironmentProperties(Collections.unmodifiableMap(properties.getBaseEnvironment()));
-        return source;
+//    @ConfigurationProperties(prefix = "security.oauth2")
```
jtreuting: Was this maybe left over from testing and isn't needed?

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServices.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public OAuth2AccessToken readAccessToken(final String accessToken) {
+        return tokenConverter.extractAccessToken(accessToken, getTokenInfo(accessToken));
+    }
+
+    private Map<String, Object> getTokenInfo(final String token) {
+        final Map<String,String> params = newHashMap();
+        params.put("access_token", token);
```
bgsavage: Should this be a constant, similar to how you did expires_in?

malaffoon: Yeah, probably ... will change.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
malaffoon: No. I have a note to myself to get this working instead of using a @Value annotation on the class itself.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacUser.java
malaffoon: Not yet, no. If we chose to use the role of ASMTDATALOAD to map to a permission then i will use Landon's stuff. Currently it is based solely on the role existing in the tenancy chain.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -0,0 +1,80 @@
...
+        if (tenancyChain.hasRole(DataLoadAuthority)) {
+            authorities.add(new SimpleGrantedAuthority(DataLoadAuthority));
+        } else {
+            logger.info("User {} doesn't have data load authority: {}", username, tenancyChain.toString());
+        }
+        final SbacUser user = SbacUser.builder()
+                .username(username)
+                .authorities(newArrayList(authorities))
+                .uid((String) map.get("uid"))
+                .sbacUuid((String) map.get("sbacUUID"))
```
bgsavage: I'm guessing it's probably not likely to happen in production or else other systems would break as well, but just curious what would happen if one of the fields aren't present in the token that we are expecting to be there.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
malaffoon: Good catch on the first one, i'll tweak that.

TDSReport doesn't know anything about imports so we have to pass both.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
malaffoon: They are all allowed to be null here. When they are converted to authentication objects, an exception will be thrown if the username is missing (i think that is the only one that is required).

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -0,0 +1,80 @@
...
+
+        return new OAuth2Authentication(request, authentication);
+    }
+
+    private static Set<String> extractScope(final Map<String, ?> map) {
+        return map.containsKey(SCOPE) ? newHashSet((List<String>) map.get(SCOPE)) : null;
+    }
+
+    private static Long extractExpirationAsTime(final Map<String, ?> map) {
+        return System.currentTimeMillis() + 1000 * (map.containsKey(EXPIRES_IN) ? (Integer) map.get(EXPIRES_IN) : 3600);
```
bgsavage: Java doesn't have constants for these numbers?

Also, is the else of 3600 a common scenario you think we will hit?  If yes, we may want to consider making it configurable.  If it's just a safeguard then no need.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
jtreuting: i wasn't able to find `toTDSReport()` in GitHub for some reason so not sure what it's doing wi the importId.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
malaffoon: I will probably make the 3600 configurable although the handling of the EXPIRES_IN is the real gatekeeper on the credentials. 

As for the 1000, i'm mixed. Sure, it's a number but is 
```
    return System.currentTimeMillis() + 1000 * secs;
```
really that much more readable than
```
    return System.currentTimeMillis() + MILLIS_PER_SECOND * secs;
```

import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServicesTest.java
```diff
@@ -0,0 +1,102 @@
...
+        final Map<String, Object> map = passwordGrantTokenMap(accessToken);
+        map.put(EXPIRES_IN, 1);
+        when(restOperations.getForObject(anyString(), any(), any(Map.class))).thenReturn(map);
+
+        assertPasswordGrantAuth(services.loadAuthentication(accessToken));
+        assertPasswordGrantAuth(services.loadAuthentication(accessToken));
+
+        verify(restOperations, times(1)).getForObject(anyString(), any(), any(Map.class));
+
+        sleep(1200);
```
bgsavage: uh oh, why are we sleeping here?  Is there something we can listen for instead?

import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServicesTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        verify(restOperations, times(1)).getForObject(anyString(), any(), any(Map.class));
+    }
+
+    @Test
+    public void itShouldInvalidateBasedOnExpiresInValue() throws InterruptedException {
+        final String accessToken = "abc123";
+
+        final Map<String, Object> map = passwordGrantTokenMap(accessToken);
+        map.put(EXPIRES_IN, 1);
```
bgsavage: I wonder if you can put -1 here instead so you don't have to sleep on line 73.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
bgsavage: Maybe breaking up this line would help with the readability.
```
 final int tokenExpiresInMillis = map.containsKey(EXPIRES_IN) ? (Integer) map.get(EXPIRES_IN) : 3600
 return System.currentTimeMillis() + 1000 * expiresInSeconds;

```

Idk, just thinking out loud.

malaffoon: if it helps readability i can do that

import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServicesTest.java
malaffoon: Unfortunately it would then instantly invalidate which is not the code path i'm trying to test. Yeah, i dislike sleep() a lot but i'm not sure how else to test this.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
agorina: It uses it in case of the error. The code is in this file:
```
    private TDSReport toTDSReport(final ByteArrayInputStream xml, final long importId) {
        try {
            return tdsReportFromXml(xml);
        } catch (final RuntimeException e) {
            logger.info("failed to convert xml to tds report with error: " + e.getMessage());
            importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
        }
        return null;
    }
```

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServices.java
```diff
@@ -0,0 +1,153 @@
...
+     */
+    private class ExpiringAuthenticationCache
+            extends ForwardingCache<String, OAuth2Authentication>
+            implements LoadingCache<String, OAuth2Authentication> {
+
+        private final LoadingCache<String, OAuth2Authentication> delegate;
+
+        ExpiringAuthenticationCache() {
+            delegate = CacheBuilder.newBuilder()
+                    .maximumSize(100)
```
chdietz: 100 seams reasonable but what it they are send are large volume of requests and do something we did not anticipate?

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServices.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static org.springframework.security.oauth2.common.OAuth2AccessToken.EXPIRES_IN;
+
+/**
+ * An implementation of {@link ResourceServerTokenServices} backed by ForgeRock's OpenAM, which doesn't provide
+ * the POST /oauth2/check_token end-point. Instead it has GET /oauth2/tokeninfo. And it doesn't require client info.
+ *
```
chdietz: Could add a comment that in addition to having to add the get a cache is also added.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -0,0 +1,80 @@
+package org.opentestsystem.rdw.ingest.auth;
+
+import org.opentestsystem.rdw.utils.TenancyChain;
```
chdietz: should common modules have 'common' in the package?
- i know, a bit off topic, but most of our code follows that pattern
- i am thinking of people browsing the code not in an IDE

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
chdietz: On note 2 in your into: "The data fetch for the reconciliation report is broken since it is using Basic auth."

When data is fetched for the reconciliation report we do not have the credentials of the user so we are doing this as 'RDW'?  
If so, do we have to do Oauth2? Or could we have a private endpoint that is only exposed to our network and protected with something like basic auth?

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServices.java
```diff
@@ -0,0 +1,153 @@
...
+            logger.debug("getTokenInfo returned error: {}", result.get("error"));
+            throw new InvalidTokenException(token + ": " + result.get("error"));
+        }
+
+        return (Map<String, Object>) result;
+    }
+
+    /**
+     * Wrap a LoadingCache to invalidate any entries where the expires_in value has passed.
+     * This requires that the loader set the "expires_in" request parameter in the auth request.
```
chdietz: Do we really need to do this?  I see the cache is 15 minutes.  I think the recommended token expire is seconds to minutes, the default in OpenAM 13.5 is 60 seconds, not sure about version 12.

If I understand this correct, when the cache finds a hit, this implementation checks the original expires_in to see if the token is expired.  

My understanding is this is added to force clients to exchange the refresh with credentials for a new access token. Primarily done if the token falls into the wrong hands, so why not just leave this out and set the cache to a smaller time like 30 or 60 seconds.  Even at 15 seconds that is a huge jump over 50ms.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
malaffoon: I'm fiddling with that right now. My first approach was going to be to use a client-credential grant to authenticate. I really don't want to have credentials going through our system so, if that approach doesn't work, we will need to expose certain internal functionality on a different port so it can be kept private without auth.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ForgeRockTokenServices.java
malaffoon: I see the settings in OpenAM claiming the configured value is 60 seconds. However, empirically when i get a password token through OpenAM the expires_in value is 10 hours in the future. So i figured 15 minutes would be oodles of underestimate. 

The way this works is just an optimization though: the expires_in always trumps the existence of the token in the cache. If the expires_in is really 10 seconds, this will honor that (within ms accuracy).

import-service/src/main/resources/application.yml
chdietz: ok

PR #127, opened 2017-05-08
tweak

malaffoon:
@bgsavage can i make this tweak? You aren't really using that env variable are you?


bgsavage:
This is a relic of something I was testing.  I ended up overriding this value via diff config profiles anyway.  You're good to remove.

PR #128, opened 2017-05-08
Added an option to pass param between steps.

agorina:
Next I will work on adding unit testing.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
@@ -3,6 +3,7 @@ package org.opentestsystem.rdw.ingest.migrate.reporting;
 
 import org.opentestsystem.rdw.ingest.common.model.ImportContent;
 
+import java.util.Collection;
 import java.util.List;
 
 import static com.google.common.collect.Lists.newArrayList;
@@ -22,7 +23,7 @@ public class MigrateBatch {
      * @param endImportId   the last import id in the batch
      * @param content       the collection of the {@link ImportContent} within the batch
      */
-    public MigrateBatch(final long startImportId, final long endImportId, final List<ImportContent> content) {
+    public MigrateBatch(final long startImportId, final long endImportId, final Collection<ImportContent> content) {
         this.startImportId = startImportId;
         this.endImportId = endImportId;
         this.importContent = newArrayList(content);
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
@@ -1,0 +1,42 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
+
+
+import org.opentestsystem.rdw.ingest.common.model.ImportContent;
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * A batch of imports to be migrated
+ */
+public class MigrateBatch {
+    final long startImportId;
+    final long endImportId;
+    final List<ImportContent> importContent;
+
+    /**
+     * Constructor
+     *
+     * @param startImportId the first import id in the batch
+     * @param endImportId   the last import id in the batch
+     * @param content       the collection of the {@link ImportContent} within the batch
+     */
+    public MigrateBatch(final long startImportId, final long endImportId, final List<ImportContent> content) {
+        this.startImportId = startImportId;
+        this.endImportId = endImportId;
+        this.importContent = newArrayList(content);
+    }
+
+    public long getStartImportId() {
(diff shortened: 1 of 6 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -103,8 +89,8 @@ public Job migrateReportingJob() {
                     .repository(jobRepository())
                     .incrementer(new RunIdIncrementer())
                     .listener(listener())
-                    .flow(stepReporting())
-                    .next(stepWarehouse())
+                    .flow(stepWarehouse())
```
agorina: I wanted to see that I can pass parameter from 'warehouse' to 'reporting'.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
```diff
@@ -0,0 +1,47 @@
...
+    @Autowired
+    @Qualifier("warehouseJdbcTemplate")
+    private NamedParameterJdbcTemplate warehouseJdbcTemplate;
+
+    @Override
+    public RepeatStatus execute(final StepContribution contribution, final ChunkContext chunkContext) throws Exception {
+
+        final List<ImportContent> contentList = newArrayList();
+        contentList.add(ImportContent.EXAM);
+        setJobBatch(chunkContext, new MigrateBatch(0, 1000, contentList));
```
agorina: This gets stored as JSON in the Spring Batch tables:
```
{"map":[{"entry":{"string":"batch","org.opentestsystem.rdw.ingest.migrate.reporting.MigrateBatch":{"startImportId":0,"endImportId":1000,"importContent":[{"org.opentestsystem.rdw.ingest.common.model.ImportContent":"EXAM"}]}}}]}
```
So that maybe a reason to keep our own table to be able to store the last processed id as a separate column to query it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
```diff
@@ -0,0 +1,42 @@
...
+    final long startImportId;
+    final long endImportId;
+    final List<ImportContent> importContent;
+
+    /**
+     * Constructor
+     *
+     * @param startImportId the first import id in the batch
+     * @param endImportId   the last import id in the batch
+     * @param content       the collection of the {@link ImportContent} within the batch
```
malaffoon: Comment combined with parameter is misleading: either make the parameter a Collection (or an iterable) or explain why it must be a List.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
```diff
@@ -0,0 +1,42 @@
...
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * A batch of imports to be migrated
+ */
+public class MigrateBatch {
+    final long startImportId;
+    final long endImportId;
+    final List<ImportContent> importContent;
```
malaffoon: I know you're just experimenting here but i'm curious why this is a list, or even why it exists. Either we're going to migrate everything or we're not, aren't we?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
malaffoon: I don't disagree for storing the result of the batch migrate process, we have to get the last import id from somewhere that we own.

chdietz: I agree

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
chdietz: This may be the list of root types that exist in the content, 
as in the import_content table

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
agorina: Ok, will change.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatch.java
agorina: Some data get updated infrequently. Also, codes do not have import_id to look up if they have been changed. The idea is that when we get the import_ids, we will also retrieve a list of distinct content types. During the migrate, if there is not CODES type, then codes will not be migrated.

PR #130, opened 2017-05-09
Remove reporting datasource 

chdietz:
and have reporting adbc template use batch datasource to use the same transaction manager.

Next step review batch configuration to see if it can be simplified.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
@@ -21,6 +21,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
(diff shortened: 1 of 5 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -67,6 +68,18 @@ public DataSource batchDataSource() {
...
+    /**
+     * The jdbc template for reporting uses the same datasource as spring batch.
+     * This allows spring batch to control transactions to business data and spring batch tables.
+     * Prefix all tables with reporting or staging in sql when working on business data.
+     *
+     * @return NamedParameterJdbcTemplate
+     */
+    @Bean(name = "reportingJdbcTemplate")
+    public NamedParameterJdbcTemplate reportingJdbcTemplate() {
+        return new NamedParameterJdbcTemplate(batchDataSource());
```
malaffoon: At some point we should stop calling it the batchDataSource and just just let it be *the* dataSource.

chdietz: Yes, good point

PR #131, opened 2017-05-10
generate reconciliation report using jdbc (instead of service call)

malaffoon:
I decided to generate the reconciliation report using jdbc instead of calling out to the import-service. Why?

1. By supporting this in the import-service, that service was complicated by having to know about non-import stuff. Knowledge of exam, student, etc. It felt cleaner in import-service to remove it from there.

2. That means to do it "right", a call would be made to import-service to get the import records of interest, then call some other service (or perform a jdbc query) to get the rest of the data needed for the report. I really, really don't like client-side joins. I don't care that it violates one of the tenants of micro-service architecture, it's just non-performant. So i want a solution where one process is getting all the data ...

3. The next best "right" approach is to have another service that knows about joining data between import and exam/student data. Another moving part to avoid what -- combining the scheduling and the querying. I admit that sucks but the only reason that task exists is to produce this report that we don't expect to need as time goes on. So why not, for now, combine the scheduling and the querying. If things get more complex in the future, we can refactor things.

4. Downside is that any change to import persistence will require a change to both import-service (obviously) and task-service. Any change to exam/student persistence will also require a change to task-service. So this independent micro-service isn't all that independent and will get redeployed together with other services. But i think it is unlikely that there would be a change that affects persistence while not affecting the service API so really this is a specious argument.


import-service/API.md
```diff
@@ -162,38 +162,5 @@ This end-point may be used to get the current status of an import request.
 curl --header "Authorization:Bearer {access_token}" https://import-service/exams/imports/19529
 ```
 
-#### Get Import Reconciliation Report
```
malaffoon: good thing i didn't get as far as sending this to ETS

import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/JdbcRdwImportRepository.java
```diff
@@ -28,7 +26,7 @@
  * Default implementation of {@link RdwImportRepository}.
  */
 @Repository
-class RdwImportRepositoryImpl implements RdwImportRepository {
+class JdbcRdwImportRepository implements RdwImportRepository {
```
malaffoon: unrelated rename

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportContent.java
```diff
@@ -1,54 +0,0 @@
-package org.opentestsystem.rdw.ingest.common.model;
```
malaffoon: No need to communicate this info between processes anymore.

import-service/src/main/resources/application.sql.yml
```diff
@@ -31,40 +30,6 @@ sql:
        AND (:after IS NULL OR i.created > :after)
      LIMIT 100000
 
-    content:
```
malaffoon: moved to task-service

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepository.java
```diff
@@ -0,0 +1,115 @@
...
+        } catch (final IOException e) {
+            logger.warn("Error creating reconciliation report {}", e.getMessage());
+        }
+
+        return new ReconciliationReport(filename, baos.toByteArray());
+    }
+
+    @Override
+    public boolean isAvailable() {
+        // TODO - verify some random query into the database
```
malaffoon: oops, i'll do this and patch PR

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultReconciliationService.java
```diff
@@ -15,27 +16,33 @@
...
 import javax.mail.internet.MimeMessage;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 @Component
-public class ReconciliationServiceImpl implements ReconciliationService {
-    private static final Logger logger = LoggerFactory.getLogger(ReconciliationServiceImpl.class);
-    final ImportRepository importRepository;
-    final JavaMailSender mailSender;
+public class DefaultReconciliationService implements ReconciliationService {
```
malaffoon: unrelated rename as long as i was in here

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultReconciliationService.java
```diff
@@ -15,27 +16,33 @@
...
-        this.importRepository = importRepository;
+    public DefaultReconciliationService(final ReportRepository reportRepository, final JavaMailSender mailSender) {
+        this.reportRepository = reportRepository;
         this.mailSender = mailSender;
     }
 
     @Override
     public void sendReport(final ReconciliationReportProperties properties) {
-        ReconciliationReport report = importRepository.getReconciliation(properties.getQuery());
+        // because it is a configuration option (and not user input), assume query is set properly
```
malaffoon: @bgsavage, think this is okay?

task-service/src/main/resources/application.yml
```diff
@@ -16,15 +16,10 @@ spring:
...
-repository:
-  import:
-    scheme: http
-    host: localhost:8080
-
 task:
   send-reconciliation-report:
-    cron: 0 37 14 * * *
-    query:
+    cron: 0 43 19 * * *
```
malaffoon: i'll set this to `cron 0 0 14 * * *` before checking in

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepositoryIT.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@Transactional
+@ActiveProfiles("test")
+@Sql(statements = {
+        "DELETE FROM exam WHERE id IN (-399, -398, -397)",
```
malaffoon: ugh

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepository.java
```diff
@@ -0,0 +1,115 @@
...
+                        rs.getString("natural_id"),
+                        formatter.format(rs.getTimestamp("completed_at").toInstant())
+                ));
+            } catch (final IOException e) {
+                logger.warn("Error writing results {} to reconciliation report {}", rs, e.getMessage());
+            }
+        }
+    }
+
+    private MapSqlParameterSource mapQueryToSqlParameterSource(final RdwImportQuery query) {
```
malaffoon: This function is the only code that is duplicated between import-service and task-service.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
```diff
@@ -94,6 +95,30 @@ public Builder copy(final RdwImportQuery query) {
             return this;
         }
 
+        /**
+         * Extract values from a params-like (ampersand-delimited) string.
+         *
+         * @param params params-like string, e.g. content=EXAM&status=PROCESSED&after=-PT10H
+         * @return this builder
+         */
+        public Builder params(final String params) {
```
malaffoon: Useful helper that should've been here (to match toParamString() functionality).

import-service/build.gradle
```diff
@@ -23,9 +23,6 @@ dependencies {
     compile 'org.springframework.security.oauth:spring-security-oauth2'
     testCompile 'org.springframework.security:spring-security-test'
 
-    // csv reader
```
malaffoon: A big part of this change is to take all the CSV handling out of import-service and move it to task-service.

task-service/src/main/resources/application.yml
bgsavage: fyi,  You can set @Scheduled to fixedRate while you're testing if you don't want to continuously manipulate the cron expression and wait for it to hit.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
```diff
@@ -94,6 +95,30 @@ public Builder copy(final RdwImportQuery query) {
...
+                final int equals = param.indexOf('=');
+                if (equals == -1) continue;
+                final String value = (equals+1) < param.length() ? emptyToNull(param.substring(equals+1).trim()) : null;
+                switch(param.substring(0, equals).trim().toLowerCase()) {
+                    case "content": content(value); break;
+                    case "status":  status(value); break;
+                    case "batch":   batch(value); break;
+                    case "creator": creator(value); break;
+                    case "before":  before(value); break;
+                    case "after":   after(value); break;
```
bgsavage: Typically I would prefer to create dictionaries instead of switch / case statements just because it makes it easier to soft code and/ore make it dynamic later.  That might just be a personal preference though and may not be applicable here.

common/src/test/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQueryTest.java
```diff
@@ -77,4 +76,32 @@ public void itShouldParseInstant() {
                 .build();
         assertThat(query.getBefore()).isLessThan(Instant.now());
     }
+
+    @Test
+    public void itShouldParseParams() {
+        final RdwImportQuery query = RdwImportQuery.builder().params("status=ACCEPTED&content=EXAM").build();
+        assertThat(query.getContent()).isEqualTo(ImportContent.EXAM);
+        assertThat(query.getStatus()).isEqualTo(ImportStatus.ACCEPTED);
+        assertThat(query.asParamString()).isEqualTo("content=EXAM&status=ACCEPTED");
```
bgsavage: It's interesting the order isn't preserved.  Might be worth adding that note to the javadocs about what order this will be in.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
malaffoon: Yeah, not thrilled with this but it works for now.

common/src/test/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQueryTest.java
malaffoon: added comment

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepository.java
```diff
@@ -0,0 +1,115 @@
...
+    public ReconciliationReport getReconciliationReport(final RdwImportQuery query) {
+        final String filename = "reconciliation_" + new SimpleDateFormat("yyyy-MM-dd").format(new Date()) + ".csv";
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        try (final CsvResultSetWriterAdapter writer = new CsvResultSetWriterAdapter(new OutputStreamWriter(baos), CsvPreference.EXCEL_PREFERENCE)) {
+            writer.writeHeader(DefaultHeaders);
+            final MapSqlParameterSource parameterSource = mapQueryToSqlParameterSource(query);
+            try {
+                jdbcTemplate.query(sqlImportContentExams, parameterSource, writer);
+                jdbcTemplate.query(sqlImportContentIabExams, parameterSource, writer);
```
bgsavage: I'm not sure how I feel about coupling the data retrieval with the CSV writing.  To me, they are separate responsibilities.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
```diff
@@ -1,98 +0,0 @@
...
-import org.springframework.web.client.RestTemplate;
-import org.springframework.web.util.UriComponentsBuilder;
-
-import java.nio.charset.Charset;
-import java.util.Base64;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-@Component
-public class ImportRepositoryImpl implements ImportRepository {
```
bgsavage: Guess you don't have to worry about this anymore.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultReconciliationService.java
bgsavage: We should allow for an empty query though if we just want the default.  Unless you're suggesting we can't make a reasonable default because we at the very least need to know the creator (user).

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepositoryIT.java
bgsavage: Do we need this because these IT aren't truly transactional?

chdietz:
Should the client be able to get the same thing on the reconciliation report by calling a get on an import id?

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
bgsavage: Another thought, a dictionary here would reduce the cyclomatic complexity.  Is that a metric we care about?

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/impl/ImportRepositoryImpl.java
malaffoon: And you were correct all along  ;)

task-service/src/main/resources/application.yml
malaffoon: handy, thanks

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultReconciliationService.java
malaffoon: Well, i was thinking that since this is a configuration option set by ops, we don't have to be as forgiving as if it was being provided by a client of the REST API. And it will complain if there is no query configured.

malaffoon:
@chdietz "Should"? Perhaps. Can they right now, no. In a HATEOAS payload i suppose the import object could have links to exams, students, etc. that are related. I'd want to discuss the separation of concerns though because that would bring back awareness of exam/student in the import service.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
malaffoon: We aren't currently measuring that (although it may be in the coverage report). Although i don't disagree, i'm having trouble coming up with code that is simpler. The dictionary you're picturing would be a map of param key (e.g. "content") to setter method? I know how i would do that in C, C++, or Python but not Java.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepository.java
```diff
@@ -0,0 +1,124 @@
...
+    public ReconciliationReport getReconciliationReport(final RdwImportQuery query) {
+        final String filename = "reconciliation_" + new SimpleDateFormat("yyyy-MM-dd").format(new Date()) + ".csv";
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        try (final CsvResultSetWriterAdapter writer = new CsvResultSetWriterAdapter(new OutputStreamWriter(baos), CsvPreference.EXCEL_PREFERENCE)) {
+            writer.writeHeader(DefaultHeaders);
+            final MapSqlParameterSource parameterSource = mapQueryToSqlParameterSource(query);
+            try {
+                jdbcTemplate.query(sqlImportExams, parameterSource, writer);
+                jdbcTemplate.query(sqlImportIabExams, parameterSource, writer);
```
bgsavage: Now we have some coupling between the data retrieval and the csv output.  To me, these should be separate concerns.

malaffoon: I can't disagree that there is coupling. But this is the JDBC implementation of generating the CSV report. And this is the only implementation of the CSV report. So, until there is need for reusability, is it really a bad thing? It has cost associated with it (intermediate objects, maintaining that extra code, etc.). And i'm willing to refactor as necessary in the future.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/TaskServiceApplication.java
```diff
@@ -19,6 +19,6 @@
 public class TaskServiceApplication {
 
     public static void main(final String[] args) throws Exception {
-        SpringApplication.run(TaskServiceApplication.class);
+        SpringApplication.run(TaskServiceApplication.class, args);
```
malaffoon: Without this, none of the --property=value things work.

bgsavage:
Given your points above, honestly, I'm not really sold on this being a better solution than what we had before.  It's like great, we removed the coupling that import-service we had on exams, but now task-service has to have inner knowledge of both imports and exams.  I guess I'm kind of bothered that this approach was motivated by circumventing the inter-service authentication problem.  I think we will probably bump up against that problem again later, and if we had it already solved, this implementation probably wouldn't even be considered.

I just feel like we are losing functionality and maintainability and not really sure what we are gaining overall.  

That being said, I'm probably just biased since I had already invested so much time in the previous implementation.  Ultimately, I'm going to defer to what you think is best because I know you've already carefully considered it all.

PR #132, opened 2017-05-10
Feature/refactor datasource

chdietz:
Moved to batch data source to the data source config class.

Makes the MigrateReportingConfiguration just a spring batch Job configuration.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingApplication.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingApplication.java
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.ingest.migrate.reporting;
 
 import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
+import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Import;
@@ -10,6 +11,7 @@ import org.springframework.context.annotation.Import;
  * Note on startup warnings: https://github.com/codecentric/spring-boot-starter-batch-web/issues/13
  */
 @SpringBootApplication
+@EnableBatchProcessing
 @Import({MigrateReportingConfiguration.class, YamlPropertiesConfigurator.class})
 public class MigrateReportingApplication {
 
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
@@ -6,7 +6,6 @@ import org.opentestsystem.rdw.ingest.migrate.reporting.step.Warehouse;
 import org.springframework.batch.core.Job;
 import org.springframework.batch.core.JobExecutionListener;
 import org.springframework.batch.core.Step;
-import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.batch.core.configuration.annotation.JobBuilderFactory;
 import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
 import org.springframework.batch.core.launch.support.RunIdIncrementer;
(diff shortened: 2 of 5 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -6,102 +6,49 @@
...
-
-    @Autowired
-    private StepBuilderFactory stepBuilderFactory;
-
 
-    @Autowired
     private Reporting reportingStep;
-
-    @Autowired
     private Warehouse warehouseStep;
```
malaffoon: these can be `final`

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -6,102 +6,49 @@
...
-    @Bean(name = "reportingJdbcTemplate")
-    public NamedParameterJdbcTemplate reportingJdbcTemplate() {
-        return new NamedParameterJdbcTemplate(batchDataSource());
-    }
-
-    @Bean
-    BatchConfigurer configurer(final DataSource dataSource) {
-        return new DefaultBatchConfigurer(dataSource);
+    @Autowired
+    public MigrateReportingConfiguration(JobBuilderFactory jobBuilderFactory,
```
malaffoon: params can be `final`

migrate-reporting/src/main/resources/application.yml
```diff
@@ -16,4 +16,10 @@ spring:
     testWhileIdle: true
     validationQuery: SELECT 1
     driverClassName: com.mysql.jdbc.Driver
-    initialize: false
\ No newline at end of file
+    initialize: false
+
+logging:
```
malaffoon: I wonder if we should introduce a log xml config file. I just recall that the spring logging properties don't let you do a whole lot and having logging config in two places is confusing. Guess we can burn that bridge after we build it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingApplication.java
```diff
@@ -10,6 +11,7 @@
  * Note on startup warnings: https://github.com/codecentric/spring-boot-starter-batch-web/issues/13
  */
 @SpringBootApplication
+@EnableBatchProcessing
```
malaffoon: I admit i'm confused by this change. Wasn't it good to have all the Spring Batch stuff in one place? Suppose we completely changed the implementation for this, removing spring batch. With this change we'd have to remember to remove this line when we remove MigrateReportConfiguration.

Not a big deal, just wondering about the reason?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -6,102 +6,49 @@
...
-        return factory.getObject();
+    public JobExecutionListener listener() {
+        return new MigrateJobCompletionListener();
     }
 
     @Bean
     public Job migrateReportingJob() {
         try {
             return jobBuilderFactory.get("Migrate Reporting Job")
-                    .repository(jobRepository())
```
malaffoon: So it's just using the default job repository now?

migrate-reporting/src/main/resources/application.yml
chdietz: I am open to that.  Left this in because I plan on doing additional analysis of transactions.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
chdietz: Yes, originally added it when trying to dictate the datasource to use and found it was not needed.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingApplication.java
chdietz: This really just makes all the batch beans available on the application context.  I am moving towards making MigrateReportingConfiguration really just a job configuration.  We could have multiple child jobs and a parent job.

PR #134, opened 2017-05-10
Schema placeholder in SQL yml

agorina:


migrate-reporting/src/test/resources/application-test.yml
```diff
@@ -20,4 +25,10 @@ spring:
...
     job:
-      enabled: false
\ No newline at end of file
+      enabled: false
+
+logging:
+  level:
+    org.springframework.transaction: INFO
+    org.springframework.jdbc.core.JdbcTemplate: DEBUG
+    org.springframework.jdbc.datasource.DataSourceTransactionManager: INFO
```
agorina: Found it useful during testing too.

migrate-reporting/src/test/resources/application-test.yml
```diff
@@ -1,3 +1,8 @@
+rdw:
+  sql:
+    reporting: "`reporting-test`"
```
malaffoon: seem like a lot of decoration (i.e. why do you need all the quotes?)

migrate-reporting/src/main/resources/application.sql.yml
```diff
@@ -1,13 +1,13 @@
...
 
   warehouse:
     findMaxExamImportId: >-
-      SELECT MAX(import.id)
-        FROM import
-          JOIN import_content ON import.content = import_content.id
-        WHERE import_content.name = 'EXAM'
\ No newline at end of file
+      SELECT MAX(i.id)
+        FROM ${rdw.sql.warehouse}.import i
```
malaffoon: well, this is going to get ugly; fine for a couple SQL strings but ...

migrate-reporting/src/test/resources/application-test.yml
agorina: We choose to use a dash as a separator. MySQL or MySQL driver are not happy with that and require a special character `. Yml treats ` as a special character. So I had to decorate it here too. To solve it we could change "-test" to "_test". But this would require changes in multiple CI related scripts.

migrate-reporting/src/main/resources/application.sql.yml
agorina: Agree, completely. What are you seeing as an alternative?

malaffoon: The only thing that comes to mind is to an actual sed-like replacement of values in the yml files before bundling it for tests. So the sql would look like:
```
    SELECT MAX(i.id)
      FROM warehouse.import i
```
and the build would do the equivalent of `s/warehouse./`warehouse-test`./g`

agorina: Let me look at how easy it would be to do.

chdietz: that is what I was thinking, leave the production sql as is and modify it for test with some parse logic

migrate-reporting/src/test/resources/application-test.yml
chdietz: if MySql has a problem with "-test" should we not nip this in the but now?

migrate-reporting/src/main/resources/application.sql.yml
rcordeau: Could put it in the build.gradle script.

migrate-reporting/src/test/resources/application-test.yml
rcordeau: I don't have a problem with changing it from '-' to '_' in the CI stuff.

agorina: @rcordeau, I am about to submit a PR that handles it with the '-'.
When you have time, if you could replaced it with the '_', I think it would be better.
I was not sure how big of a change this is.

PR #138, opened 2017-05-11
fix problem with app name not showing up

malaffoon:


common/src/main/java/org/opentestsystem/rdw/ingest/common/status/StatusEndpoint.java
```diff
@@ -15,10 +14,7 @@
 @ConfigurationProperties(prefix = "endpoints.status")
 public class StatusEndpoint extends AbstractEndpoint<SystemStatus> {
 
-    @Value("${spring.application.name}")
```
malaffoon: I moved the application name out of here and into the "applicationInfo" section of the LocalSystemStatusIndicator. Didn't necessarily want to but the @Value annotation isn't working properly with ConfigurationProperties and the lack of a PropertyPlaceholder.

PR #140, opened 2017-05-11
updates to fix package parsing issues, conditional validation on subt…

rcordeau:
…ype, and missing scaled score values

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/140)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -81,14 +81,21 @@ public Assessment parse(final CSVRecord record, final DataElementErrorCollector
...
+        if(!record.get(ASSESSMENT_TYPE).trim().equals("ICA")) {
+            assessmentBuilder.cutPoints(parserHelper.basicValidation(CUTPOINT_LOW, record.get(CUTPOINT_LOW).trim(), Doubles::tryParse),
+                    parserHelper.basicValidation(CUTPOINT_MIDDLE, record.get(CUTPOINT_MIDDLE).trim(), Doubles::tryParse),
+                    parserHelper.basicValidation(CUTPOINT_HIGH, record.get(CUTPOINT_HIGH).trim(), Doubles::tryParse));
+
+            assessmentBuilder.minScore(parserHelper.basicValidation(MIN_SCORE, record.get(MIN_SCORE).trim(), Doubles::tryParse));
+
+            assessmentBuilder.maxScore(parserHelper.basicValidation(MAX_SCORE, record.get(MAX_SCORE).trim(), Doubles::tryParse));
+        }
+        //Terrible, awful hack until Tabulator is fixed.
```
rcordeau: Don't judge

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultItemParser.java
```diff
@@ -86,7 +86,7 @@ public Item parse(final CSVRecord record, final DataElementErrorCollector errorC
             itemBuilder.allowCalculator(record.get(ALLOW_CALCULATOR).trim().equals("Yes"));
         }
 
-        if (!record.get(MATH_PRACTICE).isEmpty()) {
+        if (!record.get(MATH_PRACTICE).isEmpty() && !record.get(MATH_PRACTICE).toLowerCase().equals("none")) {
```
rcordeau: Apparently, blank and None are valid values.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -17,8 +17,8 @@
     private final int subjectId;
     private final int schoolYear;
     private final String version;
-    private final double minScore;
-    private final double maxScore;
+    private final Double minScore;
+    private final Double maxScore;
```
bgsavage: What's the difference between Double and double?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
malaffoon: Instead of branching on assessment type you could, for now, have a special validator for all these that extracts the double value if it is there but returns 0.0 if no value is present. Then it would just start working when the tabulator provides values.

rcordeau: Yep, I agree.  Will fix now.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
rcordeau: Object versus primitive

malaffoon: Double can be null, double cannot.

bgsavage: Java is funny to me sometimes :)  

Got it, thanks.

rcordeau: C# doesn't have the same concept?

bgsavage: It does, but you can have nullable primitives by using the `?` qualifier.  But under the hood it's probably just creating an object.  So in this example we would just use `double? maxScore` which makes it more obvious why you made the change.  But I guess this change is obvious already for java dev if you're used to it.  I'm still acclimating...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -113,4 +111,10 @@ private String toSchoolYear(final String value) {
         return split.length == 0 ? value : split[split.length - 1]; //nothing to split, so return the value
         //able to split (e.g. 2016-2017), return last value
     }
+
+    private double toScaleScore(final String value) {
+        Double scaleScore = Doubles.tryParse(value);
```
malaffoon: nit: could be `final`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -113,4 +111,10 @@ private String toSchoolYear(final String value) {
         return split.length == 0 ? value : split[split.length - 1]; //nothing to split, so return the value
         //able to split (e.g. 2016-2017), return last value
     }
+
+    private double toScaleScore(final String value) {
```
malaffoon: 👍

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
rcordeau: fixed.

PR #141, opened 2017-05-11
stubauth

malaffoon:
Sorry to drag everybody into this review but i wanted to make sure you see it in case you need to know how to do authorized calls against import in tests or local dev. This hack introduces a special token parser that will resolve the token locally, setting authority properly. So setting a bearer token in your request doesn't require connectivity to OpenAM or a valid user there, just set a header with the client, username, and tenancy chain that you want:
```
Authorization: Bearer sbac;alice;|SBAC|ASMTDATALOAD|CLIENT|SBAC||||||||||||||
```
If you just need authentication without authority (e.g. hitting status end-point) it can be simpler:
```
Authorization: Bearer ;;
```

The changes actually cleaned tests up a lot and allowed proper testing of auth restrictions on the end-points.


import-service/src/test/java/org/opentestsystem/rdw/ingest/web/ExamControllerIT.java
```diff
@@ -113,31 +101,42 @@ public void itShouldUseServiceToGetImportsByBatch() throws Exception {
...
                 .andExpect(content().string(containsString(status.toString())))
                 .andExpect(content().string(containsString("123")));
     }
 
     @Test
-    public void itShouldReturn4xxForInvalidStatus() throws Exception {
-        mvc.perform(get("exams/imports?status=fubar"))
-                .andExpect(status().is4xxClientError());
+    public void itShouldReturnBadRequestForInvalidStatus() throws Exception {
+        mvc.perform(get("/exams/imports?status=fubar").header(AuthHeader, AuthHeaderWithAuthority))
```
malaffoon: This test was broken ... it was return 404 because path was wrong, not 400 as expected.

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/ExamControllerIT.java
```diff
@@ -113,31 +101,42 @@ public void itShouldUseServiceToGetImportsByBatch() throws Exception {
...
-    public void itShouldReturn4xxWhenNoCriteria() throws Exception {
-        mvc.perform(get("exams/imports"))
-                .andExpect(status().is4xxClientError());
+    public void itShouldReturnBadRequestWhenNoCriteria() throws Exception {
+        mvc.perform(get("/exams/imports").header(AuthHeader, AuthHeaderWithAuthority))
+                .andExpect(status().is(HttpStatus.BAD_REQUEST.value()));
+    }
+
+    @Test
+    public void itShouldReturnUnauthorizedIfNoAuthToken() throws Exception {
```
malaffoon: Can now properly test auth conditions on end-points.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/StubTokenServices.java
```diff
@@ -0,0 +1,66 @@
...
+ * client-id;username;tenancy-chain</pre>
+ * Some examples:<pre>
+ * sbac;alice;|SBAC|ASMTDATALOAD|CLIENT|SBAC||||||||||||||
+ * sbac;bob;
+ * sbac;; </pre>
+ * Token for alice grants ASMTDATALOAD authority.<br/>
+ * Token for bob is good for authentication but grants no authorities.<br/>
+ * The last token is a client token good for authentication but having no user (and thus no authorities).<br/>
+ * <p>
+ * In case it isn't obvious: <b>this is for testing and local development ONLY</b>.
```
chdietz: How do we make sure this does not run in production?

malaffoon: Make sure production is configured properly. There are a ton of properties that can be set that disable security in spring, so this is a general issue.

PR #142, opened 2017-05-11
First try at adding codes migration.

agorina:
This is just a first pass. 
After I talk to @malaffoon I want to consider how the other steps will be done to avoid refactoring later. Because this may change, I did not finish all the tests. Will do once I have a plan defined.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcMigrateCodesFromStagingRepository.java
```diff
@@ -0,0 +1,51 @@
...
+class JdbcMigrateCodesFromStagingRepository implements MigrateCodesFromStagingRepository {
+
+    @Autowired
+    @Qualifier("reportingJdbcTemplate")
+    private NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Autowired
+    private MigrateStagingToWarehouseSqlConfiguration sqls;
+
+    private final String[] codes = {"gender", "accommodation", "subject_claim_score", "item_difficulty_cuts", "administration_condition",
```
agorina: Just noticed that I am missing grades here. Will add.

PR #146, opened 2017-05-12
Refactor code migration to support future work.

agorina:
Just to explain.
@malaffoon gave me a great idea that "Migration from staging to reporting is nothing more than a collection of SQL statements executed in a certain order". With this concept of mind, the migration steps are all the same and are just some logical grouping of the SQLs. 

Here is what this change does.
A SQL entity (table) could have a set of supported SQL actions.  They are defined in the sql.yml file.  
Ex.: 
```
entities:
        # ------------ grade  -------------------------------------------------------------------
        grade:
          sql:
            update:  >-
                UPDATE ...
            insert: >-
                INSERT INTO ...
            delete: >-
                DELETE ...
```
 The above defines a "grade" entity that has three sql actions: update, insert and delete.

This is loaded into a configuration class that stores a map of entities and their supported SQL actions with the corresponding SQL.

```SqlListExecutionStep``` is a generic step that executes a collection of the given SQLs based on the Import Content Type. 

Now I can define a step by just giving it a list of SQL actions from the entities.
Below is an example:

```
 // Code entities in the order to be migrated
    public static final String[] codesEntities = {"grade", "completeness", "administration_condition", "ethnicity", "gender",
            "accommodation", "claim", "subject_claim_score", "item_difficulty_cuts", "administration_condition", "target",
            "depth_of_knowledge", "math_practice", "item_trait_score", "item_difficulty_cuts"};

 @Bean(name = "stepUpsertCodes")
    public Step stepUpsertCodes() {
        final SqlListBuilder sqlsBuilder = new SqlListBuilder(stagingToWarehouseSqlConfiguration.getEntities());
        for (final String entity : codesEntities) {
            sqlsBuilder.addNext(entity, "update").addNext(entity, "insert");
        }

        return stepBuilderFactory.get(stepUpsertCodesName)
                .tasklet(new SqlListExecutionStep(newArrayList(CODES),
                        sqlListExecutionRepository,
                        sqlsBuilder.build())).build();
    }
````

### CODE REVIEW

File: migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/configuration/DataSourceConfiguration.java [hunk 1]
```diff
@@ -50,8 +50,8 @@ public BatchConfigurer configurer(@Qualifier("datasource") final DataSource data
      *
      * @return DataSource
      */
-    @Bean(name = "reportingJdbcTemplate")
-    public NamedParameterJdbcTemplate reportingJdbcTemplate(@Qualifier("datasource") final DataSource dataSource) {
+    @Bean(name = "queryJdbcTemplate")
```
agorina: Unrelated. Renamed it since it is not just of the reporting, but for the staging too.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcReportingRepository.java
```diff
@@ -12,16 +12,16 @@
...
     @Value("${sql.reporting.findMigrateLastImportId}")
     private String findMigrateLastImportId;
 
     @Override
     @Transactional
     public Integer queryForObject() {
-        return reportingJdbcTemplate
+        return queryJdbcTemplate
                 .getJdbcOperations()
                 .queryForObject(findMigrateLastImportId, Integer.class);
```
agorina: This will be removed in the following reviews. It was here to set an example.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlListExecutionRepository.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public void execute(final Iterable<String> sqls) {
+
+        for (final String sql : sqls) {
+            jdbcTemplate
+                    .getJdbcOperations()
+                    .execute(sql);
+        }
+    }
+}
```
agorina: I am not sure if this repository has a value. 
One one hand it is nothing than a pass-through to the jdbcTemplate. 
On the other hand, without having it, we had a hidden problem with two transaction managers and it was only exposed by moving the code into its own repo. 

Would like to know what people think about it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/configuration/DataSourceConfiguration.java
chdietz: yes but it is the reporting database connection

agorina: Talked offline with @chdietz. For now I am renaming it back to the ```reportingJdbcTemplate```.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -25,18 +28,30 @@
...
+    private final SqlListExecutionRepository sqlListExecutionRepository;
+    private final MigrateStagingToWarehouseSqlConfiguration stagingToWarehouseSqlConfiguration;
 
     public static final String stepReportingName = "stepReporting";
     public static final String stepWarehouseName = "stepWarehouse";
     public static final String stepUpsertCodesName = "stepUpsertCodes";
     public static final String stepDeleteCodesName = "stepDeleteCodes";
 
+    // Code entities in the order to be migrated
+    public static final String[] codesEntities = {"grade", "completeness", "administration_condition", "ethnicity", "gender",
```
malaffoon: I'd suggest calling out any of this order that is required if any, for example:
```
    psfs[] codesEntities = {
        "grade",    // must be before "ethnicity"
        "completeness",
        "administration_condition",  // must be before "claim"
    ...
```
Of course, if there are no such order restrictions, i think that should be said too.

agorina: Good point. Will do.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -77,14 +92,27 @@ public Step stepWarehouse(final Warehouse warehouse) {
     }
 
     @Bean(name = "stepUpsertCodes")
```
malaffoon: Not urgent but, as discussed yesterday, i'd prefer that "step" be at the end of the name/method. Partly a personal preference but i believe it is the Spring convention to do it that way. So, in this case:
```
    @Bean(name="upsertCodesStep")
    public Step upsertCodesStep() {
```
Btw, do you need the `name` parameter if it is exactly the same as the method name? Can we just do that instead of the duplication?

agorina: I agree.
But based on my conversation with @chdietz about the other names, I would like to wait for his feedback on this comment before renaming it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlListExecutionRepository.java
chdietz: If you are saying you have this so we are consistent using jdbcTemplate from a @Repository then i agree.

agorina: In this case I will keep it. Thank you.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
chdietz: not that you mention it we normally have things like findSubjectRepository not repositoryFindSubject so I agree

agorina: In this case I am renaming them.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlListExecutionStep.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.batch.repeat.RepeatStatus;
+
+/**
+ * A generic step for executing {@link SqlListExecutionStep#sqls} in the given order if
+ * the {@link MigrateBatch} has at least one content from the {@link SqlListExecutionStep#supportedContents}
+ */
+public class SqlListExecutionStep extends StepTasklet {
+
+    private SqlListExecutionRepository repository;
```
malaffoon: `final`?

agorina: Have no idea how I missed it:) 
Will add with the next PR.

PR #147, opened 2017-05-12
Step and repository for checking critical code tables before starting…

chdietz:
Step to check code tables.


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
```diff
@@ -0,0 +1,63 @@
...
+public class JdbcWarehouseToStageRepositoryIT extends RepositoryIT {
+
+    @Autowired
+    WarehouseToStageRepository repository;
+
+    @Autowired
+    private JdbcTemplate jdbcTemplate;
+
+    @Test
+    public void shouldFindWarehouseSubjects() throws Exception {
```
agorina: Did you mean to insert some data before the test?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
```diff
@@ -0,0 +1,63 @@
...
+        final List<Map<String, Object>> subjects = repository.findWarehouseSubjects();
+        assertThat(countRowsInTable(jdbcTemplate, "`warehouse-test`.subject")).isEqualTo(subjects.size());
+    }
```
agorina: Since it is important here that the data are sorted, I would probably add a test showing that this is how it is done.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseCheckIT.java
```diff
@@ -0,0 +1,84 @@
...
+    private void shouldHaveFailed(JobExecution jobExecution) {
+        StepExecution stepExecution = jobExecution.getStepExecutions().iterator().next();
+
+        assertThat( stepExecution.getStatus()).isEqualTo(BatchStatus.FAILED);
+        assertThat( stepExecution.getStepName()).isEqualTo(stepWarehouseCheckName);
+        assertThat( jobExecution.getStatus()).isEqualTo(BatchStatus.FAILED);
+    }
+
+    @Test
+    public void shouldPassCheckWithDefaultCodes() {
```
agorina: And here too, did you want to load some data?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -76,6 +80,12 @@ public Step stepWarehouse(final Warehouse warehouse) {
                 .tasklet(warehouse).build();
     }
 
+    @Bean(name = "stepWarehouseCheck")
```
agorina: Nit: based on the discussion in my last review this should be renamed and there is no need for the 'name=...'.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
chdietz: Yes, true, but this is tested in the step test, if not the compare would fail, but I could add a test for the order of id's

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
chdietz: yeah, was going to do that after I fix this merge conflict
and then get rid of name=

PR #148, opened 2017-05-12
change the be an name for a step

chdietz:
refactor step name


agorina:
Did you say "make Mark happy"? That is an ambitious goal! :)

bgsavage:
I aspire to the day where I check in code and mark is happy :)

PR #149, opened 2017-05-13
DWR-358

malaffoon:
add end-point for posting form-data (file upload)
accept arbitrary properties for payload
capture filename with properties
add end-point for getting payload
add end-point for getting payload properties

I didn't intend to change anything with RDW_Schema link; i'll have to fiddle and revert that. Actually, @rcordeau if you have advice please let me know: i guess i didn't do a recursive pull before creating my branch so i must now have an older pin of the the schema?

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -36,14 +36,13 @@ public class ExamProcessorConfiguration {
     @ServiceActivator(inputChannel = Sink.INPUT)
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
-        final String payload = (String) message.getPayload();
+        final byte[] payload = (byte[]) message.getPayload();
         final long importId = accessor.getImportId();
 
         logger.info("received {} import {}", accessor.getContent(), importId);
-        logger.debug(accessor.getContentType() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
 
         try {
-            final TDSReport report = toTDSReport(new ByteArrayInputStream(payload.getBytes("UTF-8")), importId);
+            final TDSReport report = toTDSReport(new ByteArrayInputStream(payload), importId);
             if (report == null) return;
 
             process(report, importId);
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
@@ -13,8 +13,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
-import java.nio.charset.Charset;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 
@@ -32,7 +32,6 @@ class DefaultImportService implements ImportService {
     private final RdwImportRepository repository;
     private final ArchiveService archiveService;
     private final ImportSource source;
-    private final Charset utf8 = Charset.forName("UTF-8");
 
     @Autowired
     DefaultImportService(final RdwImportRepository repository,
@@ -44,7 +43,12 @@ class DefaultImportService implements ImportService {
     }
 
     @Override
-    public RdwImport importContent(final SbacUserDetails user, final String payload, final ImportContent content, final String contentType, final String batch) {
+    public RdwImport importContent(final SbacUserDetails user,
+                                   final byte[] payload,
+                                   final ImportContent content,
+                                   final String contentType,
+                                   final Map<String, String> metadata) {
+
         final String digest = DigestUtils.md5Hex(payload).toUpperCase();
 
         RdwImport rdwImport = repository.findOneByDigest(digest);
@@ -58,10 +62,13 @@ class DefaultImportService implements ImportService {
         properties.setProperty(ContentTypeProperty, contentType);
         properties.setProperty(UsernameProperty, user.getUsername());
         properties.setProperty(TenancyChainProperty, user.getTenancyChain().toString());
-        if (batch != null) {
-            properties.setProperty("batch", batch);
+        if (metadata != null) {
+            metadata.forEach((key, value) -> {
+                if (value != null) properties.setProperty(key, value);
+            });
         }
-        archiveService.writeResource(location, payload.getBytes(utf8), properties);
+
+        archiveService.writeResource(location, payload, properties);
 
         rdwImport = repository.create(RdwImport.builder()
                 .content(content)
@@ -69,7 +76,7 @@ class DefaultImportService implements ImportService {
                 .digest(digest)
                 .status(ImportStatus.ACCEPTED)
                 .creator(user.getUsername())
-                .batch(batch)
+                .batch(metadata == null ? null : metadata.get("batch"))
                 .build());
 
         source.submitContent(payload, content, contentType, rdwImport.getId());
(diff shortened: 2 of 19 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -36,14 +36,13 @@ public ExamProcessorConfiguration(final TDSReportProcessor reportProcessor,
     @ServiceActivator(inputChannel = Sink.INPUT)
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
-        final String payload = (String) message.getPayload();
+        final byte[] payload = (byte[]) message.getPayload();
```
malaffoon: As part of this, i've deferred knowing the payload type as long as possible. So the message queue now has byte content instead of Strings.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -58,18 +62,21 @@ public RdwImport importContent(final SbacUserDetails user, final String payload,
         properties.setProperty(ContentTypeProperty, contentType);
         properties.setProperty(UsernameProperty, user.getUsername());
         properties.setProperty(TenancyChainProperty, user.getTenancyChain().toString());
-        if (batch != null) {
-            properties.setProperty("batch", batch);
+        if (metadata != null) {
+            metadata.forEach((key, value) -> {
+                if (value != null) properties.setProperty(key, value);
```
malaffoon: `Properties` doesn't tolerate null values.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -53,6 +94,30 @@ public RdwImportResource getImport(@PathVariable final Long id) {
         return assembler.toResource(service.getImport(id).get());
     }
 
+    @GetMapping("/imports/{id}/payload")
+    public ResponseEntity<Void> downloadImportPayload(@PathVariable final long id, final HttpServletResponse response) {
+        final byte[] payload = service.getPayload(id).get();
+        final Properties properties = service.getPayloadProperties(id).get();
+
+        final String filename = properties.containsKey("filename") ? properties.getProperty("filename") : (id + ".xml");
```
malaffoon: oh yeah, i was going to extract some constants for these ... will do that before commit

import-service/src/main/resources/application.yml
```diff
@@ -50,6 +50,11 @@ spring:
     initialSize: 4
     maxWait: 10000
 
+  http:
+    multipart:
+      max-file-size: 32MB
```
malaffoon: Bump up the max file size a bit (default is 1MB).

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/ExamControllerIT.java
```diff
@@ -64,16 +86,16 @@ public void itShouldUseServiceToImportExam() throws Exception {
...
+        final byte[] body = "<TDSReport/>".getBytes();
+        when(importService.importContent(any(), eq(body), eq(ImportContent.EXAM), eq(MediaType.APPLICATION_XML_VALUE), any())).thenThrow(UnsupportedOperationException.class);
         mvc.perform(post("/exams/imports").header(AuthHeader, AuthHeaderWithAuthority).contentType(MediaType.APPLICATION_XML).content(body))
                 .andExpect(status().is4xxClientError());
     }
 
     @Test
     public void itShouldUseServiceToGetImport() throws Exception {
         final long id = 123;
-        given(importService.getImport(id)).willReturn(Optional.of(testImport(id)));
```
malaffoon: Not really sure where i got this `given` thing from ... changed all to normal mockito `when`.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -43,16 +45,15 @@ public DefaultAssessmentPackageProcessor(final AssessmentParser assessmentParser
     }
 
     @Override
-    public void process(final String assessmentPackage, final long importId) {
-        final DataElementErrorCollector errorCollector = new DataElementErrorCollector();
-        try {
-            final Iterable<CSVRecord> records = CSVFormat.RFC4180.withFirstRecordAsHeader().parse(new StringReader(assessmentPackage));
-
+    public void process(final byte[] assessmentPackage, final long importId) {
+        // at this point we need a Reader so assume UTF-8 payload
```
malaffoon: @rcordeau, if you want to talk about this let me know. Basically, the payloads are bytes until they are parsed. The CSV reader needs to be told the encoding, so we're assuming UTF-8 here. It is no worse than what we were doing before (assuming UTF-8 during post request) but it is buried deeper now.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
chdietz: This sounds ok, but what is the benefit?
As for the RDW_Schema link you mention I had to remove the folder and then:
  git submodule update ( Rob's advice )

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
chdietz: That would be nice.

import-service/src/main/resources/application.yml
chdietz: So we will allow up to 32MB per TRT?
If this is the we should probably document in the API, and other restrictions if there are any.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
malaffoon: Benefit? It's not a benefit thing so much as a correctness thing. Until we open the payload and look at it, we can't know that it can even be a string.

import-service/src/main/resources/application.yml
malaffoon: The setting applies to all import requests, including the packages. Tabulator output for all the assessment packages could easily be a couple of MB. You are correct in implying that a TRT payload is unlikely to be more than a couple hundred kB.

Yeah, this should make it into the documentation.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -36,14 +36,13 @@ public ExamProcessorConfiguration(final TDSReportProcessor reportProcessor,
...
         final long importId = accessor.getImportId();
 
         logger.info("received {} import {}", accessor.getContent(), importId);
-        logger.debug(accessor.getContentType() + ": " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
 
         try {
-            final TDSReport report = toTDSReport(new ByteArrayInputStream(payload.getBytes("UTF-8")), importId);
+            final TDSReport report = toTDSReport(new ByteArrayInputStream(payload), importId);
             if (report == null) return;
```
agorina: So at this point we know that they byte[] is in UTF-8?

malaffoon: XML parsing knows to look at the header and determine encoding.

agorina: And if it is not UTF-8, then do we still support it or it knows how to handle it? Just curious.

malaffoon: Yup. I tested passing a non-UTF8 document and it parsed just fine. Of course, if the encoding is not supported by the platform it will complain. That's a situation where we'd want to install additional encodings and then replay the data (assuming, of course, that we want to support it).

agorina: Thank you for explaining.

PR #150, opened 2017-05-13
DWR-359

malaffoon:
skip the report if there have been no exam imports
allow email to have multiple "to" recipients

This came from talking to Venu. He doesn't think a report should be sent if no imports were attempted in the query period. He also said they would want multiple recipients; i know we could push back and tell them to use a distribution but we're going to want to send the emails to ops ... and it was an easy change.

### DIFF
--- a/task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
+++ b/task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
@@ -13,7 +13,7 @@ public class ReconciliationReportProperties {
         return query;
     }
 
-    public void setQuery(String query) {
+    public void setQuery(final String query) {
         this.query = query;
     }
 
@@ -21,22 +21,22 @@ public class ReconciliationReportProperties {
         return email;
     }
 
-    public void setEmail(Email email) {
+    public void setEmail(final Email email) {
         this.email = email;
     }
 
     @ConfigurationProperties
     public static class Email {
-        private String to;
+        private String[] to;
         private String from;
         private String subject;
         private String message;
 
-        public String getTo() {
+        public String[] getTo() {
             return to;
         }
 
-        public void setTo(String to) {
+        public void setTo(final String[] to) {
             this.to = to;
         }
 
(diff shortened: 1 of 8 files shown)


task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/task/ReconciliationReportProperties.java
```diff
@@ -13,54 +13,54 @@ public String getQuery() {
...
 
-    public void setEmail(Email email) {
+    public void setEmail(final Email email) {
         this.email = email;
     }
 
     @ConfigurationProperties
     public static class Email {
-        private String to;
+        private String[] to;
```
malaffoon: Apart from fixing the tests, this is all it took to allow multiple "to" recipients.

task-service/src/main/resources/application.sql.yml
```diff
@@ -7,7 +7,7 @@ sql:
         JOIN exam_student es ON es.id = e.exam_student_id
         JOIN student s ON s.id = es.student_id
         JOIN asmt a ON a.id = e.asmt_id
-      WHERE (:content IS NULL OR i.content=:content)
+      WHERE (i.content=1)
```
malaffoon: Thoughts? I could query the import content table for the value that matches "EXAM"; is that any less hard-coded than the magic number?

task-service/src/main/resources/application.yml
```diff
@@ -19,7 +19,7 @@ spring:
 task:
   send-reconciliation-report:
     cron: 0 0 14 * * *
-    query: content=EXAM&status=PROCESSED&after=-PT24H
```
malaffoon: I removed the explicit content value because it is being ignored and it might confuse people into thinking they can specify different content.

task-service/src/main/resources/application.sql.yml
agorina: I am okay with the magic number (as long as we have a comment here).
However I am wondering if we need an exact count. 
I guess I missed it in the previous reviews. 
My concern is that with the large amount of data the count is slow.
Do we really need an exact count? 
Can we put a limit to (100000+1) and say that the result is larger than supported?

malaffoon: Good point. I could even go so far as to make it an exists(query) method.

agorina: Exists would be even better. Thank you.

PR #152, opened 2017-05-14
A temp fix to a failed build.

agorina:


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/StagingToReportingStepsConfigIT.java
```diff
@@ -27,9 +24,8 @@
 import static org.opentestsystem.rdw.ingest.migrate.reporting.StagingToReportingStepsConfig.stepDeleteCodesName;
 import static org.opentestsystem.rdw.ingest.migrate.reporting.StagingToReportingStepsConfig.stepUpsertCodesName;
 
-@RunWith(SpringRunner.class)
-@SpringBootTest
```
agorina: This is running a test with the real app config that does not replace the table names to "..-test". This is a temp fix, I will revisit it on Monday.

PR #154, opened 2017-05-14
Remove Reporting step.

agorina:
This step was a placeholder until we start working on the real steps.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/ReportingRepository.java
```diff
@@ -1,6 +0,0 @@
-package org.opentestsystem.rdw.ingest.migrate.reporting.repository;
-
-//TODO - remove it, just for testing now
```
malaffoon: TODONE!

PR #156, opened 2017-05-15
DWR-361

malaffoon:
truncate item response data to 64k

There is a chance we'll do something different to solve this problem (once we get more feedback from SmarterBalanced) but this seems like a safe change to make to at least get the results in.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
```diff
@@ -45,7 +45,7 @@
         for (final Item item : items) {
```
malaffoon: unrelated, fixing warnings

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcIabExamRepository.java
```diff
@@ -18,6 +18,8 @@
 import java.sql.Timestamp;
 import java.util.List;
 
+import static org.opentestsystem.rdw.ingest.processor.repository.impl.JdbcUtils.safeText;
+
```
bgsavage: nit: extra space :)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
```diff
@@ -45,7 +45,7 @@
...
             final String response = (item.getResponse() != null && !isNullOrEmpty(item.getResponse().getContent())) ? item.getResponse().getContent() : null;
             final ScoreInfoType scoreInfo = item.getScoreInfo();
-            final String naturalId = new Long(item.getBankKey()).toString() + "-" + item.getKey();
+            final String naturalId = Long.toString(item.getBankKey()) + "-" + item.getKey();
```
bgsavage: Do you think it would make sense for there to be a getter on the item model that "gets" the natural id and does this string concatenation there?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcIabExamRepository.java
malaffoon: fixed

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
malaffoon: Good idea. I was just fixing the warning though, not the logic.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
```diff
@@ -70,11 +70,10 @@
 
         for (final ScoreInfoType subScore : scoreInfo.getSubScoreList().getScoreInfo()) {
             final String trait = subScore.getScoreDimension();
-            if (scoreInfo == null || !itemTraitIds.containsKey(trait)) continue;
+            if (!itemTraitIds.containsKey(trait)) continue;
```
jtreuting: The `scoreInfo == null` check is no longer needed for some reason?

malaffoon: The diff doesn't show it but there is guard code just above here that checks `scoreInfo == null`

bgsavage: Would it be cleaner and more/less performant to stream and filter the `ScoreInfoType` collection based on this condition instead of using `for` / `continue` ?   (This is an honest question, I don't know the answer :) )

malaffoon: I agree it would be cleaner (more readable). It would definitely be less performant (streams suck), but i'm not sure it would matter in this context of preparing data since the IO will be orders of magnitude slower.

PR #157, opened 2017-05-15
Remove all entities and upsertOrganizations.

agorina:
The flow (at least for now) is to handle all the 'delete' first, and then process updates/inserts.
This adds the 'delete' all entities step and a step to update/insert School and Districts.
The tests for 'delete' is not completed, because I was lazy to load all the data. I am going to keep adding to that test as more migration is done.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/StagingToReportingStepsConfig.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/StagingToReportingStepsConfig.java
@@ -13,6 +13,8 @@ import java.util.List;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.ingest.common.model.ImportContent.CODES;
+import static org.opentestsystem.rdw.ingest.common.model.ImportContent.EXAM;
+import static org.opentestsystem.rdw.ingest.common.model.ImportContent.PACKAGE;
 
 /**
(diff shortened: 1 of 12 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/StagingToReportingStepsConfig.java
```diff
@@ -36,6 +40,56 @@ public StagingToReportingStepsConfig(final StepBuilderFactory stepBuilderFactory
         this.sqlListExecutionRepository = sqlListExecutionRepository;
     }
 
+    // main entities in the order to be migrated due to the dependencies
+    public static final List<String> entities = newArrayList(
+            "district",
+            "school",
+
```
agorina: An empty line separates Steps that I am planning on doing.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/ReportingMigrateJobIT.java
```diff
@@ -24,8 +24,10 @@
     private JdbcTemplate jdbcTemplate;
 
     @SqlGroup({
+            //TODO: this should be changed to be handled by the migration from warehouse to staging
             //loads all code tables with one row each with id = -98
```
agorina: @chdietz, this is the test for the entire Job and the comment I mentioned to you this morning.

PR #162, opened 2017-05-16
factor out common import controller code

malaffoon:
There is already some duplication between Exam and Package import controllers. And we anticipate more controllers (accessibility codes, districts, schools). So i pulled out the common functionality.


malaffoon:
Sigh, not sure how RDW_Schema got in here ... i was really friggin' careful about refreshing before branching. I think i really don't like git submodules.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
```diff
@@ -19,135 +17,66 @@
...
-        return assembler.toResource(service.importContent(sbacUser, body, ImportContent.EXAM, contentType, requestParams));
+        return super.postImport(sbacUser, contentType, body, requestParams);
     }
 
     @PostMapping(value = "/imports", consumes = "multipart/form-data")
     @ResponseStatus(HttpStatus.ACCEPTED)
     public RdwImportResource uploadImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
                                           final MultipartHttpServletRequest request) {
-        // get the best MultipartFile (either the only one or the one named "file")
-        final MultipartFile file;
```
chdietz: So, you created common logic to handle support for posting body and file upload?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ReportingPackageController.java
```diff
@@ -27,14 +31,11 @@
...
-class ReportingPackageController {
-
-    private final ImportService service;
-    private final ReportingPackageImportResourceAssembler assembler = new ReportingPackageImportResourceAssembler();
+class ReportingPackageController extends ImportController {
 
     @Autowired
     public ReportingPackageController(final ImportService service) {
-        this.service = service;
+        super(service, ImportContent.PACKAGE, new ImportResourceAssembler(ReportingPackageController.class));
```
chdietz: now package loader can support file upload?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ExamController.java
malaffoon: That was the main driving force, yes. But also the payload and payload/properties end-points.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ReportingPackageController.java
malaffoon: Yup.

rcordeau: Yeah they all will - Accommodation codes as well.

rcordeau:
I need this update, so I'll merge it in.

PR #164, opened 2017-05-16
Feature/accessibility processor

rcordeau:
This is just for the accommodation codes controller and some other clean up and refactoring of common code.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/164)


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
```diff
@@ -82,7 +82,10 @@ public void configure(final HttpSecurity http) throws Exception {
                 http.authorizeRequests()
                     .antMatchers("/exams/imports/resubmit").authenticated()
                     .antMatchers("/exams/**").hasAuthority(DataLoadAuthority)
+                    .antMatchers("/packages/imports/resubmit").authenticated()
```
malaffoon: i'm really beginning to think we should flip the path ... sigh

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
```diff
@@ -0,0 +1,85 @@
...
+@RestController
+@RequestMapping({ "/accommodations" })
+class AccommodationController extends ImportController {
+
+    @Autowired
+    public AccommodationController(final ImportService service) {
+        super(service, ImportContent.CODES, new ImportResourceAssembler(AccommodationController.class));
+    }
+
+    @PostMapping(value = "/imports", consumes = "application/xml")
```
malaffoon: i took the `consumes` off. Why? Because the payload is opaque to this service so i think it can be more permissive and let stuff through.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
```diff
@@ -0,0 +1,85 @@
...
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public RdwImportResource postImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
+                                        @RequestHeader(value = CONTENT_TYPE, required = false) final String contentType,
+                                        @RequestBody final byte[] body,
+                                        @RequestParam(required = false) final String batch) {
+        return super.postImport(sbacUser, contentType, body, singletonMap("batch", batch));
+    }
+
+    @PostMapping(value = "/imports", consumes = "multipart/form-data")
```
malaffoon: Of course, this `consumes` has to be here because it indicates the file upload.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationImportResourceAssembler.java
```diff
@@ -0,0 +1,31 @@
...
+import org.opentestsystem.rdw.ingest.common.model.RdwImport;
+import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
+
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
+
+/**
+ * A {@link org.springframework.hateoas.ResourceAssembler} for {@link RdwImportResource}s
+ */
+class AccommodationImportResourceAssembler extends ResourceAssemblerSupport<RdwImport, RdwImportResource> {
```
malaffoon: Did you need this? Can't you just instantiate the ImportResourceAssembler with the controller class?

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/AccommodationsControllerIT.java
```diff
@@ -0,0 +1,240 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@WebMvcTest(AccommodationController.class)
+@ContextConfiguration(classes = TestAppConfig.class)
+@WebAppConfiguration
+public class AccommodationsControllerIT {
+
+    private final String ACCOMMODATIONS_URI = "/accommodations/imports";
```
malaffoon: 👍   i should do that in the other ones

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationImportResourceAssembler.java
rcordeau: This is mistake - left over class that I don't use.  I thought I removed it. My bad.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
rcordeau: It is supposed to be a xml document, so wouldn't you want to narrow it down right away? If a csv comes in, I won't process it because I'll be expecting xml.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
rcordeau: /exams/resubmit/imports?

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/AccommodationsControllerIT.java
rcordeau: It was more helpful when I had it as a generic controller tester and I had an array of URIs that I looped through, but I liked that it make it look cleaner.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
malaffoon: That puts the knowledge about the payload in two places, which is not as good as not doing that. Let's say the format for the codes changes to CSV tomorrow. Clearly we have to change the codes processor to deal with that, but why make it so we have to change this?

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
malaffoon: I lost sleep last night thinking about this. We can discuss offline but i'm leaning towards removing the content from the URLs that don't need it (e.g. `GET /imports/{id}`, `GET /imports/{id}/payload`) and flipping the order for ones that do (e.g. `POST /imports/exams`). For things like resubmit, i would argue that the content is a filter so it should be a param, not in the URL path (e.g. `POST /imports/resubmit?content=EXAM`), although i could be talked out of that.

PR #165, opened 2017-05-17
Migrator scheduler

agorina:
This is about https://jira.fairwaytech.com/browse/DWR-336, and it is not a complete solution. I want to see what people think before I continue with it.

It does the following things:
1. Puts a batch execution in a continuous non-stop loop.
2. Stops a batch from running if another one is running. 
3. Has a TODO to configure batch steps to repeat them until there is nothing to process.

There are some issues with this solution:
1. How would someone stop the process?  Should we connect it to the Spring boot 'shutdown' endpoint somehow?  http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-application-exit?

2. If a batch was 'killed' without recording a FAILED or COMPLETED status, it thinks that another version is running. Pretty annoying. See more comments in the code changes.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/BatchRunner.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/BatchRunner.java
@@ -1,0 +1,6 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
+
+public interface BatchRunner {
+     void run(String[] args);
+}
+
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
@@ -1,0 +1,60 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.core.Job;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobExecutionException;
+import org.springframework.batch.core.JobParameter;
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
+import org.springframework.batch.core.launch.JobLauncher;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Import;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static java.lang.Thread.sleep;
+
+@Component
+@EnableBatchProcessing
+@Import({MigrateReportingConfiguration.class})
+public class MigrateBatchRunner implements BatchRunner {
+    private static final Logger logger = LoggerFactory.getLogger(MigrateReportingApplication.class);
+
+    @Autowired
+    private JobLauncher jobLauncher;
+
+    @Value("${migrate_batch_runner.sleep:50000}")
+    private long sleepMillis;
+
+    public void setSleepMillis(final long sleepMillis) {
+        this.sleepMillis = sleepMillis;
+    }
+
+    @Autowired
+    @Qualifier("migrateReportingJob")
+    private Job job;
+
+    public void run(String[] args) {
+        try {
+            while (true) {
+                final Map<String, JobParameter> jobParametes = newHashMap();
+                jobParametes.put("time", new JobParameter(System.currentTimeMillis()));
+                final JobExecution jobExecution = jobLauncher.run(job, new JobParameters(jobParametes));
+
+                logger.info("Completed job with id [" + jobExecution.getJobId() + "]");
+                sleep(sleepMillis);
+            }
+        } catch (JobExecutionException ex) {
+            logger.error("Job execution exception [" + ex.getMessage() + "]");
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
+}
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
@@ -10,12 +10,16 @@ import org.springframework.batch.core.JobExecutionListener;
 import org.springframework.batch.core.Step;
 import org.springframework.batch.core.configuration.annotation.JobBuilderFactory;
 import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
-import org.springframework.batch.core.launch.support.RunIdIncrementer;
+import org.springframework.batch.core.explore.JobExplorer;
+import org.springframework.batch.core.explore.support.JobExplorerFactoryBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.jdbc.support.lob.DefaultLobHandler;
+
+import javax.sql.DataSource;
 
 /**
  * MigrateReportingConfiguration.
@@ -45,6 +49,21 @@ public class MigrateReportingConfiguration {
         return new MigrateJobCompletionListener();
     }
 
+//    @Bean
+//    public SimpleJobOperatorFactor JobExplorerFactoryBean(@Qualifier("datasource") final DataSource dataSource) throws Exception {
+//        JobExplorerFactoryBean jobExplorerFactoryBean = new JobExplorerFactoryBean();
+//        jobExplorerFactoryBean.setDataSource(dataSource);
+//        jobExplorerFactoryBean.setLobHandler(new DefaultLobHandler());
+//        return jobExplorerFactoryBean;
+//    }
+
+    @Bean
+    public JobExplorerFactoryBean JobExplorerFactoryBean(@Qualifier("datasource") final DataSource dataSource) throws Exception {
+        JobExplorerFactoryBean jobExplorerFactoryBean = new JobExplorerFactoryBean();
+        jobExplorerFactoryBean.setDataSource(dataSource);
+        jobExplorerFactoryBean.setLobHandler(new DefaultLobHandler());
+        return jobExplorerFactoryBean;
+    }
 
     /**
      * This job migrates data from the 'warehouse' to 'reporting' databases.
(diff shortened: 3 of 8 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Autowired
+    @Qualifier("migrateReportingJob")
+    private Job job;
+
+    public void run(String[] args) {
+        try {
+            while (true) {
+                final Map<String, JobParameter> jobParametes = newHashMap();
+                jobParametes.put("time", new JobParameter(System.currentTimeMillis()));
```
agorina: It looks like it needs some unique value every time, otherwise it finds an existing batch.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -45,6 +49,21 @@ public JobExecutionListener listener() {
...
     }
 
+//    @Bean
+//    public SimpleJobOperatorFactor JobExplorerFactoryBean(@Qualifier("datasource") final DataSource dataSource) throws Exception {
+//        JobExplorerFactoryBean jobExplorerFactoryBean = new JobExplorerFactoryBean();
+//        jobExplorerFactoryBean.setDataSource(dataSource);
+//        jobExplorerFactoryBean.setLobHandler(new DefaultLobHandler());
+//        return jobExplorerFactoryBean;
+//    }
+
```
agorina: Oops, this is not needed.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -92,8 +112,9 @@ public Job migrateReportingJob(@Qualifier("warehouseStep") final Step warehouse,
         try {
 
             return jobBuilderFactory.get("Migrate Reporting Job")
-                    .incrementer(new RunIdIncrementer())
                     .listener(listener())
+                    .validator(new SingletonJobParameterValidator(jobExplorer, "Migrate Reporting Job"))
+                    //TODO: configure the batch step to loop until there is something to migrate
```
agorina: Everything that I read about the batch says that it is meant to repeat steps until it is done. So I feel that this would be the right approach.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
```diff
@@ -0,0 +1,30 @@
...
+        this.jobName = jobName;
+    }
+
+    @Override
+    public void validate(final JobParameters parameters) throws JobParametersInvalidException {
+        Set executions = jobExplorer.findRunningJobExecutions(jobName);
+        if (executions != null && executions.size() > 1) {
+            //it's running somewhere
+            throw new JobParametersInvalidException("the job " + jobName + " is currently running");
+        }
```
agorina: This makes it pretty restrictive. There is also a batch admin API that actually has a method to remove 'stailed' jobs:
https://github.com/spring-projects/spring-batch-admin/blob/master/spring-batch-admin-manager/src/main/java/org/springframework/batch/admin/service/SimpleJobService.java#L609

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
chdietz: Are you saying the job will run again or the steps will repeat until done?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Map;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static java.lang.Thread.sleep;
+
+@Component
+@EnableBatchProcessing
+@Import({MigrateReportingConfiguration.class})
+public class MigrateBatchRunner implements BatchRunner {
```
chdietz: I was thinking we would investigate @EnableScheduling + Batch such as this example:
https://github.com/sidnan/spring-batch-scheduler-example

You can use fixedDelay to "the period will be measured from the completion time of each preceding invocation."

If we can make spring batch keep running until there is no work to do plus a fixed delay from completion of previous invocation this may work?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
chdietz: If we stop a job because another is running, maybe it should require manual intervention because this should not happen.  I would prefer a scheduling mechanism that knows if a job is running and starts one at the appropriate time. Unless of course we allowed multiple jobs to run at the same time and we are not doing that.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
chdietz: couple of thoughts on: "How would someone stop the process?"

1. Pause / resume: Tell scheduler to not start any more jobs and later tell it to start again.
1.1 may be better just to do a graceful shut down and then start again when desired.

2. Graceful shut down.  Tell scheduler to not start any more jobs and shut down.
2.1 actuator has shut down capabilities. maybe we could tie into that.

3. Shutdown and stop running job.
3.1 maybe actuator again, tie into it. 
3.2 it would first to spring batch to abort any running job, if there is a spring batch way of doing this.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
agorina: I am sorry, but I do not know if I read you comments right. Are you saying that you would prefer this to be done outside of the Job and within a scheduling?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
chdietz: Ah, working late :)
I am suggesting using the @Scheduled from Spring:
https://docs.spring.io/spring/docs/current/spring-framework-reference/html/scheduling.html

And use @Scheduled(fixedDelay='some-time')
But embed it in migrate-reporting, not leverage task-service module.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
chdietz: If we use a fixedDelay schedule two Job's running at the same time would be an exceptional and unexpected state, if it works the way I understand it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
chdietz: If the last step circles back to the first this would work well. 
If we did this it would be nice to put in the first step a check to see if a graceful shutdown has been requested.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
agorina: Is it smart enough to know that if two independent services are started, then only one job should be allowed to run?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
agorina: I like this idea.

chdietz: My last thought on this is what is the 'best practice' for spring batch 
1. loop back to step 1 to check if it should run again
2. or run the job again to check if it should run again and complete the job if not.
3. or there is no 'best practice' in this regard.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SingletonJobParameterValidator.java
chdietz: I suspect from what I read it is design like this but I have not tested this or read too deep into it.
I do agree with you we should also check in the batch execution.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/BatchRunner.java
```diff
@@ -0,0 +1,6 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
+
+public interface BatchRunner {
```
malaffoon: Haven't read the whole review yet but i would tend towards using `Runnable` with the args passed in to the constructor of the impl that is providing Runnable.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
malaffoon: This comment thread gives the reasons for not using @Scheduled:
# we want to immediately run again under some circumstances, based on the return value
# we want to be able to pause

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Autowired
+    @Qualifier("migrateReportingJob")
+    private Job job;
+
+    public void run(String[] args) {
+        try {
+            while (true) {
+                final Map<String, JobParameter> jobParametes = newHashMap();
```
malaffoon: typo in name

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateBatchRunner.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public void setSleepMillis(final long sleepMillis) {
+        this.sleepMillis = sleepMillis;
+    }
+
+    @Autowired
+    @Qualifier("migrateReportingJob")
+    private Job job;
+
+    public void run(String[] args) {
```
malaffoon: see, you're not even using the args

malaffoon:
It feels like we should discuss this vs. trying to figure out in comment threads.

agorina:
Agree.

PR #166, opened 2017-05-17
change imports end-points

malaffoon:
Please vote on this change. 

The import service has evolved as more import tasks have been added. The end-points for getting the import resources included the `content` (e.g. EXAM, PACKAGE). That is odd since the import doesn't really care what content it represents. So, this patch changes the URL for those end-points that don't care about content. For example, all these go away:
```
GET ./exams/imports/{id}
GET ./packages/imports/{id}/payload
GET ./accommodations/imports/{id}/payload/properties
```
and get replaced with the generic end-points:
```
GET ./imports/{id}
GET ./imports/{id}/payload
GET ./imports/{id}/payload/properties
```

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
@@ -23,6 +23,6 @@ public enum ImportContent {
         for (final ImportContent status : ImportContent.values()) {
             if (value == status.value) return status;
         }
-        throw new IllegalArgumentException("Unknown ImportContent value.");
+        throw new IllegalArgumentException("Unknown ImportContent value: " + value);
     }
 }
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
@@ -9,7 +9,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
 import org.springframework.web.bind.annotation.GetMapping;
-import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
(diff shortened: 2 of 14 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
```diff
@@ -23,6 +23,6 @@ public static ImportContent fromValue(final int value) {
         for (final ImportContent status : ImportContent.values()) {
             if (value == status.value) return status;
         }
-        throw new IllegalArgumentException("Unknown ImportContent value.");
+        throw new IllegalArgumentException("Unknown ImportContent value: " + value);
```
malaffoon: Unrelated; i got in a state where i had some invalid content indicators and the message without `value` was rather useless.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
```diff
@@ -19,32 +18,30 @@
...
     }
 
     @PostMapping(value = "/imports")
     @ResponseStatus(HttpStatus.ACCEPTED)
     public RdwImportResource postImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
                                         @RequestHeader(value = CONTENT_TYPE, required = false) final String contentType,
                                         @RequestBody final byte[] body,
-                                        @RequestParam(required = false) final String batch) {
-        return super.postImport(sbacUser, contentType, body, singletonMap("batch", batch));
+                                        @RequestParam final Map<String, String> requestParams) {
```
malaffoon: Unrelated; allow any metadata to be specified.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ImportController.java
```diff
@@ -1,103 +1,45 @@
 package org.opentestsystem.rdw.ingest.web;
```
malaffoon: This looks like a bunch of changes but it is really a brand new class for handling the generic end-points.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/ScopedImportController.java
```diff
@@ -0,0 +1,100 @@
+package org.opentestsystem.rdw.ingest.web;
```
malaffoon: This looks like a new class but it is the renamed abstract controller class.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
rcordeau: Did you get the latest "develop" after I merged? I made that change.

rcordeau:
👍

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
malaffoon: Thought i did get latest. Hmm.

rcordeau: Well no matter - I get what you're saying about it. No need to have knowledge of the content in two places.

chdietz:
if your new pattern mirror the request to create the import then I am ok with it.

PR #167, opened 2017-05-17
add organizations end-points

malaffoon:
I believe the organization data needs to be processed all at once so i created a single end-point / import content. The alternative would be to have DISTRICT, SCHOOL, etc. but then there would be order-of-operation and resolution problems.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
@@ -7,7 +7,8 @@ public enum ImportContent {
     PROBE(0),
     EXAM(1),
     PACKAGE(2),
-    CODES(3);
+    CODES(3),
+    ORGANIZATION(4);
 
     private final int value;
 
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
```diff
@@ -7,7 +7,8 @@
     PROBE(0),
     EXAM(1),
     PACKAGE(2),
-    CODES(3);
+    CODES(3),
+    ORGANIZATION(4);
```
malaffoon: Anybody changing the data load scripts wanna add this for me? Turns out the app runs fine without adding it to the reference table but i do need to add it ...

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/AccommodationControllerIT.java
```diff
@@ -43,7 +43,7 @@
 @WebMvcTest(AccommodationController.class)
 @ContextConfiguration(classes = TestAppConfig.class)
 @WebAppConfiguration
-public class AccommodationsControllerIT {
+public class AccommodationControllerIT {
```
malaffoon: Unrelated; fix typo

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/ImportContent.java
agorina: I got the hint. Will add it.

PR #171, opened 2017-05-18
Added service, repo and tests for external get id's for job

chdietz:
1. Added new repo to get id's from warehouse and reporting
2. created object just to hold id's
3. created service that handles logic of id's and build list of id's for next batch job
4. designed to be called before spring batch to simplify job

a. need to move max batch size to properties

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/MigrateImportRepository.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/MigrateImportRepository.java
@@ -1,0 +1,13 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting.repository;
+
+import java.util.Optional;
+
+public interface MigrateImportRepository {
+
+    Optional<Long> findMaxImportId();
+
+    Optional<Long> findMinImportId();
+
+    Optional<Long> findLastMigratedImportId();
+
+}
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobService.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobService.java
@@ -1,0 +1,53 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting.service;
+
+
+import org.opentestsystem.rdw.ingest.migrate.reporting.MigrateBatchId;
+import org.opentestsystem.rdw.ingest.migrate.reporting.repository.MigrateImportRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+import java.util.Optional;
+
+@Service
+public class DefaultNextJobService implements NextJobService {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultNextJobService.class);
+
+    // todo: move this to properties.
+    private static final Long BatchSizeLimit = 10L;
+
+    private final MigrateImportRepository migrateImportRepository;
+
+    public DefaultNextJobService(MigrateImportRepository migrateImportRepository) {
+        this.migrateImportRepository = migrateImportRepository;
+    }
+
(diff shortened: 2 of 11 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobService.java
```diff
@@ -0,0 +1,53 @@
...
+            return Optional.of(new MigrateBatchId(last.get() + 1,
+                    (max.get() - last.get()) <= BatchSizeLimit ? max.get() : last.get() + BatchSizeLimit));
+        } else {
+            // If there has been no migration start with the first.
+            Optional<Long> first = migrateImportRepository.findMinImportId();
+            if (!first.isPresent()) {
+                return Optional.empty();
+            } else {
+                return Optional.of(new MigrateBatchId(first.get(),
+                        (max.get() - first.get()) < BatchSizeLimit ? max.get() : first.get() + (BatchSizeLimit - 1)));
```
agorina: The ids are not true sequence. 
There are could be gaps in them. 
We either need to make the ids true sequence or change this logic to do something like this:
```
select max(id) id from
  (SELECT id
   FROM import import
   WHERE id > 10
   ORDER BY id ASC
   LIMIT 100
) as a;
```
Since id is indexed I think the above should be fast and easier than doing this math. 
In the above example, the value 10 would come from the reporting/migrate table, and 100 would be something configurable for the batch size.

chdietz: ok thought it was sequence.  
So we have id's not in sequence, but we know there relative order.

We still have to account for edge cases the first 
- there are no migrate records.
- there are no import records.
I do not want to any assumptions about the state of warehouse.import or reporting.migrate.

For an edge case where there is no migrate record we could have another query
`select max(id) id from
  (SELECT id
   FROM import import
   ORDER BY id ASC
   LIMIT 100
) as a;`

agorina: Makes sense, I agree with your comment.

chdietz: if i have time i'll work on this change before I leave

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobService.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Optional;
+
+@Service
+public class DefaultNextJobService implements NextJobService {
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultNextJobService.class);
+
+    // todo: move this to properties.
+    private static final Long BatchSizeLimit = 10L;
```
agorina: This should be configurable.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobService.java
agorina: Sounds good, thank you.

agorina:
Do you mind merging it in anyway so that I can use it even if you do not get to change it before leaving?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/MigrateImportRepository.java
```diff
@@ -0,0 +1,13 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting.repository;
+
+import java.util.Optional;
+
+public interface MigrateImportRepository {
+
+    Optional<Long> findMaxImportId();
```
bgsavage: nit: Maybe these should be named 'getMaxImportId' instead of find since you're not taking any parameters.

PR #179, opened 2017-05-22
Feature/migrate exams

agorina:
This completes the 'happy path' for migrating from staging to reporting.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -102,7 +103,9 @@ public Job migrateReportingJob(final Step warehouseStep,
                     .next(upsertOrganizationsStep)
                     .next(upsertAsmtsStep)
                     .next(upsertStudentsAndGroupsStep)
+                    .next(upsertExamsStep)
                     .next(deleteCodesStep)
+                    //TODO: plug in a step to clean up the staging
```
agorina: This step should be repeated before and after.

PR #180, opened 2017-05-22
Changes to the migrate to have its own id

agorina:
it is addressing a code review comment.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/School.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/School.java
@@ -49,7 +49,7 @@ public class School {
 
     @Override
     public int hashCode() {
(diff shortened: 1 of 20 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/School.java
```diff
@@ -49,7 +49,7 @@ public boolean equals(final Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(District.class, name, naturalId, district.hashCode());
+        return Objects.hash(School.class, name, naturalId, district.hashCode());
```
agorina: Unrelated, just found that it was wrong.

agorina: Done.

PR #181, opened 2017-05-23
Change service and repository to account for import id not in sequence

chdietz:
@agorina you may want to merge your feature/migrate-id first and then I will check for conflicts.

1. updated repo to account for id's not in true sequence
2. modified service to account for repo change

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseImportRepository.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseImportRepository.java
@@ -4,7 +4,8 @@ import java.util.Optional;
 
 public interface WarehouseImportRepository {
 
-    Optional<Long> findMaxImportId();
-
     Optional<Long> findMinImportId();
+
+    Optional<Long> findMaxImportIdNotMigrated(Long lastImportId, Integer batchSize);
+
 }
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
@@ -1,10 +1,14 @@
 package org.opentestsystem.rdw.ingest.migrate.reporting.repository.impl;
 
 import org.opentestsystem.rdw.ingest.migrate.reporting.repository.WarehouseImportRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.stereotype.Repository;
 
 import java.util.Optional;
@@ -12,11 +16,13 @@ import java.util.Optional;
 @Repository
 public class JdbcWarehouseImportRepository implements WarehouseImportRepository {
 
-    @Value("${sql.warehouse.import.findMaxImportId}")
-    private String findMaxImportId;
+    private static final Logger logger = LoggerFactory.getLogger(JdbcWarehouseImportRepository.class);
 
     @Value("${sql.warehouse.import.findMinImportId}")
     private String findMinImportId;
+
+    @Value("${sql.warehouse.import.findMaxImportIdNotMigrated}")
+    private String findMaxImportIdNotMigrated;
 
     private NamedParameterJdbcTemplate warehouseJdbcTemplate;
 
@@ -26,19 +32,25 @@ public class JdbcWarehouseImportRepository implements WarehouseImportRepository 
     }
 
     @Override
-    public Optional<Long> findMaxImportId() {
+    public Optional<Long> findMinImportId() {
         try {
-            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMaxImportId, Long.class));
+            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMinImportId, Long.class));
         } catch (IncorrectResultSizeDataAccessException ex) {
             return Optional.empty();
         }
     }
 
     @Override
-    public Optional<Long> findMinImportId() {
+    public Optional<Long> findMaxImportIdNotMigrated(Long lastImportId, Integer batchSize) {
+
         try {
-            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMinImportId, Long.class));
-        } catch (IncorrectResultSizeDataAccessException ex) {
+            final SqlParameterSource parameterSource = new MapSqlParameterSource()
+                    .addValue("last_import_id", lastImportId)
+                    .addValue("batch_size", batchSize);
+            return Optional.ofNullable(warehouseJdbcTemplate.queryForObject(findMaxImportIdNotMigrated, parameterSource, Long.class));
+
+        } catch (final IncorrectResultSizeDataAccessException ignore) {
+            logger.warn("Incorrect number of results found with lastImportId [" + lastImportId + "] and batchSize [" + batchSize + "] ");
             return Optional.empty();
         }
     }
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamService.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamService.java
@@ -6,6 +6,7 @@ import org.opentestsystem.rdw.ingest.migrate.reporting.repository.ReportingMigra
 import org.opentestsystem.rdw.ingest.migrate.reporting.repository.WarehouseImportRepository;
 import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParameters;
+import org.springframework.dao.DataRetrievalFailureException;
 import org.springframework.stereotype.Service;
 
 import java.util.Map;
@@ -13,6 +14,9 @@ import java.util.Optional;
 
 import static com.google.common.collect.Maps.newHashMap;
 
+/**
+ *  Default implementation of NextJobParamService to find the first and last import ID's for a migrate batch.
+ */
 @Service
 class DefaultNextJobParamService implements NextJobParamsService {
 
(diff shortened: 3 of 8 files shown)


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepositoryIT.java
```diff
@@ -16,35 +16,62 @@
...
-        assertThat(max)
+    public void itShouldFindSetupMinImportId() {
+        final Optional<Long> min = repository.findMinImportId();
+        assertThat(min)
                 .isPresent()
-                .hasValue(defaultMaxImportId);
+                .hasValue(setupMinImportId);
     }
 
     @Sql(statements = "DELETE FROM warehouse_test.import")
```
chdietz: This works but have issues when using sql in same manager to delete specific record(s) based on id.  May be related to transactions.

migrate-reporting/src/main/resources/application.sql.yml
```diff
@@ -15,9 +15,18 @@ sql:
...
-          SELECT MIN(i.id) FROM warehouse.import i WHERE i.status >= 0
+        SELECT MIN(i.id) FROM warehouse.import i WHERE i.status >= 0
+
+      findMaxImportIdNotMigrated:  >-
+        SELECT max(available.id) id
+        FROM
+          (SELECT id
+           FROM warehouse.import
+           WHERE id > :last_import_id
+            AND status >= 0
```
agorina: I suggest to change to ```status = 1```. 
We may want to discuss with @malaffoon his vision for status  ```PUBLISHED``` (=2), but regardless it feels safer to use an explicit status.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepositoryIT.java
agorina: I am curious what the problem is.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseImportRepository.java
```diff
@@ -4,7 +4,8 @@
 
 public interface WarehouseImportRepository {
 
-    Optional<Long> findMaxImportId();
-
     Optional<Long> findMinImportId();
```
malaffoon: We haven't discussed it but there is a pattern where the repository layer does not use Optional, just returning null. The service layer, if appropriate, introduces the Optional.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
```diff
@@ -26,19 +32,25 @@ public JdbcWarehouseImportRepository(final NamedParameterJdbcTemplate warehouseJ
     }
 
     @Override
-    public Optional<Long> findMaxImportId() {
+    public Optional<Long> findMinImportId() {
         try {
-            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMaxImportId, Long.class));
+            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMinImportId, Long.class));
         } catch (IncorrectResultSizeDataAccessException ex) {
```
malaffoon: nit: `final`

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
```diff
@@ -26,19 +32,25 @@ public JdbcWarehouseImportRepository(final NamedParameterJdbcTemplate warehouseJ
...
+
         try {
-            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMinImportId, Long.class));
-        } catch (IncorrectResultSizeDataAccessException ex) {
+            final SqlParameterSource parameterSource = new MapSqlParameterSource()
+                    .addValue("last_import_id", lastImportId)
+                    .addValue("batch_size", batchSize);
+            return Optional.ofNullable(warehouseJdbcTemplate.queryForObject(findMaxImportIdNotMigrated, parameterSource, Long.class));
+
+        } catch (final IncorrectResultSizeDataAccessException ignore) {
```
malaffoon: Hmm. If we get this exception, that is really, really bad right? That means we've borked something in the code? Should we be propagating it instead of eating it?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
```diff
@@ -26,19 +32,25 @@ public JdbcWarehouseImportRepository(final NamedParameterJdbcTemplate warehouseJ
     }
 
     @Override
-    public Optional<Long> findMaxImportId() {
+    public Optional<Long> findMinImportId() {
         try {
-            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMaxImportId, Long.class));
+            return Optional.ofNullable(warehouseJdbcTemplate.getJdbcOperations().queryForObject(findMinImportId, Long.class));
         } catch (IncorrectResultSizeDataAccessException ex) {
             return Optional.empty();
```
malaffoon: Should we be distinguishing between incorrect result size and empty result set? Seems like empty result set is an expected corner case, while incorrect result size that is not an empty result set is ... bizarre. And might deserve a more drastic reaction.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamService.java
```diff
@@ -25,34 +29,57 @@
...
             Optional<Long> first = warehouseImportRepository.findMinImportId();
             if (!first.isPresent()) {
-                return Optional.empty();
+                return Optional.empty(); // If there are no migrate and no import records.
+            } else {
+                Optional<Long> maxFromImport = warehouseImportRepository.findMaxImportIdNotMigrated(first.get() - 1, batchSize);
+                if (maxFromImport.isPresent()) {
+                    return createJobParams(batchSize, first.get(), maxFromImport.get());
+                } else {
+                    throw new DataRetrievalFailureException("findMaxImportIdNotMigrated returned no results after returning findMinImportId");
```
malaffoon: I'd suggest that a Service should not be throwing a data exception.

migrate-reporting/src/main/resources/application.sql.yml
malaffoon: 👍

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamServiceTest.java
```diff
@@ -27,132 +39,92 @@ public void setUp() {
...
-        assertThat(jobParameters.getLong(JobParams.fistImportId)).isEqualTo(-1007L);
-        assertThat(jobParameters.getLong(JobParams.lastImportId)).isEqualTo(-1002L);
-        assertThat(jobParameters.getLong(JobParams.batchSize)).isEqualTo(10L);
+        assertThat(jobParameters.getParameters().get("jobIdentifier").getType()).isEqualTo(JobParameter.ParameterType.LONG);
+        assertThat(jobParameters.getParameters().get("jobIdentifier").isIdentifying()).isTrue();
+        assertThat(jobParameters.getLong(JobParams.fistImportId)).isEqualTo(firstInBatch);
+        assertThat(jobParameters.getLong(JobParams.lastImportId)).isEqualTo(lastInBatch);
+        assertThat(jobParameters.getLong(JobParams.batchSize)).isEqualTo(defaultBatchSize);
+
+        verify(reportingMigrateRepository, times(1)).findLastMigratedImportId();
```
malaffoon: Not sure if you're doing it to be explicit but `times(1)` is the default for verify and not needed.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamService.java
bgsavage: I'd second that the method names shouldn't be in the exception message (that will be in the call stack and will be more up to date should the implementation change)

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
chdietz: ok

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepositoryIT.java
chdietz: will have to show you this

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseImportRepository.java
chdietz: ok, I have read a few posts after you mention this that it is an anti-pattern,
I will re evaluate this.

migrate-reporting/src/main/resources/application.sql.yml
chdietz: ok, will change, so we want 2 for published and exclude 1=PROCESSED

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
chdietz: I think you may be correct.  Maybe I should let this percolate up and design a test that properly causes this exception.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/service/DefaultNextJobParamService.java
chdietz: both good points will make adjustments

migrate-reporting/src/main/resources/application.sql.yml
malaffoon: Just want to clarify what you mean by that last comment, bit worred by "exclude 1=PROCESSED". Perhaps by stating what i understand:
0=ACCEPTED, payload captured, ready for import
1=PROCESSED, import work complete, *ready for migration*
2=PUBLISHED, migrate work complete
So it is precisely `status=1` that should be in the filter isn't it?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepositoryIT.java
chdietz: Issue was @SQL class level and at method level.
appears to be stated here : https://stackoverflow.com/questions/32871817/using-annotation-sql-is-it-possible-to-execute-scripts-in-class-level-before-m

migrate-reporting/src/main/resources/application.sql.yml
agorina: @malaffoon - I agree with 1=PROCESSED and this is what we need to include for the migration. I do not understand how 'PUBLISHED' will be set though.

malaffoon: That is left-over. Perhaps it will never be used.

chdietz: ok, sorry, I was typing while not thinking, I will make it `status=1`

PR #183, opened 2017-05-24
Changes to support update import id.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -53,6 +53,7 @@ public long create(final Exam exam) {
                 .addValue("exam_student_id", studentExamId)
                 .addValue("school_year", exam.getSchoolYear())
                 .addValue("import_id", exam.getImportId())
+                .addValue("upd_import_id", exam.getImportId())
```
chdietz: upd vs update, where do we draw the line on abbreviation?
I prefer less abbreviation.

agorina: I really do not have a strong opinion here.

jtreuting: If anyone asked me (which they didn't), i would go for "update_import_id".  Personally i don't mind longer variables or column names as often they are easier to understand

agorina: Ok, majority wins :) Will rename to ```update_import_id```.

bgsavage: Should we rename import_id to created_import_id ?  Also I was thinking both should be past tense.

malaffoon: We intentionally chose not to change `import_id`. 

As for the tense, the way i read this is "the import_id associated with the update", not "the updated import_id".

malaffoon:
I'm confused how the RDW_Schema pointer could be moved since the RDW_Schema changes aren't merged to develop yet.

agorina:
@malaffoon , I was confused too. Maybe @rcordeau could clarify? I was not going to merge until I merge the schema, but I was wondering the same.

PR #191, opened 2017-05-25
Feature/dwr 368

malaffoon:
Adds a couple lifecycle controls to running the migrate job:
 - pause/resume based on spring app running state
 - enable/disable based on state of previous migrate attempts

Adds diagnostic status indicator.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobExecutionListener.java
```diff
@@ -1,8 +1,5 @@
-package org.opentestsystem.rdw.ingest.migrate.reporting.listener;
```
malaffoon: @agorina i moved this out of the `listener` package; it is tightly coupled to the rest of the stuff in the main package and it was the only thing in the listener package.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateLifecycle.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
```
malaffoon: Even though the job runner does its own disabling, i added this interface so that we could, if we wanted, control that behavior from outside the job runner.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -16,42 +17,101 @@
...
     public void run() {
+        if (!checkRunJobs()) return;
 
         try {
             Optional<JobParameters> jobParameters = nextJobParamsService.getNextJobParams(batchSize);
             while (jobParameters.isPresent()) {
-
                 jobLauncher.run(job, jobParameters.get());
+
+                if (!checkRunJobs()) break;
```
malaffoon: Hmm, i'm still fiddling with this logic. Not sure i have it right since this will check twice in a cycle even if there is only one batch of things to do. I'll get back to you on this ...

migrate-reporting/src/main/resources/application.yml
```diff
@@ -1,7 +1,11 @@
+management:
+  security:
+    enabled: false
+
 migrate:
   batch:
     size: 1000
-    delay: 50000
```
malaffoon: @agorina was there something magical about 50s? Any reason to not change this to an even minute?

migrate-reporting/src/main/resources/application.yml
```diff
@@ -26,6 +30,11 @@ spring:
     job:
       enabled: false
 
+  jackson:
```
malaffoon: So Status is marshaled properly.

migrate-reporting/src/test/resources/MigrateImportFromWarehouseSetup.sql
```diff
@@ -6,8 +6,8 @@ WHERE id = 1;
 -- Completed migration import id's in reporting
 -- First migrate completed does not exist in warehouse, just for testing
 INSERT INTO reporting_test.migrate (id, job_id, status, first_import_id, last_import_id) VALUES
-  (-1, 1, 20, -1020, -1011),
-  (-2, 2, 20, -1010, -1001);
+  (-10, 1, 20, -1020, -1011),
```
malaffoon: i needed to make room to insert a more recent entry

migrate-reporting/src/test/resources/application-test.yml
```diff
@@ -1,20 +1,3 @@
 spring:
-
   batch_datasource:
     url: jdbc:mysql://localhost:3306/spring_batch_reporting_test?useSSL=false
-    username: root
```
malaffoon: All of this was duplicative from application.yml; the only difference for test is the schema.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/ReportingMigrateRepository.java
```diff
@@ -3,17 +3,15 @@
...
  * Repository for managing migrate table in the reporting datamart
  */
 public interface ReportingMigrateRepository {
 
     /**
-     * Returnst the last migrate import id
+     * @return the last migrate import id, null if no successful migrations
      */
-    Optional<Long> findLastMigratedImportId();
+    Long findLastMigratedImportId();
```
malaffoon: Unrelated to the changes but since i was in here ...

migrate-reporting/src/main/resources/application.yml
agorina: I was impatient when testing it :) I think it should be more like 5 min?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/Migrate.java
```diff
@@ -71,7 +75,15 @@ public Migrate build() {
...
             return migrate;
+        }
 
+        public Builder copy(final Migrate migrate) {
+            id = migrate.id;
+            firstImportId = migrate.firstImportId;
+            lastImportId = migrate.lastImportId;
+            jobId = migrate.jobId;
+            status = migrate.status;
+            return this;
```
agorina: Thank you! I know I was lazy and took a shortcut when doing it.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobExecutionListener.java
agorina: ok

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
malaffoon: Okay, did this instead:
```
    public void run() {
        try {
            Optional<JobParameters> jobParameters;
            while ((jobParameters = nextJobParamsService.getNextJobParams(batchSize)).isPresent() && checkRunJobs()) {
                jobLauncher.run(job, jobParameters.get());
            }
        } catch (final Exception e) {
            logger.error("migrate reporting job failure [" + e.getMessage() + "]");
        }
    }
```
The only minor issue is that, if there is no work to do, the system will never discover that it should be disabled.

agorina: This looks better.

migrate-reporting/src/main/resources/application.sql.yml
```diff
@@ -1,24 +1,32 @@
...
+        DELETE FROM reporting.migrate WHERE id = :id
+
+      updateStatusById: >-
+        UPDATE reporting.migrate SET status = :status
+          WHERE id = :id
+
+      lastStatus: >-
+        SELECT status FROM reporting.migrate ORDER BY id DESC LIMIT 1
+
+      lastStatusAlt: >-
```
malaffoon: I need to figure out which of these two is more better.

PR #196, opened 2017-05-27
Feature/truncate step

chdietz:


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -106,14 +110,20 @@ public Job migrateReportingJob(final Step warehouseStep,
                     .next(upsertStudentsAndGroupsStep)
                     .next(upsertExamsStep)
                     .next(deleteCodesStep)
-                    //TODO: plug in a step to clean up the staging
+                    .next(truncateStageStep)
```
chdietz: I created a truncate step an call it at the start and end.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseCheck.java
```diff
@@ -37,7 +37,6 @@ public WarehouseCheck(WarehouseToStageRepository warehouseToStageRepository,
     @Override
     public RepeatStatus execute(final StepContribution contribution, final ChunkContext chunkContext) throws Exception {
         if (checkSubject() && checkAsmtTypes() && checkSubjectClaimScores()) {
-            warehouseToStageRepository.truncateStaging(warehouseToStagingSqlConfiguration.getTruncateStageList());
```
chdietz: removed truncate from the other step

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
```diff
@@ -41,13 +41,19 @@ public void itShouldDeleteButNotUpsertCodes() {
...
+                        -1 * jdbcTemplate.queryForObject("SELECT count(0) FROM reporting_test." + table + " rid WHERE NOT EXISTS(SELECT accommodation_id FROM staging_test.staging_" + table + " WHERE accommodation_id = rid.accommodation_id)", Integer.class),
+                        "1=1"));
             } else if ("math_practice".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table, -1, "1=1"));
+                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table,
+                        -1 * jdbcTemplate.queryForObject("SELECT count(0) FROM reporting_test." + table + " rid WHERE NOT EXISTS(SELECT practice FROM staging_test.staging_" + table + " WHERE practice = rid.practice)", Integer.class),
+                        "1=1"));
             } else {
-                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table, -1, "1=1"));
+                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table,
```
chdietz: The test ReportingMigrateJobIT moves codes in warehouse from initial DML to reporting and does not delete them.
If this test runs first there are more codes in reporting than this test knows about.  And I think I have this correct when codes are processed all codes not in stage are deleted.

chdietz: I am not sure about:

- The order of IT test execution.
- Can multiple tests execute at the same time in one test class or multiple classes, for IT tests.
- Note: gradle has maxParallelForks and I am not sure of the default
- Note: I could not get `./gradlew build IT` to fail from my environment but it failed on build server.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
```diff
@@ -41,13 +41,19 @@ public void itShouldDeleteButNotUpsertCodes() {
         // collect counts of rows in each `reporting` code table before the step
         final List<TableTestCountHelper> tableTestCounts = newArrayList();
         for (final String table : codesEntities) {
-            // existing row must be deleted and nothing new should be added
-            if ("administration_condition".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table, -1, "1=1"));
+            // existing rows must be deleted if they do not exist in staging
+            if ("accommodation_translation".equals(table)) {
```
chdietz: We have two code tables that do not have `id` as the primary key?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
malaffoon: Tests should be made to not care about order of execution. In a good CI environment, tests are run in random order each time. 

We are not parallelizing IT execution. And i don't think we really need to.

agorina: "_The test ReportingMigrateJobIT moves codes in warehouse from initial DML to reporting and **does not delete them**_." - why doesn't it delete it after the test?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
agorina: We have a few tables that do not have a primary key. Those are the tables that represent many-to-many or many-to-one relationships. Do we you see it as a problem?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
chdietz: @malaffoon Yes, I agree test should run in any order.  I think the problem I was seeing showed up in a specific order.

As for parallelizing I do no think we can.  We can not parallelize tests that are integration tests against the same schema and doing verification by counting rows.  Especially since one of the tests is a job test that runs truncate tables.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
chdietz: ok, i was wonder why the convention looked different.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
chdietz: @agorina 'What to do with codes in initial DML' we talked about this onThursday and I though we said we would leave them in reporting after a test run.  We may need to discuss this

agorina: Yes, you are right. I put a comment on my comment  above :)

PR #197, opened 2017-05-27
Feature/dwr 298 accessibility codes from trt

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/197)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
```diff
@@ -189,5 +196,10 @@ public B examItems(final List<ExamItem> examItems) {
             this.examItems = examItems;
             return (B) this;
         }
+
+        public B accommodationIds(final List<Integer> accommodationIds) {
+            this.accommodationIds = accommodationIds;
```
malaffoon: copy list?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AccommodationRepository.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * AccommodationRepository - repository for dealing with accommodations made available in an exam
+ */
+public interface AccommodationRepository {
+    /**
+     * Finds an id of the accommodation by its code
+     *
+     * @param code the code of the accommodation
+     * @return the id
+     * @throws IllegalArgumentException is the given code is not found
```
malaffoon: not true, it returns null if code is not found

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAccommodationRepository.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return the id
+     * @throws IllegalArgumentException is the given code is not found
+     */
+    @Override
+    @Cacheable(value = "accCode")
+    public Integer findIdByCode(String code){
+        try {
+            return jdbcTemplate.queryForObject(findIdByCodeSql, new MapSqlParameterSource("code", code), Integer.class);
+        } catch (final EmptyResultDataAccessException ignore) {
+            logger.warn("unknown accommodation code [" + code + "]");
```
malaffoon: i don't think this should be logging anything: the repository just returns null, it doesn't know if that is worth logging or not since the client context is unknown.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -135,4 +139,17 @@ private void batchCreateExamItems(final List<ExamItem> examItems, final long exa
         jdbcTemplate.batchUpdate(sqlExamItemCreate, batchParameters);
     }
 
+    private void batchAddAccommodations(final List<Integer> accommodationIds, final long examId) {
+        if (accommodationIds.isEmpty()) return;
+
+        final MapSqlParameterSource[] batchParameters = new MapSqlParameterSource[accommodationIds.size()];
```
malaffoon: i know it is slightly less efficient but i prefer to use java collections as long as possible; in this case, collect the batch parameters in a java collection and then convert to any array just in the batchUpdate() call. As @bgsavage  has been heard to say, "if you find yourself writing a for loop, you should probably change things."

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
```diff
@@ -99,6 +112,13 @@ private E parse(final Test test, final Opportunity opportunity, final StudentExa
                 .opportunity(opportunity.getOpportunity())
                 .sessionId(opportunity.getSessionId());
 
+        for (Opportunity.Accommodation acc : opportunity.getAccommodation()) {
+            Integer id = parserHelper.repositoryValidation(acc.getCode(), acc.getCode(), Strings::emptyToNull, accommodationRepository::findIdByCode);
+            if (id != null) accommodationIds.add(id);
```
malaffoon: So the logging would be here; and would it be `warn` or just `info` or `debug` since we expect to see codes that we don't know about?

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -13,6 +13,10 @@
     private final DataElementErrorCollector elementErrorCollector;
 
 
+    public ParserHelper() {
+        this.elementErrorCollector = new DataElementErrorCollector();
```
malaffoon: Hmm, have to see how this pans out but i'm not fond of it. The pattern is that the client using the ParserHelper owns the error collector; allowing this alternate approach will be confusing i think.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
agorina: Could we validate all accommodations at once instead of one at a time?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
rcordeau: ok, changed to "this.accommodationIds = new ArrayList<>(accommodationIds);"

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/AccommodationRepository.java
rcordeau: Fixed - thanks

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
rcordeau: I kept it similar to the other implementations in the file for consistency.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
rcordeau: Ok - I'll move the logging

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
rcordeau: Ok - based on our off line conversations, I have changed the ParserHelper to remove the unused functions, documented the no arg constructor, handled the case of a null internal error collector, and then instantiated a validationWarnings error collector for validation warnings when doing an AnyExam parse of the accessibility codes, logging any if they exist.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
rcordeau: Validate all at once - ok I'll check it out

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
```diff
@@ -99,6 +116,15 @@ private E parse(final Test test, final Opportunity opportunity, final StudentExa
...
+        }
+
+        if (!validationWarnings.isEmpty()) logger.warn(validationWarnings.toJson());
```
malaffoon: Although they are called `warnings` here i don't think the log level should be `warn`, more like `info`. Remember, these are expected data issues.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/AnyExam.java
bgsavage: @malaffoon Just testing my own understanding here, if the collection was immutable, then copy wouldn't be required, right?  So shouldn't the preferred approach be to use an immutable collection instead? (Sorry that may be a dumb question :) )

malaffoon: I do have minor reservations with immutable collections which we can discuss offline (i don't object to them so much as to not use them). However, this was doing neither making it unsafe.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/AnyExamProcessor.java
rcordeau: Validating all codes at once for one database call. The downside is that we won't know exactly which codes didn't validate, so it will be a general warning that x number of codes didn't validate.

agorina: We should probably add all the codes as part of the warning too.

PR #198, opened 2017-05-28
Added migration of common core standard and item other target.

agorina:
Corresponds to https://github.com/SmarterApp/RDW_Schema/pull/74

I am missing tests for the item common core migrations. Will work on adding it next.


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/DeleteCodesIT.java
```diff
@@ -41,19 +41,18 @@ public void itShouldDeleteButNotUpsertCodes() {
...
         final List<TableTestCountHelper> tableTestCounts = newArrayList();
         for (final String table : codesEntities) {
-            // existing rows must be deleted if they do not exist in staging
-            if ("accommodation_translation".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table,
-                        -1 * jdbcTemplate.queryForObject("SELECT count(0) FROM reporting_test." + table + " rid WHERE NOT EXISTS(SELECT accommodation_id FROM staging_test.staging_" + table + " WHERE accommodation_id = rid.accommodation_id)", Integer.class),
-                        "1=1"));
+            if ("administration_condition".equals(table)) {
+                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table, -1, "id IN (-99)"));
+                tableTestCounts.add(new TableTestCountHelper(jdbcTemplate, "reporting_test." + table, 0, "id IN (-98)"));
```
agorina: @chdietz, I think we should be only testing for the ids we are intending on migrating. Do you agree with this change?

migrate-reporting/src/test/resources/MigrateEntitiesFromStagingSetup.sql
```diff
@@ -22,16 +22,22 @@ INSERT INTO staging_test.staging_asmt_score (asmt_id, cut_point_1, cut_point_2,
   (-98, 2442, 2502, 2582, 2201, 2701, -99);
 
 INSERT INTO staging_test.staging_item (id, claim_id, target_id, natural_id, asmt_id, dok_id, difficulty, max_points, math_practice, allow_calc, migrate_id) VALUES
-  (-990, -99, null, '200-2010',  -99, -99, -0.23, 2, -99, 0, -99),
```
agorina: Target really should not be null.

PR #199, opened 2017-05-29
Added unit tests for item common core migration.

agorina:


migrate-reporting/src/test/resources/MigrateEntitiesPreloadReportingSetup.sql
```diff
@@ -30,6 +30,10 @@ INSERT INTO reporting_test.item_other_target(item_id, target_id) values
  (-990, -98),
  (-991, -98);
 
+INSERT INTO reporting_test.item_common_core_standard(item_id, common_core_standard_id) values
+ (-990, -98),
+ (-991, -98);
+
```
agorina: I added the clean up scripts with the last change, that is the reason I did not have to add it now.

chdietz: common_core_standard is an entity under asmt and would be imported when ImportContent.PACKAGE
- correct?

agorina: ```common_core_standard``` table is preloaded into warehouse. ```item_ common_core_standard``` is loaded with the assessment package. Not sure if this was your question?

chdietz: ok, so:
`common_core_standard ` is treated like a code table then?
or is this another Critical code table?

and:
`item_common_core_standard ` is migrated when asmt is migrated?

agorina: ```common_core_standard``` is treated like a code table.
```item_common_core_standard``` is migrated when asmt is migrated

PR #200, opened 2017-05-30
Feature/migrate import content

chdietz:


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
```diff
@@ -2,48 +2,61 @@
...
 
 import static com.google.common.collect.Lists.newArrayList;
 
+/**
+ *  This step uses the {@link Migrate} param set by the {@link org.opentestsystem.rdw.ingest.migrate.reporting.MigrateJobExecutionListener}
+ *  to query the warehouse.import table and determine the {@link ImportContent}.
+ *  This is save for next steps in the {@link MigrateBatch} param.
+ */
 @Component
 public class Warehouse extends StepTasklet {
```
chdietz: We may want to change the name of this step, it just sets up the MigrateBatch param.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/ReportingMigrateJobIT.java
```diff
@@ -59,23 +61,26 @@ public void itShouldUpsertAndDeleteCodesIfBatchHasImportContentWithCodes() throw
...
 
     @SqlGroup({
-            //TODO: this should be changed to be handled by the migration from warehouse to staging
+            // Import content is implemented from content in batch.
+            @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:ImportContentSetup.sql"}),
+            // Codes migrate from warehouse.
             @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:MigrateCodesFromWarehouseSetup.sql"}),
+            // todo: entities need to be implemented from warehouse.
             @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:MigrateEntitiesFromStagingSetup.sql"}),
             @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:MigrateEntitiesFromStagingForDeleteSetup.sql"}),
```
chdietz: the 'FromStaging' scripts will go away as I finish the entities in warehouse to stage.
well they will go away in this test.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseToStageRepository.java
```diff
@@ -19,4 +17,13 @@
...
     void truncateStaging(Iterable<String> stagingTruncateStatements);
+
+    /**
+     * Find the content types in this batch.
+     *
+     * @param fistImportId First import id of the batch.
+     * @param lastImportId Last import id of the batch.
+     * @return list of content type in the batch.  May be empty.  {@link org.opentestsystem.rdw.ingest.common.model.ImportContent}
+     */
+    List<Integer> findImportContentForBatch(Long fistImportId, Long lastImportId);
```
malaffoon: I don't like the repository method having business logic in its name, i.e. `ForBatch`. What is it really doing and name it that. Also, can it return a List<ImportContent> instead? The fact that it is stored as an integer is a repository implementation details. Finally, i doubt the import ids can be null. So, i'd suggest:
```
    List<ImportContent> findDistictImportContentInRange(long firstImportId, long lastImportId);
```

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
malaffoon: I agree. It should at least have 'step' or 'tasklet' in there.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
```diff
@@ -2,48 +2,61 @@
...
 
     @Value("${sql.warehouseToStage.findMaxExamImportId}")
     private String findMaxExamImportId;
 
-    @Autowired
-    @Qualifier("warehouseJdbcTemplate")
-    private NamedParameterJdbcTemplate warehouseJdbcTemplate;
+    private final WarehouseToStageRepository warehouseToStageRepository;
+
+    public Warehouse(final WarehouseToStageRepository warehouseToStageRepository) {
```
malaffoon: shouldn't this be auto-wired?

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -40,6 +40,13 @@ sql:
       - TRUNCATE TABLE staging.staging_language
       - TRUNCATE TABLE staging.staging_accommodation_translation
 
+    findImportContentForBatch: >-
+      SELECT wi.content from warehouse.import wi
+      WHERE wi.status = 1
+        AND wi.id >= :first_import_id
+        AND wi.id <= :last_import_id
+      GROUP BY wi.content
```
malaffoon: is `GROUP BY` better/worse than `DISTINCT`?

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
```diff
@@ -58,4 +59,73 @@ public void shouldFindReportingSubjectClaimScores() throws Exception {
                 .isEqualTo(subjectClaimScores.size());
     }
 
+    @Sql(statements = {
```
malaffoon: Consider a script file?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
chdietz: hmm how is that working, i converted the field autowire to constructor and must have left that off.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
chdietz: actually I may do that.  I think I can use the same one I made for the steps.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseToStageRepository.java
chdietz: I think i like these suggestions, the Set is actually a good idea.

The list can be empty if the first and last import id's do not find any import records.  Possible but not likely.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
chdietz: I will integrate the name change in my next round of work I have already started.
Actually there are three classes `Warehouse`, `WarehouseCheck` and `TruncateStage` that do not have 'Step' in
the name.  I may do one push that just does name changes.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: Since I only want one distinct may be more concise?
Not sure about performance but the range of import id's i would think will make the query more efficient.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
agorina: I was told that the latest version of Spring will autowire by default if there is only one constructor. But I do like explicit annotation.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: I think we should try to rewrite it with EXISTS. I will post a sample here. And we also should create an index on ```content```.

agorina: After I have added the index:
```
ALTER TABLE import ADD INDEX idx__import_id_content (id, content);
```
I am not sure what if **exists** is better than **distinct**:
```
explain
select ic.id from import_content ic where exists (select id from import i where i.content = ic.id and i.id > 0 and i.id < 5000);

```
content1
```
explain select distinct i.content from import i where i.id > 0 and i.id < 5000
```
content2

```
explain
  select  i.content from import i where i.id > 0 and i.id < 5000
 GROUP BY i.content
```
content3

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/Warehouse.java
chdietz: i agree, may that explains how it was working but will add the Autowire

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: So group by and distinct have similar plans but group by adds a filesort.
- should i go with distinct?

agorina: I would probably go with distinct for now and we can change if needed later. But please add the index.

malaffoon: `GROUP BY` is clearly the worst. Although `EXISTS` approach does two index lookups it doesn't rely on a temporary table at all, so i would favor that one; unfortunate that the SQL is the least readable.

agorina: I am fine with either at this point. I am not convinced one is better than another.

malaffoon: Okay, i yield: `DISTINCT` is best compromise between readability and performance.

chdietz: This is what it looks like:

      SELECT DISTINCT wi.content
      FROM warehouse.import wi
      WHERE wi.status = 1
            AND wi.id >= :first_import_id
            AND wi.id <= :last_import_id

chdietz: i think we need that `status` where in there

agorina: Makes sense. In this case we need to add ```status``` to the index too.

PR #201, opened 2017-05-30
cleanup error message handling a bit

malaffoon:
Goal was to use the collector everywhere. 
Next step would be to have collector contain both errors and warnings.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/DataElementError.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/DataElementError.java
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.ingest.common.error;
+package org.opentestsystem.rdw.ingest.common.util;
 
 import com.google.gson.Gson;
 import java.util.Objects;
@@ -23,10 +23,6 @@ public class DataElementError {
         this.elementName = elementName;
         this.value = value;
         this.error = error;
-    }
-
-    public String toJson() {
-        return new Gson().toJson(this);
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/DataElementError.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.ingest.common.error;
+package org.opentestsystem.rdw.ingest.common.util;
```
malaffoon: Moved this out of `error` package since it was the only thing in there and should live with the collector. Yeah, i could've moved the collector into the `error` package instead ... but, well, i didn't.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/DataElementErrorCollector.java
```diff
@@ -25,11 +23,18 @@ public void reset() {
...
     public String toJson() {
-
         return new Gson().toJson(this);
     }
 
     public boolean isEmpty() {
         return messages.isEmpty();
     }
+
+    public int size() {
```
malaffoon: These two methods were added mostly for tests but they are consistent for a Collection-type class.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/AnyExamProcessor.java
```diff
@@ -1,9 +1,9 @@
-package org.opentestsystem.rdw.ingest.processor.service;
+package org.opentestsystem.rdw.ingest.processor.service.impl;
```
malaffoon: moved this base class to `impl` package

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/utils/ParseUtil.java
```diff
@@ -18,17 +18,16 @@
 /**
  * Helper methods for parsing
  */
-public abstract class ParseUtil {
+public final class ParseUtil {
```
malaffoon: switch to our favored technique for util classes

PR #203, opened 2017-05-31
Feature/migrate warehouse entity

chdietz:
Added step for organizations school and district.
Updated ReportingMigrateJobIT
- edited and created sql test scripts

todo: 
- add an IT for the actual step
- next entity

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseToStagingStepsConfig.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseToStagingStepsConfig.java
@@ -12,6 +12,8 @@ import java.util.List;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.ingest.common.model.ImportContent.CODES;
+import static org.opentestsystem.rdw.ingest.common.model.ImportContent.EXAM;
+import static org.opentestsystem.rdw.ingest.common.model.ImportContent.ORGANIZATION;
 
 /**
  * Configuration for batch steps responsible for moving the data from staging to reporting
@@ -23,6 +25,8 @@ public class WarehouseToStagingStepsConfig {
     private final MigrateWarehouseToStagingSqlConfiguration warehouseToStagingSqlConfiguration;
 
     public static final String StageCodesStepName = "stageCodesStep";
+    public static final String StageOrganizationsStepName = "stageOrganizationsStep";
+
 
     @Autowired
     public WarehouseToStagingStepsConfig(final StepBuilderFactory stepBuilderFactory,
@@ -58,6 +62,20 @@ public class WarehouseToStagingStepsConfig {
                         sqlsBuilder.build())
                 ).build();
     }
+
+    @Bean
+    public Step stageOrganizationsStep() {
+        final SqlCopyBuilder sqlsBuilder = new SqlCopyBuilder(warehouseToStagingSqlConfiguration.getEntities());
+
+        sqlsBuilder.addNext("district").addNext("school");
+
+        return stepBuilderFactory.get(StageOrganizationsStepName)
+                .tasklet(new SqlCopyExecutionStep(newArrayList(EXAM, ORGANIZATION),
+                        repository,
+                        sqlsBuilder.build())
+                ).build();
+    }
+
 
 
 }
--- a/migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
+++ b/migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
@@ -39,6 +39,8 @@ sql:
       - TRUNCATE TABLE staging.staging_item_difficulty_cuts
       - TRUNCATE TABLE staging.staging_language
       - TRUNCATE TABLE staging.staging_accommodation_translation
+      - TRUNCATE TABLE staging.staging_school
+      - TRUNCATE TABLE staging.staging_district
 
     findDistinctImportContentInRange: >-
       SELECT DISTINCT wi.content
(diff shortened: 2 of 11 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseToStagingStepsConfig.java
```diff
@@ -59,5 +63,19 @@ public Step stageCodesStep() {
...
+        sqlsBuilder.addNext("district").addNext("school");
+
+        return stepBuilderFactory.get(StageOrganizationsStepName)
+                .tasklet(new SqlCopyExecutionStep(newArrayList(EXAM, ORGANIZATION),
+                        repository,
+                        sqlsBuilder.build())
+                ).build();
+    }
+
+
```
chdietz: will remove extra white space

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -39,6 +39,8 @@ sql:
       - TRUNCATE TABLE staging.staging_item_difficulty_cuts
       - TRUNCATE TABLE staging.staging_language
       - TRUNCATE TABLE staging.staging_accommodation_translation
+      - TRUNCATE TABLE staging.staging_school
```
chdietz: adding truncate tables as they are migrated

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -200,6 +202,16 @@ sql:
...
                 (:accommodation_id, :language_id, :label)
 
+        ##### Entities migrated by import id range ##################################################
+        # Entities are 'Root' if their table has import_id and update_import_id
+        # When selecting a range of entities only import.id's with a status = 1 are included
+        # 'root-entity'.update_import_id always contains the most recent import.id
+        # staging and reporting 'root-entities` do not have an update_import_id
+        # staging and reporting 'root-entities` have import_id set to the warehouse update_import_id
+        #############################################################################################
+
```
chdietz: does this sound correct?

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -213,10 +225,32 @@ sql:
...
+        # ------------ school  -------------------------------------------------------------------
+        school:
+          sql:
+            warehouseRead: >-
+              SELECT
+                ws.id,
+                ws.natural_id,
+                ws.name,
+                ws.update_import_id AS import_id,
```
chdietz: needed because the generic from warehouseRead: - insert to stagingInsert: logic in JdbcSqlCopyWarehouseToStagingRepository

migrate-reporting/src/test/resources/WarehouseEntitiesTearDown.sql
```diff
@@ -0,0 +1,71 @@
+--
```
chdietz: thinking about refactoring the names of the test supporting sql files when complete

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
malaffoon: Sounds about right.

@agorina, @chdietz how are replayed imports being caught by migrate? Scenario: a bunch of imports either "get lost" or are rejected based on missing assessments or organization. So their status is not 1 (might be 0, might be negative). Time goes on. Lots of other imports succeed and are migrated. Then we replay those imports and their status changes to 1. And ... magic happens??

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
malaffoon: Can't you just skip the prefix on the fields?

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: This is how it is set up now. We may change it once we work on the provenance. Not sure. Because of this I do not think this comment should stay in here. It would be hard to remember to update it later if things change.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
malaffoon: Perhaps a comment in the file; if this happens a lot, perhaps just a comment at the top explaining it?

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: @malaffoon, no the replay will not be picked up the way it is done now.
We either need to change the import id with the replay or change the logic how we determine what ids to migrate.  Probably deserve a chat.

agorina:
Since you do not have a step test, it is hard to see that it works. I am assuming that you test with the job IT?

chdietz:
@agorina Yes, I tested it with ReportingMigrateJobIT, have a todo for the step IT.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: good idea, i think this will be common.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: Interesting comments and that you, @agorina @malaffoon , thus the reason for this comment :)
Suggestions on where to put this?
Maybe put this in a Jira or Confluence?

PR #78, opened 2017-03-23
fix unit tests to deal with existing data

malaffoon:

PR #207, opened 2017-06-01
More unit test and bug fixes.

agorina:
I hope this is the last bug I introduced in this area, at least for the next week or so.


package-processor/src/main/resources/application.sql.yml
```diff
@@ -58,12 +58,11 @@ sql:
 
     commonCoreStandardCreate: >-
       INSERT INTO item_common_core_standard(item_id, common_core_standard_id)
-          SELECT (SELECT id from item where natural_id = :item_natural_id), :common_core_standard_id
+          SELECT (SELECT i.id from item i join asmt a on i.asmt_id = a.id where i.natural_id = :item_natural_id and a.natural_id=:asmtId) , :common_core_standard_id
```
agorina: An Item natural_id is not unique, a combination of an asmt and item should be unique, I think.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessorIT.java
```diff
@@ -0,0 +1,45 @@
...
+    public void itShouldTest() throws IOException {
+
+        final int beforeAssessmentCount = countRowsInTable(jdbcTemplate, "asmt") + countRowsInTable(jdbcTemplate, "asmt");
+        final int beforeItemCount = countRowsInTable(jdbcTemplate, "item") + countRowsInTable(jdbcTemplate, "item");
+
+        packageProcessor.process(ByteStreams.toByteArray(this.getClass().getResourceAsStream("/IAB_ICA_Combined.items.csv")), -99);
+        assertThat(countRowsInTable(jdbcTemplate, "asmt")).isEqualTo(beforeAssessmentCount + 9);
+        assertThat(countRowsInTable(jdbcTemplate, "item")).isEqualTo(beforeItemCount + 81);
+    }
+}
```
agorina: With the change in https://github.com/SmarterApp/RDW_Schema/pull/78, this test now passes.

PR #209, opened 2017-06-01
add status end-point to package-processor

malaffoon:
fix (broken?) test file

### DIFF
--- a/common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
+++ b/common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
@@ -1,11 +1,5 @@
 package org.opentestsystem.rdw.ingest.common.util;
 
-<<<<<<< HEAD
-/**
- * Created by marklaffoon on 6/1/17.
- */
-public class ParserHelperTest {
-=======
 import com.google.common.base.Function;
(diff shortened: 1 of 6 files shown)


common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
```diff
@@ -1,11 +1,5 @@
 package org.opentestsystem.rdw.ingest.common.util;
 
-<<<<<<< HEAD
```
malaffoon: i'm not sure what happened here ... trying to fix

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/status/DatabaseStatusIndicator.java
```diff
@@ -0,0 +1,79 @@
...
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * A {@link StatusIndicator} for database status.
+ */
+@Component
+public class DatabaseStatusIndicator extends AbstractStatusIndicator {
+
+    @Value("${sql.status.readTest}")
```
malaffoon: slightly different approach on this one ... instead of using a XxxRepository, i'm just hitting the database directly

PR #214, opened 2017-06-02
use AssessmentType from rdw common

malaffoon:
Most of the changes are actually moving TRT to rdw.common.model.trt


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -148,7 +148,7 @@ Assessment safeBuild(final DataElementErrorCollector elementErrorCollector) {
 
         // could put this in ParserHelper but it is pretty specialized
         private Double optionalForIAB(final String name, final String rawValue) {
-            if (IAB.sameAs(asmtType) && Strings.emptyToNull(rawValue) == null) return null;
+            if (asmtTypeId != null && IAB.id() == asmtTypeId && Strings.emptyToNull(rawValue) == null) return null;
```
malaffoon: I chose to not implement this in the enum since this is the only place it is used in this special way (where argument may be null)

agorina: OK

PR #217, opened 2017-06-04
Fix for DWR-392 Migration failure with student groups and sample data

agorina:
This PR fixes the above mentioned issues, but while testing it I found more problems.

Migrating at the same time as the Import is running fails. After researching it a little bit further I know the following facts:
1. When migrate started it found only one record to migrate with **import id 1** and content CODE. **We preload this record when the schema is created**.
2. It was immediately migrated and then the migrate process went to sleep for 3 minutes while import was running.
4. After the migrate woke up, it ran a few batches of import ids and failed between import id 4000-5000.  
5. The failure was while moving exams due to missing assessment in the reporting DB.
6. I found that the reporting was missing assessments with ids 1-4 (at least).  

What is interesting and puzzling is that the warehouse had **4** assessment records with **import id 1** ??!!  I do not have an explanation on how this could have happened and planning on looking into it further.  I suspect that some of it may be related to the message queue (4 records), so most likely will involve Mark into this. 

While analyzing the above failure I have come to the conclusion that we have the following issues (outside of the mentioned above): 
1.	Stubbed out assessment and student group create methods are not conscious of the concurrency. If two threads/processes will try to create the same assessment (or group), they may collide. We have not planned for the stubbed out version to live that long but since they are still around we need to do something about it.
2.	Identifying migrate records by import id has two problems:
a.	We made a conscious design decision to not wrap the entire TRT into one transactions when writing into the warehouse. Because of that an import status may be failed while some traces of the TRT could had been successfully created: for example, a Student is created but not an exam. Since we are migrating by the status = 1, we will ignore the student in the above example. It is problematic.
b.	Import Id indicates a start of the import into the warehouse, but not the end. We want to pull records in the ‘end order of import’ not the ‘start order’. For example, if two threads run two imports in parallel, let’s say the first one creates an import id 100, and the second one 101. They may finish in the different order, so the status will be updated on the 101 first, and then on the 100th. If the migrate request X records from the 99 that have been processed, depending on the timing it may miss the 100th record.

There are multiple ways to address the above issues. I will call a meeting to discuss it. Just wanted to share it and give people time to ponder.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StubbedOutAssessmentRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StubbedOutAssessmentRepository.java
@@ -12,7 +12,8 @@ public interface StubbedOutAssessmentRepository {
      * Inserts a new {@link Assessment}
      *
      * @param assessment the {@link Assessment} to be created
+     * @param importId the id that creates the assessment
      * @return the given {@link Assessment} with the id populated
      */
-    Assessment create(Assessment assessment);
+    Assessment create(Assessment assessment, long importId);
 }
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StudentGroupRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StudentGroupRepository.java
@@ -33,6 +33,7 @@ public interface StudentGroupRepository {
      *
      * @param studentId student id
      * @param groupIds group ids
+     * @param importId id of the import that adds the student to the groups
      */
-    void addStudentToGroups(long studentId, List<Integer> groupIds);
+    void addStudentToGroups(long studentId, List<Integer> groupIds, long importId);
 }
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
@@ -6,7 +6,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.BatchPreparedStatementSetter;
-import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
@@ -16,10 +15,11 @@ import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.sql.PreparedStatement;
-import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
 
 /**
  * JDBC implementation of {@link StudentGroupRepository}.
@@ -28,7 +28,6 @@ import java.util.List;
 class JdbcStudentGroupRepository implements StudentGroupRepository {
 
     private final NamedParameterJdbcTemplate jdbcTemplate;
-    private final StudentGroupRowMapper mapper = new StudentGroupRowMapper();
 
     @Value("${sql.studentGroup.create}")
     private String sqlCreate;
@@ -96,24 +95,11 @@ class JdbcStudentGroupRepository implements StudentGroupRepository {
             }
         });
 
+        final List<MapSqlParameterSource> batchParameters = newArrayList();
         for (final Integer groupId : groupIds) {
-            jdbcTemplate.update(sqlUpdateImportId, new MapSqlParameterSource("update_import_id", importId).addValue("id", groupId));
+            batchParameters.add(new MapSqlParameterSource("update_import_id", importId)
+                    .addValue("id", groupId));
         }
-    }
-
-    private static class StudentGroupRowMapper implements RowMapper<StudentGroup> {
-        @Override
-        public StudentGroup mapRow(final ResultSet rs, final int rowNum) throws SQLException {
-            return StudentGroup.builder()
-                    .id(rs.getInt("id"))
-                    .name(rs.getString("name"))
-                    .schoolId(rs.getInt("school_id"))
-                    .schoolYear(rs.getInt("school_year"))
-                    .subjectId(rs.getInt("subject_id"))
-                    .active(rs.getBoolean("active"))
-                    .creator(rs.getString("creator"))
-                    .created(rs.getTimestamp("created").toInstant())
-                    .build();
-        }
+        jdbcTemplate.batchUpdate(sqlUpdateImportId, batchParameters.toArray(new MapSqlParameterSource[batchParameters.size()]));
     }
 }
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
@@ -13,6 +13,7 @@ import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.jdbc.support.GeneratedKeyHolder;
 import org.springframework.jdbc.support.KeyHolder;
 import org.springframework.stereotype.Repository;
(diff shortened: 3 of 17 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StubbedOutAssessmentRepository.java
```diff
@@ -12,7 +12,8 @@
      * Inserts a new {@link Assessment}
      *
      * @param assessment the {@link Assessment} to be created
+     * @param importId the id that creates the assessment
      * @return the given {@link Assessment} with the id populated
      */
-    Assessment create(Assessment assessment);
+    Assessment create(Assessment assessment, long importId);
```
agorina: There was a shortcut in the sql.yml. It worked fine until the migrate came into life.  At the time it was created it meant to be a temporary solution so I have not planned for it to live that long.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StudentGroupRepository.java
```diff
@@ -33,6 +33,7 @@
      *
      * @param studentId student id
      * @param groupIds group ids
+     * @param importId id of the import that adds the student to the groups
      */
-    void addStudentToGroups(long studentId, List<Integer> groupIds);
+    void addStudentToGroups(long studentId, List<Integer> groupIds, long importId);
```
agorina: This was missing too.

exam-processor/src/main/resources/application.sql.yml
```diff
@@ -58,7 +58,7 @@ sql:
     # this is a stubbed out method - TODO: to be deleted
     create: >-
         INSERT INTO asmt (natural_id, name, label, grade_id, type_id, subject_id, school_year, import_id, update_import_id) VALUES
-                       (:natural_id, :name, :label, :grade_id, :type_id, :subject_id, :school_year, (select id from import limit 1), (select id from import limit 1))
```
agorina: This was a short cut.

migrate-reporting/src/main/resources/application.yml
```diff
@@ -10,7 +10,7 @@ migrate:
 
 spring:
   batch_datasource:
-    url: jdbc:mysql://localhost:3306/spring_batch_reporting?useSSL=false
+    url: jdbc:mysql://localhost:3306/spring_batch_reporting?useSSL=false&rewriteBatchedStatements
```
agorina: Oops, I will remove this. Was testing something, did not mean to include it.

migrate-reporting/src/test/resources/ImportContentSetup.sql
```diff
@@ -30,7 +30,8 @@ INSERT INTO warehouse_test.import (id, status, content, contentType, digest, bat
   (-89, 1, 5, 'application/xml', 'hash-student', 'batch', 'dwtest@example.com'),
   (-88, 1, 5, 'application/xml', 'hash-student', 'batch', 'dwtest@example.com'),
   (-87, 1, 5, 'application/xml', 'hash-student', 'batch', 'dwtest@example.com'),
-  (-86, 1, 5, 'application/xml', 'hash-student', 'batch', 'dwtest@example.com');
+  (-86, 1, 5, 'application/xml', 'hash-student', 'batch', 'dwtest@example.com'),
+  (2000, 1, 2, 'application/xml', 'hash-student-future-time', 'batch', 'dwtest@example.com');
```
agorina: I know we do not want to use a real positive ids. I needed it to be outside of the boundaries on the import batch size and doing it other way would require a lot tweaking of the test data. Since it inserts into ```warehouse_test``` I thought it may be ok. Please let me know if you strongly disagree.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -359,8 +359,9 @@ sql:
                 JOIN warehouse.student_group wsg ON wsg.id = wsgm.student_group_id
                 JOIN warehouse.student ws ON ws.id = wsgm.student_id
               WHERE
-                ( wsg.import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id >= :start_import_id AND id <= :end_import_id) OR
-                  wsg.update_import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id >= :start_import_id AND id <= :end_import_id) )
+                    ( wsg.import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id >= :start_import_id AND id <= :end_import_id) OR
+                      wsg.update_import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id >= :start_import_id AND id <= :end_import_id) )
+                AND (ws.import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id <= :end_import_id) OR  ws.update_import_id IN (SELECT id FROM warehouse.import WHERE status = 1 AND id <= :end_import_id) )
                 AND (wsg.deleted = 0 and wsg.active = 1)
```
agorina: This is a bug fix.

migrate-reporting/src/test/resources/WarehouseEntitiesTearDown.sql
```diff
@@ -36,7 +36,7 @@ DELETE FROM reporting_test.exam where id in (-88, -87, -86, -85, -84);
 DELETE FROM reporting_test.student_group_membership where student_group_id in ( -91, -8, -7);
 DELETE FROM reporting_test.user_student_group where student_group_id in ( -91, -8, -7);
 DELETE FROM reporting_test.student_ethnicity where student_id in ( -89, -88, -87, -86);
-DELETE FROM reporting_test.student where id in ( -11, -89, -88, -87, -86);
+DELETE FROM reporting_test.student where id in ( -11, -89, -88, -87, -86, -33);
 DELETE FROM reporting_test.student_group where id in ( -91, -8, -7);
```
agorina: There is no change to the tests, just the test data. Before the fix, just by adding the test data I was able to reproduce the reported failure.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
```diff
@@ -90,6 +95,10 @@ public int getBatchSize() {
                 return groupIds.size();
             }
         });
+
+        for (final Integer groupId : groupIds) {
```
malaffoon: Do this in one SQL call?

agorina: Good point. Will change.

migrate-reporting/src/test/resources/ImportContentSetup.sql
chdietz: I think that is ok.  I am working on refactoring these setup files now and was thinking of adjusting the numbers so there is more room, may adjust this one when I do that.

chdietz:
2. b. Import id and transactions.
Could we make use of a different import_status? currently 1 = processed?
Maybe we need started, processed and failed?

malaffoon:
I don't think the actual status value is the problem for 2b, the issue is the non-contiguous set of imports in the expected state. I think we just need to change the query to get just the range of imports that have no status gaps. In Alla's example, it should try to migrate 101 until 100 has moved to the correct state as well.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StubbedOutAssessmentRepository.java
chdietz: not sure i understand this do you mean `application.sql.yml`

migrate-reporting/src/test/resources/WarehouseEntitiesTearDown.sql
chdietz: so this change in test data ensures we are covering the issue you found correct?

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: I think this is ok as we discussed in the Jira, but we are essentially emulating having an import_id in membership table, do we think that would be better long term, for example when we implement replay.  And we are adding overhead to this query.  I have been going back and forth on this :)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StubbedOutAssessmentRepository.java
agorina: Yes, this is what I meant. Sorry for not being clear.

migrate-reporting/src/test/resources/WarehouseEntitiesTearDown.sql
agorina: That is correct.

PR #219, opened 2017-06-06
adding artifactory settings, command line override for rdw-project ve…

rcordeau:
…rsion dependencies

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/219)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,16 @@ buildscript {
     }
 }
 
+plugins {
+    id 'com.jfrog.artifactory' version '4.4.18'
+}
+
 apply plugin: 'io.spring.dependency-management'
+
+// Set version dependencies
+
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "21"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "100"
 
 allprojects {
     group = 'org.opentestsystem.rdw.ingest'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -14,8 +14,17 @@ buildscript {
...
 
+plugins {
+    id 'com.jfrog.artifactory' version '4.4.18'
+}
+
 apply plugin: 'io.spring.dependency-management'
 
+// Set version dependencies
+
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "21"
```
malaffoon: I don't understand this. Why can't it just assign rdwCommonVersion unconditionally? It's only used in here for dependency management, right?

rcordeau: Because when you specify -Pcommon on the command line, that is when the property is set. It is a pain in the ass, but you can't assign rdwCommonVersion = 21 in the build.gradle file and then override it on the command line - doesn't seem to work that way. https://stackoverflow.com/questions/25582173/override-property-in-build-gradle-from-the-command-line

rcordeau: I could have written it as:
//Default
String rdwCommonVersion = "21"

//Apply override if present
if(project.hasProperty("common")) {
    rdwCommonVersion = project.property("common")
}

malaffoon: Ah, thanks for explaining. Perhaps add a comment in the file like:
```
// Allow cli to force version, e.g. `-Pcommon=SNAPSHOT`
```

rcordeau: Ok, sounds good.

PR #227, opened 2017-06-08
bigerror

malaffoon:
When the parsing errors accumulate, the message gets too big to save. This fix just truncates the message. That's not great but if there are that many errors during parsing, i'm sure the first few hundred will get the point across ...


common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcUtils.java
```diff
@@ -1,9 +1,9 @@
-package org.opentestsystem.rdw.ingest.processor.repository.impl;
+package org.opentestsystem.rdw.ingest.common.repository;
```
malaffoon: already had this, just moved it to common

PR #228, opened 2017-06-08
Bug/loop

malaffoon:
The job runner is set up to loop as long as there is work to do. This is great except it leads to the migrator constantly pulling off sub-batches of work when ingest is processing things. For example, suppose ingest started up; before migrate kicks off it ingests 2500 records and is still cranking. So migrate processes 1000, then another 1000, then 650, then 120, then 10, then 10, then 10, then 10, etc. 

This change just stops that looping if there isn't a full batch of work to do. So in the example, it would process 1000, then another 1000, then it would wait until the next scheduled time. And probably do 1000 then.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -14,6 +14,7 @@
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
+import javax.swing.text.html.Option;
```
malaffoon: hmm, how'd that get in here?

PR #230, opened 2017-06-09
build file cleanup and docker credentials

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/230)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,15 +1,3 @@
-/*
- * Copyright 2017 Regents of the University of California. Licensed under the Educational Community License, Version
- * 2.0 (the "license"); you may not use this file except in compliance with the License. You may obtain a copy of the
- * license at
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -1,3 +1,17 @@
+/*
```
malaffoon: @jtreuting did we ever finalize the requirement of putting the copyright in every file or only in the LICENSE file in the project?

rcordeau: IAT is using copyright at least in the build.gradle.

malaffoon: And TDS has them in every source file. 

Doesn't mean that's the necessary and/or correct thing to do.

rcordeau: Removed copyright statement.

PR #231, opened 2017-06-09
completed_at must be set

malaffoon:
FYI.  If you want to write an IT that proves the fix, that would be great. If not, let me know and i'll do it. I just needed to get a fix in to get awsdev running again.


agorina:
I am confused on how it passed the IT before? It is not NULL in the reporting schema.

PR #237, opened 2017-06-14
Renamed asmt table claim columns to be consistent with the exam table.

agorina:
Renamed asmt table claim columns to be consistent with the exam table.


malaffoon:
commit summary is wrong

PR #239, opened 2017-06-15
Fix the issues with mySQL timestamp defaulting the current timestamp

agorina:
This will affect everybody, so PLEASE read.
MySQL 5.6 has this very strange default setting that causes the NULL timestamp to be automatically replaced with the current timestamp: https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp

Aurora does not have this problem.

In order to have the right configuration you will need to configure your local mySQL.
@rcordeau , I will need your help with configuring CI too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,32 @@ mysql> GRANT ALL PRIVILEGES ON *.* TO 'root'@'%';
 mysql> exit
 ```
 
+Additionaly, you need to configure MySQL setting in my.cnf file. Locate the file and add the following lines:
+```
+[mysqld]
+explicit_defaults_for_timestamp=1
+```
+
+If you do not know the location of my.cnf file, one way to check it is to run:
+```bash
+sudo fs_usage | grep my.cnf
+```
+This will report any filesystem activity in real-time related to that file.
+In another Terminal, restart your MySQL:
+```bash
+brew services restart mysql@5.6
+```
+On terminal with fs_usage, the proper location should be shown, e.g.
+```
+15:52:22  open            /usr/local/var/mysql/my.cnf
+```
+
+To verify that MySQL explicit_defaults_for_timestamp was set to 1:
+```bash
+mysql 
+mysql> SELECT @@explicit_defaults_for_timestamp;
+```
+
 The applications depend on the database being configured properly. See instructions below under [Running](#running)
 
 
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
@@ -12,6 +12,7 @@ import org.opentestsystem.rdw.ingest.processor.model.StudentExamAttributes;
 import org.opentestsystem.rdw.ingest.processor.repository.ExamRepository;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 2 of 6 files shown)


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -243,4 +244,16 @@ public void itShouldCreateExams() {
         assertThat(countRowsInTable(jdbcTemplate, "exam_item")).isEqualTo(beforeExamItemCount + 4);
         assertThat(countRowsInTable(jdbcTemplate, "exam_available_accommodation")).isEqualTo(beforeExamAccommodationCount + 2);
     }
+
+    /**
+     * The exam's completed date is defined as not null in the schema.
+     * If `explicit_defaults_for_timestamp` mySQL setting is turned off, this causes an undesirable behaviour.
+     * This test verifies that this setting is turned on.
```
chdietz: This test is dependent on the setting documented in the read-me?
explicit_defaults_for_timestamp=1

1. not saying we do not need the IT.  But do we need a start-up settings check?  Maybe one or more components in ingest should check this and if not correct stop work?

README.md
```diff
@@ -42,6 +42,32 @@ mysql> GRANT ALL PRIVILEGES ON *.* TO 'root'@'%';
...
+On terminal with fs_usage, the proper location should be shown, e.g.
+```
+15:52:22  open            /usr/local/var/mysql/my.cnf
+```
+
+To verify that MySQL explicit_defaults_for_timestamp was set to 1:
+```bash
+mysql 
+mysql> SELECT @@explicit_defaults_for_timestamp;
+```
```
malaffoon: Did you do a dump of the variables on Aurora vs. your MySQL instance? Were any others different?

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
malaffoon: This is an IT to check that *developers* instances match Aurora. The idea is that all developers need to be "forced" to have matching settings so they don't waste any time/effort with incompatible settings.

chdietz: ok, so is tis the default on Aurora?

malaffoon: correct

chdietz: we could investigate a docker version of mysql with compliant settings to ease development, that way every dev runs the same MySql when they want to run their own.

README.md
agorina: Now, since you asked, I did :)

Yes, there are multiple differences. Most are either performance or aws/aurora related, but there are some that may be worth looking into. For Ex.:

mysql:character_set_server	utf8
aurora: character_set_server	latin1

mysql: collation_server	utf8_general_ci
aurora:collation_server	latin1_swedish_ci

malaffoon: Oh jeez, that character set on aurora scares the crap out of me ... we need to fix that.

agorina: I am not sure what it means though, there are three values:
character_set_results	utf8
character_set_server	latin1
character_set_system	utf8

I will create a JIRA to research the diff and see if there is anything concerning.

rcordeau:
Ok, the CI servers have their mysql environments updated. I will update the flow on the CI server side to include running integration tests against Aurora as well.

PR #244, opened 2017-06-20
Fix for DWR-456: Un-scored TRT should not be migrated

agorina:
This fixes the issue for the newly loaded data.
I am not sure what to do with the already loaded data. 
Options I see:
1. write a SQL script to clean them up in the reporting. 
2. write a SQL script to mark them as 'deleted' in the warehouse and expect migrate to clean them up. This way we could also test the migrate/delete. The risk is that if it fails - we will have more issues to deal with.
3. do nothing. I think the 'data generator' data do not have unscored IABs( @malaffoon, please correct me if needed). ETS loaded data do have this problem. If this is the case, would it be easier to load ETS data?

Please let me know what you think.

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlCopyExecutionStep.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlCopyExecutionStep.java
@@ -1,18 +1,13 @@
 package org.opentestsystem.rdw.ingest.migrate.reporting.step;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.opentestsystem.rdw.ingest.common.model.ImportContent;
-import org.opentestsystem.rdw.ingest.migrate.reporting.Migrate;
 import org.opentestsystem.rdw.ingest.migrate.reporting.MigrateBatch;
-import org.opentestsystem.rdw.ingest.migrate.reporting.MigrateStatus;
 import org.opentestsystem.rdw.ingest.migrate.reporting.repository.SqlCopyWarehouseToStagingRepository;
 import org.springframework.batch.core.StepContribution;
 import org.springframework.batch.core.scope.context.ChunkContext;
 import org.springframework.batch.repeat.RepeatStatus;
 
 import java.util.List;
-
-import static com.google.common.collect.Lists.newArrayList;
 
 /**
  * A generic step for executing {@link SqlCopyExecutionStep#sqlCopyStatements} in the given order if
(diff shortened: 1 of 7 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlCopyExecutionStep.java
```diff
@@ -41,28 +36,12 @@ public SqlCopyExecutionStep(final Iterable<ImportContent> supportedContents,
...
-
-    @VisibleForTesting
-    List<SqlCopyStatements> getSqlCopyStatements() {
-        return sqlCopyStatements;
-    }
-
-    @VisibleForTesting
-    Iterable<ImportContent> getSupportedContents() {
-        return newArrayList(supportedContents);
-    }
```
agorina: Unrelated, just cleaned it up.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseToStagingStepsConfig.java
```diff
@@ -139,7 +139,7 @@ public Step stageExamStep() {
                 .addNext("exam_claim_score_by_update_import_id");
 
         return stepBuilderFactory.get(StageExamStepName)
-                .tasklet(new SqlCopyExecutionStep(newArrayList(EXAM, PACKAGE),
+                .tasklet(new SqlCopyExecutionStep(newArrayList(EXAM),
                         repository,
```
agorina: Ditto.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/StageExamIT.java
```diff
@@ -48,31 +47,47 @@
...
+                                .build(),
+                        newArrayList(newArrayList(ImportContent.GROUPS,
+                                ImportContent.PACKAGE,
+                                ImportContent.ORGANIZATION,
+                                ImportContent.CODES))));
 
-        JobExecution jobExecution = launchStep(StageExamStepName);
+        final JobExecution jobExecution = launchStep(StageExamStepName);
         assertThat(jobExecution.getExitStatus()).isEqualTo(ExitStatus.COMPLETED);
-        //verifyTableCountsAfterTest(tableTestCounts);
```
agorina: Not sure why this was commented out.

chdietz: may have been me, when i was working on it

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -801,6 +803,7 @@ sql:
               FROM warehouse.exam we
                JOIN warehouse.import wi on wi.id = we.import_id
               WHERE wi.status = 1 AND wi.id >= :start_import_id AND wi.id <= :end_import_id
+                AND (we.type_id in (1,3) OR (we.type_id = 2 and we.scale_score is not null))
```
chdietz: so if IAB has null scale score we do not migrate to data mart?

agorina:
I am not sure what the other approach would be. We do not want to reject this data. 
And I do not think we want to report it as an error to ETS. 
This was actually done this way before we merged two exams table and I forgot to include this where condition.
I am open to suggestions.

agorina:
I am merging the clean up changes, but rolled back the change to not migrate IABs with null scores. We will discuss this feature with the client tomorrow.

PR #247, opened 2017-06-22
DWR-398 Set up tests to reproduce the errors.

agorina:
This does not solve anything. 
It sets up the test environment to reproduce the failures.

I was not able to use junit framework due to a difference in the transaction management.
This creates a console app that runs some Repos and Services in the multi-threaded set up.

I am able to reproduce two errors @malaffoon had seen while ingesting the data and two more:

1. "call school_upsert(?, ?, ?, ?, ?, ?)}]; Column 'district_id' cannot be null"
2. "UPDATE student_group SET update_import_id = ? WHERE id = ?]; Deadlock found when trying to get lock; try restarting transaction"
3. "INSERT IGNORE INTO student_group_membership (student_group_id, student_id) VALUES (?, ?)]; Deadlock found when trying to get lock;"
4. null student id returned for the student upsert.

Issues 1 and 4 are solved by the retry.
Issues 2 and 3 are not really solved by the retry - it depends on the number of the concurrent threads. At least one thread succeeds with the retry, but not all.

I am planning on trying to find a solution for these issues, but wanted to get the feedback on the testing set up.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSchoolRepository.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSchoolRepository.java
@@ -7,6 +7,7 @@ import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.simple.SimpleJdbcCall;
 import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
 
 @Repository
 class JdbcSchoolRepository implements SchoolRepository {
@@ -20,6 +21,7 @@ class JdbcSchoolRepository implements SchoolRepository {
 
     @Override
     @Cacheable(value = "school", key = "#school")
+    @Transactional
     public int upsert(final School school, final long importId) {
         return (int) new SimpleJdbcCall(jdbcTemplate)
                 .withProcedureName("school_upsert").execute(new MapSqlParameterSource()
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
@@ -1,0 +1,56 @@
+package org.opentestsystem.rdw.ingest.processor.multithreaded;
+
+import org.opentestsystem.rdw.ingest.processor.configuration.DataSourceConfiguration;
+import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.JdbcTemplateAutoConfiguration;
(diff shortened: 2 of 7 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSchoolRepository.java
```diff
@@ -20,6 +21,7 @@
 
     @Override
     @Cacheable(value = "school", key = "#school")
+    @Transactional
```
agorina: Without this I cannot make the repo fail. You can read more in the JIRA if you are interested.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
```diff
@@ -0,0 +1,56 @@
...
+        "org.opentestsystem.rdw.ingest.processor.repository",
+        "org.opentestsystem.rdw.ingest.processor.repository.impl",
+        "org.opentestsystem.rdw.ingest.processor.model",
+        "org.opentestsystem.rdw.ingest.processor.multithreaded"
+})
+@Import({YamlPropertiesConfigurator.class,
+        DataSourceAutoConfiguration.class,
+        JdbcTemplateAutoConfiguration.class,
+        DataSourceConfiguration.class})
+@Profile("test")
```
agorina: I want to use the test schema.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
```diff
@@ -0,0 +1,56 @@
...
+        final List<Integer> failed = newArrayList();
+        for (final String test : ctx.getBeanNamesForType(MultiThreadedTest.class)) {
+            logger.info("Running test for [" + test + "]");
+            final boolean passed = ctx.getBean(test, MultiThreadedTest.class).run();
+            logger.info(test + " " + (passed ? "passed" : "failed"));
+
+            failed.add(passed ? 0 : 1);
+        }
+
+        System.exit(failed.stream().mapToInt(Integer::intValue).sum());
```
agorina: @malaffoon had an idea of configuring the CI to run this and check for the exit code. 
I like this idea, but we need to resolve the issue first. Maybe CI people could help me with hooking it to the test run then.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedTest.java
```diff
@@ -0,0 +1,88 @@
...
+     * A multithreaded test runner
+     *
+     * @return a flag indicating a pass or fail
+     */
+    public boolean run() {
+        cleanUp();
+        setUp();
+
+        final ExecutorService pool = Executors.newWorkStealingPool(TEST_THREAD_COUNT);
+        final Runnable r = newRunnable(pool);
```
malaffoon: It doesn't make sense to me that the ExecutorService is being passed into the newRunnable method. I'll poke around a bit but i don't think this is the right pattern.

agorina: I agree, I meant to research it and ...forgot. Thank you. Let me know if you find something.

PR #249, opened 2017-06-24
DWR-401

malaffoon:
Some explanation and solicitation for feedback (sorry about the rambling) ...

The fundamental fix is to set `useLegacyDatetimeCode=false`. This works great, enabling the new version of timezone handling. With that flag it does what you'd think it would converting properly between the client and server time offsets. I will have to make similar changes in RDW_Reporting (https://github.com/SmarterApp/RDW_Reporting/pull/164), and all the sbac-config-repo files (https://gitlab.com/fairwaytech/sbac-config-repo/merge_requests/47).

A problem arises because, with this change, the mysql server *must* have a valid timezone set (see digression 1 below). For our Aurora instances, this isn't a problem: they are all UTC, awesome possum. For our local systems, we need to change them because PDT (the default for most of us) is an ambiguous timezone. We could load all the timezone info and set it to 'America/Los Angeles'; we could use '-07:00'; or we could use UTC -- the latter is my recommendation (see digression 2 below). We'll have to see how the CI MySQL's are set up.

So, whatcha all think about this? ~~@rcordeau i will need your help tweaking the CI MySQL instances if necessary and i'd like to do that *before* committing these changes.~~ I verified that mysql on the build agent is set to 'UTC'.

---

Digression 1: okay, before any timezone junkies jump all over me: yes, mysql is messed up and the documentation and verbiage is very misleading. Strictly speaking what mysql calls a timezone is really a zone offset. That is why PDT is ambiguous and UTC (which isn't actually a timezone) is okay. It basically loads a map of timezone -> zone offsets and tries to smooth over the inconsistencies.

---

Digression 2: setting to UTC makes things a little odd with a mysql client because you'll have to remember that the values you are setting are in UTC, not local timezone. And, because mysql doesn't accept timezone indicators in SQL you have to use convert_ts. So, to save a value that is 7:30am (PDT) you should do something like `SET created=convert_tz(timestamp('2017-06-24 07:30:00'), '-07:00', @@session.time_zone)`. Alternatively you could set your session.time_zone to 'PDT' in the client. OTOH it will make interacting with your local database more similar to interacting with the aurora instances. I think.

### DIFF
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
@@ -12,7 +12,6 @@ import org.opentestsystem.rdw.ingest.processor.model.StudentExamAttributes;
 import org.opentestsystem.rdw.ingest.processor.repository.ExamRepository;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
-import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.jdbc.Sql;
@@ -79,8 +78,7 @@ public class ExamRepositoryIT {
             "administration_condition_id = 1 and " +
             "asmt_version = '345' and " +
             "asmt_id = -1 and " +
-//            TODO: Mark will fix it
-//            "completed_at = convert_tz(timestamp('2007-01-02T14:30:00Z'), '+00:00', @@session.time_zone) and " +
+            "completed_at = convert_tz(timestamp('2007-01-02T14:30:00Z'), '+00:00', @@session.time_zone) and " +
             "completeness_id = 1 and " +
             "opportunity = 7 and " +
             "cast(scale_score as decimal(5,2)) =  77.7 and " +
--- a/README.md
+++ b/README.md
@@ -42,30 +42,35 @@ mysql> GRANT ALL PRIVILEGES ON *.* TO 'root'@'%';
 mysql> exit
 ```
 
-Additionaly, you need to configure MySQL setting in my.cnf file. Locate the file and add the following lines:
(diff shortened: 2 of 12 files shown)


wtritch:
Interesting.  Does the DB record the time zone of the insert, or are we just storing time stamps?  i.e. If I store 12am PDT does it pass back 12am PDT or 7am UTC?

malaffoon:
For TIMESTAMP fields, the value is stored in the database in UTC. The server converts to/from UTC to the server's timezone (which, for us is UTC, so i'll just ignore that part). The java driver knows that and converts the value to/from the time zone of the client. If you create an instant that is 12am in the current timezone (PDT), the java code will convert it from PDT to the server's timezone (UTC) and send that. When fetching the data, the reverse happens: the server returns 7am (UTC) and that is converted from the server's timezone (UTC) to the current timezone (PDT) and will look like 12am, same as was passed in.

wtritch:
Got it.  So if I'm in New York I'll see California test times in EDT rather than PDT. (Unless we add some smarts to the client.)

malaffoon:
To that last point ... yeah, this is all about getting the data into and out of the database and into the API server. How it is serialized and presented in the UI is still TBD. I'm not sure there are really any timestamps that the user will be interested in seeing so maybe not an issue.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -79,8 +78,7 @@
             "administration_condition_id = 1 and " +
             "asmt_version = '345' and " +
             "asmt_id = -1 and " +
-//            TODO: Mark will fix it
-//            "completed_at = convert_tz(timestamp('2007-01-02T14:30:00Z'), '+00:00', @@session.time_zone) and " +
+            "completed_at = convert_tz(timestamp('2007-01-02T14:30:00Z'), '+00:00', @@session.time_zone) and " +
```
chdietz: So I understand this works if I run an IT on my workstation in PDT if my @@session.time_zone is UTC.
I assume our CI server would do the same or just run in UTC

malaffoon: Yes. This should work no matter what @@session.time_zone is.

PR #253, opened 2017-07-04
fix dependency versions

malaffoon:


build.gradle
```diff
@@ -65,7 +65,15 @@ subprojects {
             dependency "org.opentestsystem.rdw.common:rdw-common-status:${rdwCommonVersion}"
             dependency "org.opentestsystem.rdw.common:rdw-common-utils:${rdwCommonVersion}"
 
+            // commonly used libraries
+            dependency 'com.google.code.gson:gson:2.8.0'
```
malaffoon: Similar to recent changes in reporting, this moves the library *version* management up in this section ...

build.gradle
```diff
@@ -74,10 +82,11 @@ subprojects {
     }
 
     dependencies {
-        compile "org.slf4j:slf4j-api:1.7+"
+        compile 'com.google.guava:guava'
```
malaffoon: ... while this section is just for compiling in commonly used libraries

PR #256, opened 2017-07-05
DWR-480, capture oppId

malaffoon:
Read https://jira.fairwaytech.com/browse/DWR-480 for some details but: this is a field that is in the TRT, and Alla had already added the database field. It is considered optional, and there is no need to migrate it to the data mart.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -78,6 +78,7 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
                 .assessmentId(assessment.getId())
                 .asmtVersion(parserHelper.validate("Test, AssessmentVersion", test.getAssessmentVersion(), 30, false))
                 .opportunity(opportunity.getOpportunity())
+                .oppId(opportunity.getOppId())
```
agorina: Since it is  varchar(60) it needs to go through ```parserHelper.validate(...,60...)```.

agorina:
Just to note, this may conflict with this https://github.com/SmarterApp/RDW_Ingest/pull/255

PR #257, opened 2017-07-05
DWR-443 - more clean up and unit tests

agorina:
Still not done, but moving forward. More tests to add.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Student.java
```diff
@@ -96,7 +95,7 @@ public static Builder builder() {
 
         public Student build() {
             final Student student = new Student();
-            student.ssid = checkNotBlank(ssid);
```
agorina: The service does the validation, so this is not needed.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/StudentGroup.java
```diff
@@ -67,7 +63,6 @@ public static Builder builder() {
         private Instant created;
 
         public StudentGroup build() {
-            checkArgument(!StringUtils.isNullOrEmpty(name), "invalid name");
```
agorina: Moved to the service layer.

task-service/src/main/resources/application.sql.yml
```diff
@@ -13,6 +13,7 @@ sql:
         AND (:creator IS NULL OR i.creator = :creator)
         AND (:before IS NULL OR i.created < :before)
         AND (:after IS NULL OR i.created > :after)
+      ORDER BY i.id
       LIMIT 100000
```
agorina: Unrelated, but it was causing the IT to fail, randomly.

PR #259, opened 2017-07-06
DWR-443 - more unit tests

agorina:


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessorTest.java
```diff
@@ -0,0 +1,176 @@
...
+        final ExamineeProcessor examineeProcessor = mock(ExamineeProcessor.class);
+
+        examProcessor = new DefaultStudentExamProcessor(examItemProcessor, adminConditionRepository, completenessRepository, accommodationRepository, examineeProcessor);
+
+        when(examItemProcessor.parseExamItems(any(), any(), any())).thenReturn(examItems);
+        when(examineeProcessor.parseStudentExamAttributes(any(), any(Integer.class), any())).thenReturn(studentExamAttributes);
+        when(adminConditionRepository.findIdByCode("NS")).thenReturn(77);
+        when(completenessRepository.findIdByCode("Complete")).thenReturn(44);
+
+        tdsReport = new TDSReport();
```
malaffoon: Guess this works but i wonder if it would be easier to read the TDSReport from an XML file.

agorina: It is easier to simulate errors this way.

PR #260, opened 2017-07-06
DWR-443 - more unit tests

agorina:
I think this change completes this JIRA.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
@@ -7,8 +7,6 @@ import java.time.LocalDate;
 import java.time.Year;
 import java.util.function.Function;
 
-import static com.google.common.base.Strings.isNullOrEmpty;
-
 /**
  * Helper methods for parsing and validating data elements
  */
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -126,13 +118,7 @@ public String validate(final String name, final String rawValue, final int maxLe
     public static final Function<String, Integer> toIntegerOrNull = value ->
             StringUtils.hasText(value) ? Integer.parseInt(value.trim()) : null;
 
-    public static final Function<String, LocalDate> tryToParseToLocalDate = value -> {
-        if (isNullOrEmpty(value)) return null;
-        return LocalDate.parse(value);
-    };
+    public static final Function<String, LocalDate> toLocalDateOrNull = value ->
```
agorina: Renamed to be consistent with other names in this file.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -126,13 +118,7 @@ public String validate(final String name, final String rawValue, final int maxLe
...
+            StringUtils.hasText(value) ? LocalDate.parse(value.trim()) : null;
 
-    public static final Function<String, Boolean> tryToParseToBoolean = value -> {
```
agorina: This is the same as toBooleanOrNull.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/Precondition.java
```diff
@@ -15,7 +16,7 @@
      * @throws IllegalArgumentException if the given value is blank
      */
     public static String checkNotBlank(final String value) {
-        checkArgument(!isNullOrEmpty(value), "value may not be blank");
```
agorina: This was wrong.

malaffoon: Yeah, that surprised me too ... i just assumed isNullOrEmpty worked like the lang3 utility.

PR #261, opened 2017-07-06
coverage

malaffoon:
replace isNullOrEmpty with !hasText

Didn't find any bugs doing it, but improved coverage.


common/src/main/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQuery.java
```diff
@@ -141,10 +141,14 @@ public Builder content(final ImportContent content) {
         }
 
         public Builder content(final String content) {
-            if (isNullOrEmpty(content)) {
+            if (!hasText(content)) {
```
jtreuting: Does this now include checks for just whitespace?

malaffoon: hasText() checks that the value is not null, not empty, and has at least one non-whitespace character

PR #262, opened 2017-07-06
DWR-444 - Removed TODOs that seemed to be addressed

agorina:


migrate-reporting/src/test/resources/MigrateImportFromWarehouseSetup.sql
```diff
@@ -1,8 +1,3 @@
--- Delete imports loaded by initial dml.
--- todo: we should eventually remove this from
-DELETE FROM warehouse_test.import
-WHERE id = 1;
-
```
agorina: Since we are using a different range of import ids for all the tests, this is not needed.

migrate-reporting/src/test/resources/MigrateImportFromWarehouseTeardown.sql
```diff
@@ -1,7 +1,3 @@
--- todo: remove this from initial dml
-INSERT INTO warehouse_test.import(id, status, content, contentType, digest) VALUES
-  (1, 1, 3, 'initial load', 'initial load');
```
agorina: Ditto

migrate-reporting/src/test/resources/MigrateEntitiesPreloadReportingSetup.sql
```diff
@@ -1,6 +1,3 @@
---
---  todo: make sure import id's are aligned with ImportContentSetup.sql
```
agorina: This seems to be done.

chdietz: yes, that was a reminder to try and re use IT sql scripts

migrate-reporting/src/test/resources/MigrateImportFromWarehouseSetup.sql
chdietz: Yes, but will initial DML add this still?
The DML we use to populate the warehouse with codes, etc should not add an import correct?

migrate-reporting/src/test/resources/MigrateImportFromWarehouseTeardown.sql
chdietz: ditto my last comment

migrate-reporting/src/test/resources/MigrateImportFromWarehouseSetup.sql
agorina: I think the initial DML should have it. 
How else will we migrate all the data we are preloading into the warehouse?

PR #265, opened 2017-07-09
fixing bug where migration of translation data from warehouse could e…

rcordeau:
…rase any user-defined translation values for the frontend in the reporting database.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/265)

### DIFF
--- a/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
@@ -343,7 +343,7 @@ sql:
                       FROM staging.staging_translation st
                     WHERE
                       st.label_code = rt.label_code AND st.language_code = rt.language_code
-                      AND st.namespace = rt.namespace);
+                      AND st.namespace = rt.namespace) AND rt.namespace = (SELECT backend from reporting.namespace_constants);
 
         # ############################## Exams ####################################################################
         # ------------ exam_available_accommodation -------------------------------------------------------------------
--- a/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
@@ -336,6 +336,10 @@ sql:
                        rt.namespace = st.namespace
                 WHERE rt.label_code IS NULL;
 
+            # --- With this delete, we don't want to delete anything that isn't part of the "backend" namespace
+            # --- The "backend" namespace is hardcoded in the staging table for anything that is in the warehouse for translation
+            # --- Insert and update don't need the extra check because they are already checking against the staging table
+            # --- value, which is already set to "backend" and they will only update rows with that namespace.
             delete: >-
                DELETE rt FROM reporting.translation rt
                   WHERE NOT EXISTS(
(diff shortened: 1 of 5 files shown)


migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -343,7 +343,7 @@ sql:
                       FROM staging.staging_translation st
                     WHERE
                       st.label_code = rt.label_code AND st.language_code = rt.language_code
-                      AND st.namespace = rt.namespace);
+                      AND st.namespace = rt.namespace) AND rt.namespace = (SELECT backend from reporting.namespace_constants);
```
rcordeau: I decided not to use a hardcoded string literal for the rt.namespace comparison (in case it changed), so I added a new namespace_constants table in the reporting DB to hold the names assigned to front end and back end namespaces.

agorina: Don't you need the same logic in the update and insert statements? I think they all need to be limited to the 'backend' namespaces,otherwise you could end up overriding other labels/translations.

rcordeau: Well, I thought about that, but what gets in to the warehouse should only be backend namespace items coming from the accessibility code translations. Do you think the warehouse could be storing frontend translations?

agorina: Since you want the namespace to be driven by the table in the reporting, then the update should be driven by that too. What if someone creates an entry with the same namespace in both warehouse and reporting? This will do a silent override.

malaffoon: I know i'll probably regret this but ... why not just hardcode the string? This approach is hardcoding the column name, so i don't really see the advantage, and it requires a subselect/join everywhere.

rcordeau: Yes, if someone creates a "backend" namespace entry in the reporting table, it will get silently overridden, as well as it will get deleted unless it exists in the warehouse too. From what I thought we had discussed, no one should really be allowed to manually enter a translation entry in the warehouse, only in the reporting to override a frontend namespace entry. And, I was under the impression only backend entries would be in the warehouse. Regardless, I'll change all of them to for update, insert, and delete on backend namespaces only.

rcordeau: Yeah, I was going to have it just hard coded, but then changed my thoughts on it. It would be easier to just hard code it.

rcordeau: Since we really only have 1 namespace for the backend, hard coding seems fine. With respect to the logic of updating only backend namespaces in the reporting db, I believe that logic is already there because the namespace column exists in the staging table and I have logic in the insert and update to match on the value of that column.

migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -343,7 +343,7 @@ sql:
                       FROM staging.staging_translation st
                     WHERE
                       st.label_code = rt.label_code AND st.language_code = rt.language_code
-                      AND st.namespace = rt.namespace);
+                      AND st.namespace = rt.namespace) AND rt.namespace = "backend";
```
agorina: Could you please add a comment to this section explaining that when the data gets put into staging db, this namespace is hardcoded there? It took me a few sec to understand that this is how it work.

rcordeau: Not a problem at all.

PR #267, opened 2017-07-10
Feature/dwr 475

rcordeau:
Adding the moIT task - migrate only integration test. New task doesn't clean the database before running integration tests. It will facilitate testing flyway's ability to migrate

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/267)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,11 +167,19 @@ subprojects {
         }
     }
 
-    //Specific integration test task that will automatially clean and migrate the test database schemas
+    //Specific integration test task that will automatically clean and migrate the test database schemas
     task IT(type: Test, dependsOn:[":cleanAllTest", ":migrateAllTest"]) {
         outputs.upToDateWhen { false }
         doFirst {
             println "Running Integration Tests..."
+        }
+    }
+
+    //Specific integration test task that will automatically migrate the test database schemas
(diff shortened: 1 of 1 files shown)


rcordeau:
If this is fine, then I'll update the CI server to run this task as well as the original one.

build.gradle
```diff
@@ -167,14 +167,22 @@ subprojects {
...
+    //Specific integration test task that will automatically clean and migrate the test database schemas
     task IT(type: Test, dependsOn:[":cleanAllTest", ":migrateAllTest"]) {
         outputs.upToDateWhen { false }
         doFirst {
             println "Running Integration Tests..."
         }
     }
 
+    //Specific integration test task that will automatically migrate the test database schemas
+    task moIT(type: Test, dependsOn:[":migrateAllTest"]) {
```
malaffoon: So, this is the task. We're going to have to set up a separate database to run against?

The alternative (and i admit to not having thought it through) would be running earlier versions of cleanAll then applying the current version of migrateAll ...

rcordeau: Yeah, the only way this really makes any sense is to either have another db server to run against (aurora?) or another schema (warehouse_motest, reporting_motest?)

PR #269, opened 2017-07-12
adding in gradle task to run the multithreaded test app

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/269)


exam-processor/build.gradle
```diff
@@ -28,3 +28,13 @@ dependencies {
     compile project(':rdw-ingest-common')
     testCompile project(path: ':rdw-ingest-common', configuration: 'tests')
 }
+
+//This will allow for running of the multithreaded test application
+task runMTTest(type: JavaExec) {
```
rcordeau: automagic task that knows how to run a java app. If the multithreaded test app exits with a code different than 0, gradle will fail and show the exit code. This will be run as a separate task in the CI environment and is not automatically run on a "gradle test" or "gradle it" run.

malaffoon: Looks fine as is but ... perhaps we should set it up so it is running a set of classes.

rcordeau: Sure, no problem. Ticket was "Ingest: Create CI task to run the multithread test app". Should we create a new ticket to turn the app into a suite of test classes?

PR #275, opened 2017-07-18
DWR-602, don't waste time deleting non existent entries

malaffoon:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
```diff
@@ -37,9 +37,9 @@
     private String updateStudentImportIdSql;
 
     @Autowired
-    JdbcStudentRepository(final NamedParameterJdbcTemplate namedParameterJdbcTemplate, final JdbcTemplate jdbcTemplate) {
```
malaffoon: unrelated; opinion?

build.gradle
```diff
@@ -25,7 +25,7 @@ apply plugin: 'io.spring.dependency-management'
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=0.0.1-SNAPSHOT`
 
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "0.0.1-25"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-107"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-109"
```
malaffoon: strictly speaking, this isn't required but it will encourage the use of the more recent schema which has the unique index change

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
agorina: :+1:

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
```diff
@@ -67,10 +67,13 @@ public int upsert(final Student student, final long importId) {
...
+        // if the existing ethnicity ids match the new ids, we have nothing to do here ...
+        final List<Integer> existingIds = namedParameterJdbcTemplate.queryForList(findlAllStudentEthnicityIdsSql, new MapSqlParameterSource("student_id", studentId), Integer.class);
+        if (newIds.size() == existingIds.size() && newIds.containsAll(existingIds)) return;
 
-        if (!ethnicityIdsChanged(newIds, studentId)) return;
-
-        namedParameterJdbcTemplate.update(deleteStudentEthnicitySql, new MapSqlParameterSource("student_id", studentId));
+        if (!existingIds.isEmpty()) {
+            namedParameterJdbcTemplate.update(deleteStudentEthnicitySql, new MapSqlParameterSource("student_id", studentId));
+        }
```
agorina: Should there be a unit test for the change?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
jtreuting: why do we need both?  and couldn't the NamedParameterJdbcTemplate be injected and from that get the underlying jdbcTemplate?

agorina: CAn you get a jdbcTemplate from the NamedParameterJdbcTemplate? 

There are two types of calls in this class: stored procedure and just templated sql with named parameters.  To call a stored procedure, I could not find a way to do it with the NamedParameterJdbcTemplate (unless you know one).

malaffoon: The reason we need both is because the SimpleJdbcCall takes a JdbcTemplate instead of JdbcOperations. There is no way to get the JdbcTemplate from the NamedParameterJdbcTemplate. So this is the only way to have a single autowired component and get both things out of it.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
malaffoon: Yes there should. But there isn't.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentRepository.java
jtreuting: Looks like the answer is no you can't get it back out, so that answers that :) . I do like passing in 1 vs 2 though

PR #277, opened 2017-07-19
add import message on success

malaffoon:
@agorina made a comment yesterday about having useful status messages in the database for group loading. That made me think that it would be nice to preserve import messages for successes as well as failures. This is a start on that. 

Opinions?

### DIFF
--- a/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorConfiguration.java
+++ b/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorConfiguration.java
@@ -37,17 +37,17 @@ public class PackageProcessorConfiguration {
 
 
     @ServiceActivator(inputChannel = ImportPayloadSink.CODES_INPUT)
-    void processCodes(final Message<?> message) {
+    public void processCodes(final Message<?> message) {
         handleIncomingMessage(message, accommodationsProcessor);
     }
 
     @ServiceActivator(inputChannel = ImportPayloadSink.ORGANIZATION_INPUT)
-    void processOrganization(final Message<?> message) {
+    public void processOrganization(final Message<?> message) {
         handleIncomingMessage(message, organizationProcessor);
     }
 
     @ServiceActivator(inputChannel = ImportPayloadSink.PACKAGE_INPUT)
-    void processPackage(final Message<?> message) {
+    public void processPackage(final Message<?> message) {
         handleIncomingMessage(message, packageProcessor);
     }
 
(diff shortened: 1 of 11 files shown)


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorConfiguration.java
```diff
@@ -14,8 +15,6 @@
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.Message;
 
-import java.util.function.ObjLongConsumer;
```
malaffoon: Since we have an interface for ImportProcessors already, i'm using that instead of the consumer interface. Makes it easier to change the signature of the method (to return the success message).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAccommodationsProcessor.java
```diff
@@ -44,17 +44,17 @@
...
 
     private final AccommodationTranslationRepository translationRepository;
     private final LanguageRepository languageRepository;
 
-    private XPathExpression xpathAccSelections;
-    private XPathExpression xpathCode;
-    private XPathExpression xpathText;
-    private XPathExpression xpathLanguage;
-    private XPathExpression xpathLabel;
+    private final XPathExpression xpathAccSelections;
```
malaffoon: unrelated, fixing warnings

jtreuting:
This seems to make since to me.  Previously we would have been able to look at the message column and if it was null, know that it was a success.  I don't think that is a big deal at all and the information makes since to capture.  The only question I have is: Would having the number of successfully processed as a number instead of a string add any benefit?

malaffoon:
@jtreuting i thought about the number-vs-string thing; i was going to add a column to the import table and use it. But, for exams (which i haven't addressed yet) there are two types of things that get handled (exams and students). I guess i could have the column for the "primary" target of a processor, and then add whatever else in the message itself.

That also lead me down the path of numbers of things of each type of processing: i.e. 13500 schools parsed, 545 added as new schools, 12 updated. So, instead of dealing with it, i just took the quick and easy approach. Perhaps we could create a JIRA for more work?

jtreuting:
I think the quick approach is appropriate right now and we can create a JIRA and prioritize it.  We can get that data currently it's just not already summed up in a nice easy summary place, right?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAccommodationsProcessor.java
```diff
@@ -118,7 +119,9 @@ public void process(final byte[] accessibilityConfig) throws ImportException {
         //Insert into database
         translationRepository.create(accommodations);
 
-        logger.info(String.format("Processed %d accommodation codes", accommodations.size()));
+        final String msg = accommodations.size() + " accommodations processed";
+        logger.info(msg);
+        return msg;
```
chdietz: Additional info is good. So I agree with this

PR #278, opened 2017-07-19
DWR-584, store null strings as null

malaffoon:


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -42,7 +42,9 @@ public ParserHelper(@NotNull final DataElementErrorCollector elementErrorCollect
     }
 
     /**
-     * String value validation. It is a shortcut for {@link ParserHelper#checkNotBlank} and a string max length verification.
+     * String value validation.
+     * It is a shortcut for {@link ParserHelper#checkNotBlank} and a string max length verification.
+     * It effectively does trim-to-null on any blank or empty value.
```
malaffoon: Previously, an empty/blank string was returned as itself. Now it will be returned as null.

I inspected the use of this method and i'm pretty sure that is okay everywhere it is used. No tests failed and i was able to ingest everything i tried, and it all migrated.

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
```diff
@@ -112,41 +112,39 @@ public void itShouldValidateString() {
 
     @Test
     public void itShouldValidateEmptyString() {
-
-        assertThat(parserHelper.validate("field", "", 20, false)).isEqualTo("");
+        assertThat(parserHelper.validate("field", "", 20, false)).isNull();
```
malaffoon: But i am changing the behavior of the parser helper, so i'm a teeny bit nervous.

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
```diff
@@ -112,41 +112,39 @@ public void itShouldValidateString() {
...
-        assertThat(parserHelper.validate("field", null, 20, true)).isEqualTo(null);
+        assertThat(parserHelper.validate("field", null, 20, true)).isNull();
         assertThat(errorCollector.contains(new DataElementError("field", null, "value may not be blank"))).isTrue();
     }
 
     @Test
     public void itShouldParseYear() {
-        assertThat(toYear.apply(new Long(MIN_VALUE))).isEqualTo(MIN_VALUE);
-        assertThat(toYear.apply(new Long(MAX_VALUE))).isEqualTo(MAX_VALUE);
+        assertThat(toYear.apply((long) MIN_VALUE)).isEqualTo(MIN_VALUE);
```
malaffoon: unrelated, fixing warnings

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/ParserHelperTest.java
agorina: It is the right behaviour though.

PR #280, opened 2017-07-20
DWR-519 Perf: migrate performance

agorina:
Thanks @jtreuting!

With this change a single insert of 300 records dropped from 12 s. to 75.90 ms.

Here is more info: https://stackoverflow.com/questions/26307760/mysql-and-jdbc-with-rewritebatchedstatements-true

There are some concerns about using this flag ( I assume this is why it is not enabled by default), but I do not think it affects us. 

https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html#rewriteBatchedStatements 

I tested by migrating from AWS ETS to my local reporting, here are the stats:
select count(*) from exam; -- 420,571
select timestampdiff(SECOND,min(created), max(updated)) from reporting.migrate
-- 2495 seconds (41 min)  = 168 per second

NOTE: I did not update all the connections, just those that use ```batchUpdate``` call. Let me  know if you think I should blindly change all.

### DIFF
--- a/migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
+++ b/migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
@@ -860,7 +860,7 @@ sql:
                                               (:id, :exam_id, :item_id, :score, :score_status, :position, :response, :trait_evidence_elaboration_score,
                                                  :trait_evidence_elaboration_score_status, :trait_organization_purpose_score,
                                                  :trait_organization_purpose_score_status, :trait_conventions_score,
-                                                 :trait_conventions_score_status, migrate_id)
+                                                 :trait_conventions_score_status, :migrate_id)
 
         exam_item_by_update_import_id:
           sql:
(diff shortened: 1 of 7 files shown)


migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -860,7 +860,7 @@ sql:
                                               (:id, :exam_id, :item_id, :score, :score_status, :position, :response, :trait_evidence_elaboration_score,
                                                  :trait_evidence_elaboration_score_status, :trait_organization_purpose_score,
                                                  :trait_organization_purpose_score_status, :trait_conventions_score,
-                                                 :trait_conventions_score_status, migrate_id)
+                                                 :trait_conventions_score_status, :migrate_id)
```
agorina: Unrelated, noticed a typo.

agorina:
Yes, @jtreuting  - brilliant idea!

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
chdietz: how did this not fail?

agorina: I was wondering that too.

PR #281, opened 2017-07-21
DWR-569 Datamart: add exam.accommodation_codes column

agorina:
DWR-613 Ingest: denormalize dok
DWR-615 Ingest: math_practice is missing a code field


agorina:
I will merge it to keep reporting and ingest in synch, but will address any comment if needed.

PR #282, opened 2017-07-23
DWR-569 fix performance issues

agorina:
A previous approach made the migrate process 3 times slower. 

I guess it is a good coincidence that I was testing performance and had 'before the change' times captured. This bring it back to what it was before.

### DIFF
--- a/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
@@ -492,7 +492,7 @@ sql:
                   claim4.scale_score as claim4_scale_score,
                   claim4.scale_score_std_err as claim4_scale_score_std_err,
                   claim4.category as claim4_category,
-                  se.available_accommodation_codes
+                  code.codes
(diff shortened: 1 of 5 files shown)


migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -535,6 +535,12 @@ sql:
                   JOIN reporting.administration_condition rac ON rac.id = se.administration_condition_id
                   JOIN reporting.completeness rc ON rc.id = se.completeness_id
                   JOIN reporting.grade rg ON rg.id = ses.grade_id
```
agorina: Change to JOIN when moving from staging into reporting.

migrate-reporting/src/main/resources/application.yml
```diff
@@ -8,7 +8,7 @@ server:
 
 migrate:
   batch:
-    size: 1000
+    size: 2000
```
agorina: This works better. 

With this number I am seeing a rate of 273 exams per second when migrating from ETS to my local reporting.

PR #283, opened 2017-07-24
Feature/dwr 519 - multithread step that moved exams from warehouse to staging

agorina:
This is not final - just show the approach.

Would like to know what you think about the approach and if we want to implement it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ apply plugin: 'io.spring.dependency-management'
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=0.0.1-SNAPSHOT`
 
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "0.0.1-25"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-111"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-112"
 
 
 allprojects {
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobPartitioner.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobPartitioner.java
@@ -1,18 +1,16 @@
 package org.opentestsystem.rdw.ingest.migrate.reporting;
 
-import org.opentestsystem.rdw.ingest.migrate.reporting.repository.ReportingMigrateRepository;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.listener.JobExecutionListenerSupport;
 import org.springframework.batch.core.partition.support.Partitioner;
 import org.springframework.batch.item.ExecutionContext;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.util.Map;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.Math.abs;
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.opentestsystem.rdw.ingest.migrate.reporting.JobParams.FirstImportId;
 import static org.opentestsystem.rdw.ingest.migrate.reporting.JobParams.LastImportId;
@@ -23,18 +21,9 @@ import static org.opentestsystem.rdw.ingest.migrate.reporting.JobParams.LastImpo
  */
 @Component
 public class MigrateJobPartitioner extends JobExecutionListenerSupport implements Partitioner {
-
-    private static final Logger logger = LoggerFactory.getLogger(MigrateJobPartitioner.class);
-    private ReportingMigrateRepository reportingMigrateRepository;
-
     private static final String PARTITION_KEY = "partition";
-    private long firstImportId;
-    private long lastImportId;
-
-    @Autowired
-    void setReportingMigrateRepository(final ReportingMigrateRepository reportingMigrateRepository) {
-        this.reportingMigrateRepository = reportingMigrateRepository;
-    }
+    private Long firstImportId;
+    private Long lastImportId;
 
     @Override
     public void beforeJob(final JobExecution jobExecution) {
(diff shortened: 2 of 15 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/SqlCopyWarehouseToStagingRepository.java
```diff
@@ -1,12 +1,27 @@
...
+     * @param migrateId          the migrate id to add to the reporing write parameters
+     */
+    void execute(List<SqlCopyStatements> copyStatementsList, long firstImportId, long lastImportId, long migrateId);
+
+    /**
+     * Executes {@link SqlCopyStatements#getWarehouseRead()} and passes the result into {@link SqlCopyStatements#getStagingInsert()}}
+     *
+     * @param copyStatementsList the list of {@link SqlCopyStatements}
+     */
+    void execute(List<SqlCopyStatements> copyStatementsList);
```
agorina: There are two types of migrate SQL: with and without import ids. I am surprised it did not complain before, but I found it confusing that we are setting parameters that do not exists. Hence the change.

migrate-reporting/src/main/resources/application.yml
```diff
@@ -8,8 +8,9 @@ server:
 
 migrate:
   batch:
-    size: 2000
+    size: 4000
     delay: 60000
+    gridSize: 4
```
agorina: Here are some test stas while testing migrate from ETS to my local machine:

Without this change:
Batch  size = 2,000 -  273 per exams second

With this change:
Batch  size = 8,0000 , 4 grids - 331 exams per second
Batch size =  4,000, 4 grids -  372 exams per second

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobPartitioner.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public void beforeJob(final JobExecution jobExecution) {
+        final JobParameters jobParameters = jobExecution.getJobParameters();
+        firstImportId = jobParameters.getLong(FirstImportId);
+        lastImportId = jobParameters.getLong(LastImportId);
+    }
+
+    @Override
+    public Map<String, ExecutionContext> partition(final int gridSize) {
+        final long chunkSize = (lastImportId - firstImportId) / gridSize;
```
chdietz: My concern here as I mentioned in conversation is these are not sequential.  I guess the biggest concern is getting a proper division of labor.

malaffoon: Just cause it's how i am, i think this might be more readable/concise. I'm assuming first/last are inclusive. 
```java
    final long size = (long)Math.ceil((double)(lastImportId - firstImportId + 1) / gridSize);
    long first = firstImportId;
    for (int i = 0; i < gridSize; ++i, first += size) {
        final ExecutionContext ex = new ExecutionContext();
        ex.put(FirstImportId, first);
        ex.put(LastImportId, Math.min(first + size - 1, lastImportId));
        map.put(PARTITION_KEY + i, ex);
    }
```
You could make it even more compact by using a guava helper to create the map, replacing body of for loop with:
```
        map.put(PARTITION_KEY + i, new ExecutionContext(ImmutableMap.of(FirstImportId, first, LastImportId, Math.min(first + size - 1, lastImportId))));
```

migrate-reporting/src/main/resources/application.yml
malaffoon: Nice. Improving by 37% seems worthwhile, and it doesn't add too much complexity.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateJobPartitioner.java
agorina: Thank you :)

migrate-reporting/src/main/resources/application.yml
agorina: The rate is for the complete migration.

I will try the settings you are suggesting and post the rates here or in the final review.

build.gradle
```diff
@@ -25,7 +25,7 @@ apply plugin: 'io.spring.dependency-management'
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=0.0.1-SNAPSHOT`
 
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "0.0.1-25"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-111"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "0.0.1-112"
```
agorina: Unrelated - updated the SQL version.

migrate-reporting/src/test/resources/MigrateEntitiesFromStagingSetup.sql
```diff
@@ -89,10 +89,10 @@ INSERT INTO staging_test.staging_exam_student ( id, grade_id, student_id, school
...
 
 INSERT INTO  staging_test.staging_exam ( id, type_id,  exam_student_id, school_year, asmt_id, asmt_version, opportunity, completeness_id,
-        administration_condition_id, session_id, performance_level, scale_score, scale_score_std_err, available_accommodation_codes, completed_at, import_id, deleted, migrate_id) VALUES
-(-88, 1, -18, 2016, -99,  null, 1, -99, -99, 'session', 1, 2145, 0.17, 'code1-test|code2-test', '2016-08-14', -88, 0, -88),
-(-87, 1, -17, 2016, -11,  null, 1, -99, -99, 'session', 1, 2145, 0.17, 'code1-test', '2016-08-14', -88, 0, -88),
-(-86, 1, -16, 2016, -11,  null, 1, -99, -99, 'session', 1, 2145, 0.17, null, '2016-08-14', -88, 0, -88);
+        administration_condition_id, session_id, performance_level, scale_score, scale_score_std_err, completed_at, import_id, deleted, migrate_id) VALUES
+(-88, 1, -18, 2016, -99,  null, 1, -99, -99, 'session', 1, 2145, 0.17, '2016-08-14', -88, 0, -88),
+(-87, 1, -17, 2016, -11,  null, 1, -99, -99, 'session', 1, 2145, 0.17, '2016-08-14', -88, 0, -88),
+(-86, 1, -16, 2016, -11,  null, 1, -99, -99, 'session', 1, 2145, 0.17, '2016-08-14', -88, 0, -88);
```
agorina: Fixed to work with the latest scripts.

migrate-reporting/src/main/resources/application.yml
agorina: Here are more stats:

batch-size=2000,  1 grid  - 281 exams per second
batch-size=2000, 2 grids - 294
batch-size=2000, 4 grids - 253

batch-size= 4,000,  1 grid - 290
batch-size= 4,000, 2 grids - 350
batch-size= 4,000, 4 grids - 345

I know that the last number does not match the one reported above. There are a few variables:
1. My measurement are not very precise, I am counting the total number of exams and the overall migrate time from the first migrate entry to the last one.  And I am not waiting for the entire ETS to migrate, so I am cutting it in the middle of the process.
2. Network speed may be different between different tests. 

I am sure the proportion may change on a different machine and different network.

malaffoon: Hmm, now i'm waffling. We can discuss off-line.

PR #290, opened 2017-07-31
DWR-637 Change migrate to be based on the timestamp range

agorina:
There is a corresponding schema change, will have a PR shortly.

This addresses all of the items described here:
https://confluence.fairwaytech.com/display/SBACDW/DWR-617+Ingest%3A+exam+has+exam_student_id+that+was+created+AFTER+the+exam

The only outstanding issue is to clean up indexed that may not be needed anymore and tighten up NOT NULL constraints on the new columns.  Will do it next.


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -47,73 +40,66 @@ public JobExecutionListener listener() {
...
-                                   final Partitioner partitioner) {
+                                   final Step upsertExamsStep) {
         try {
 
             return jobBuilderFactory.get("Migrate Reporting Job")
                     .preventRestart()
                     .listener(listener())
-                    .listener((JobExecutionListener) partitioner)
                     .start(truncateStageStep)
-                    .next(warehouseStep)
```
agorina: I removed this step and moved the logic to be executed before the Job.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -129,18 +115,6 @@ public Job migrateReportingJob(final Step truncateStageStep,
     }
 
     @Bean
-    public Step truncateStageStep(final TruncateStage truncateStage) {
-        return stepBuilderFactory.get(stepTruncateStageName)
-                .tasklet(truncateStage).build();
-    }
```
agorina: I did not remove the step, just moved its creation into a different class.  It will be explained further down.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/WarehouseToStageRepository.java
```diff
@@ -18,16 +15,4 @@
     List<Map<String, Object>> findWarehouseSubjectClaimScores();
 
     List<Map<String, Object>> findReportingSubjectClaimScores();
-
-    void truncateStaging(Iterable<String> stagingTruncateStatements);
```
agorina: Removed as a duplicate of the SqlListExecutionRepository.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcReportingSqlListExecutionRepository.java
```diff
@@ -8,10 +8,10 @@
 import org.springframework.transaction.annotation.Transactional;
 
 /**
- * Jdbc implementation of {@link SqlListExecutionRepository}.
+ * Jdbc implementation of {@link SqlListExecutionRepository} on the reporting datamart
  */
 @Repository
-class JdbcSqlListExecutionRepository implements SqlListExecutionRepository {
+class JdbcReportingSqlListExecutionRepository implements SqlListExecutionRepository {
```
agorina: Renamed to explicitly state what datasource is being used here.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepository.java
```diff
@@ -31,41 +37,41 @@ public JdbcWarehouseImportRepository(final NamedParameterJdbcTemplate warehouseJ
...
-                return new MigrateImportValues(start, stop, null);
+                final Instant start = getInstantNullable(rs, "start");
+                final Instant stop = getInstantNullable(rs, "stop");
+                final int importCount = rs.getInt("cnt");
+
+                return MigrateImportValues.builder()
+                        .firstAt(start)
+                        .lastAt(stop)
+                        .importCount(importCount)
+                        .migrateCodes(importCount > 0 ? existsCodeImportContentInTimeRange(start, stop) : false)
```
agorina: Move it from the Warehouse step into here. 
After all the changes it did not feel like we need a separate step just for this.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlListCodesExecutionStep.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.batch.repeat.RepeatStatus;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * A generic step for executing {@link SqlListCodesExecutionStep#sqls} in the given order
+ */
+class SqlListCodesExecutionStep extends StepTasklet {
```
agorina: There is probably a better way of doing it: it is the same step as all other SqlListExecutionStep but it rung the queries conditionally based on the code content. Probably a better option is to have some kind of a decision fork in Spring Batch, but since I did not know a better way I am just doing it as a separate step.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseToStagingStepsConfig.java
```diff
@@ -165,8 +129,16 @@ public StepTasklet stageExamStep() {
...
-        return new SqlCopyExecutionStep(newArrayList(EXAM),
-                repository,
-                sqlsBuilder.build());
+        return stepBuilderFactory.get(StageExamStep)
+                .tasklet(new SqlCopyExecutionStep(repository, sqlsBuilder.build()))
+                .build();
+    }
+
+    @Bean
+    public Step truncateStageStep() {
```
agorina: This is where the tuncateStageStep is now defined.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/JobParams.java
```diff
@@ -4,30 +4,41 @@
...
-
-    /**
-     * Helper to create job parameters inline.
-     *
-     * @param firstImportId first import id
-     * @param lastImportId last import id
-     * @return job parameters with unique identifier and import ids set
-     */
-    public static JobParameters createJobParams(final long firstImportId, final long lastImportId) {
+final class JobParams {
```
malaffoon: I wonder if we could just have a wrapper that adds the access helpers to the parameters, similar to RdwMessageHeaderAccessor? e.g. 
```
public class MigrateJobParameters extends JobParameters {
    // constructor that takes same params as createJobParams
    public Instant getFirstAt() {
        return this.getDate(FirstAt).toInstant();
    }
...
```

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
```diff
@@ -47,73 +40,66 @@ public JobExecutionListener listener() {
...
     }
 
-
     /**
      * This job migrates data from the 'warehouse' to 'reporting' databases.
      * <p>
-     * The tables in the warehouse are divided into 'root' and 'children' tables based on their dependencies.
-     * All root tables have an import id that last updated a row as well as a delete flag to indicate soft deletes.
+     * The tables in the warehouse are divided into 'master' and 'children' tables based on their dependencies.
+     * All master tables have the crated/updated timestamps as well as a delete flag to indicate soft deletes.
```
malaffoon: typo

migrate-reporting/src/main/resources/application.sql.yml
```diff
@@ -21,11 +21,30 @@ sql:
...
+                    ORDER BY updated
+                    LIMIT :size) AS batch) s2
+          )
+
+      existsCodeImportContentInTimeRange:
+          SELECT 1 AS existsCodeImportContent
+            FROM warehouse.import wi
+          WHERE wi.content = 3
+             AND ((wi.status != 0 AND wi.updated >= :first_at AND wi.updated <= :last_at) OR
+               (wi.status = 0 AND wi.created <= :last_at))
```
malaffoon: Could this lead to codes being migrated before they are done processing?

migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -492,7 +492,8 @@ sql:
                   claim4.scale_score as claim4_scale_score,
                   claim4.scale_score_std_err as claim4_scale_score_std_err,
                   claim4.category as claim4_category,
-                  code.codes
+                  code.codes,
+                  se.updated
```
malaffoon: Shouldn't there be a corresponding change to the UPDATE query?

malaffoon:
As discussed, i'm still concerned about the entities associated with the last import record being migrated over and over (even if that migration is small and idempotent). I really don't think we should be triggering the job at all. Somehow.

The only thing that comes to mind is to adjust the last_at time based on empty query results. So, if a migrate iteration has no new results, adjust the last_at to be the db server's NOW() value. But yeah, we shouldn't be updating a completed migrate record, so maybe some placeholder migrate record?  Ugh. Let's discuss.

migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
agorina: Thank you, I missed it. Adding it.

migrate-reporting/src/main/resources/application.sql.yml
agorina: _Could this lead to codes being migrated before they are done processing, i.e. half written?_ Yes, I thought it is the whole point of it.

And yes to the second one. I am not sure how to avoid it thought? Besides detecticg stuck codes.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingConfiguration.java
agorina: Fixing, thank you.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -224,7 +224,7 @@ sql:
                 FROM warehouse.district wd
                 WHERE EXISTS(
                   SELECT ws.id from warehouse.school ws
-                    WHERE ws.updated >= :first_at AND ws.updated <= :last_at
```
agorina: Without this, it was 3 time slower. Probably after this change the indexes need to be adjusted once again, but I want to see how it runs on the awsdev before changing indexes.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/JobParams.java
agorina: Great idea! Changing.

PR #292, opened 2017-08-01
don't expect java 8 clock to support microseconds

malaffoon:
Every day is a learning opportunity ...

Apparently java 8 Instant supports nanosecond precision but the clock that generates instants does not. So, this change lets the database do all the timestamp creation while the java code merely passes the values around. Oh yeah, and Date doesn't even support microseconds so let's avoid that.


import-service/src/main/java/org/opentestsystem/rdw/ingest/repository/JdbcRdwImportRepository.java
```diff
@@ -81,7 +81,6 @@ public RdwImport create(final RdwImport rdwImport) {
                 .addValue("digest", rdwImport.getDigest())
                 .addValue("batch", rdwImport.getBatch())
                 .addValue("creator", rdwImport.getCreator())
-                .addValue("created", Timestamp.from(rdwImport.getCreated()))
```
malaffoon: Let the database use the default value of NOW

PR #293, opened 2017-08-01
make connection pool settings consistent

malaffoon:


import-service/src/main/resources/application.yml
```diff
@@ -23,6 +23,11 @@ security:
   oauth2:
     token-info-url: local
 
+server:
```
malaffoon: I pulled these numbers out of my ... uhmm ... by extensively researching on the interwebs. The default values are 1000 connections and 200 threads. Seems excessive. My experience is a single import service deployed to AWS handles about 30-40 simultaneous clients.

wtritch: But... those are so high...  If you're load-testing, I wonder if you'll get better performance with lower thread counts and connections?

malaffoon: baby steps; until we actually do the load-testing and fiddle with params, i didn't want to push it too far

PR #294, opened 2017-08-02
DWR-689 Merge all reporting schemas - Part 1

agorina:
This still uses two different schemas: reporting and warehouse. 
It also assumes that spring batch schema is part of the reporting (dependent on https://github.com/SmarterApp/RDW_Schema/pull/124). 

But now we do not need two different schemas for the spring batch.

Part 2 is to remove the schema names and use different datasources instead. Thanks to @wtritch for giving me a hint on how to do it in the ITs.


migrate-reporting/src/main/resources/application.sql.yml
```diff
@@ -36,7 +36,7 @@ sql:
                 count(*)     AS cnt
               FROM (SELECT updated
                     FROM warehouse.import
-                    WHERE status != 0 AND (:prev IS NULL OR updated >= :prev) AND updated < CURRENT_TIMESTAMP(6)
```
agorina: @malaffoon , I included this change. I tested it locally migrating from QA and it processed all and stopped.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseImportRepositoryIT.java
```diff
@@ -53,14 +53,14 @@ public void itShouldGetMigrateJobState() {
...
         assertThat(jobState.getLastAt()).isEqualTo(lastAt);
-        assertThat(jobState.getImportCount()).isEqualTo(3);
+        assertThat(jobState.getImportCount()).isEqualTo(1);
         assertThat(jobState.migrateCodes()).isTrue();
 
         jobState = repository.getMigrateImportValues(lastAt, 10);
         assertThat(jobState.getFirstAt()).isEqualTo(lastAt);
-        assertThat(jobState.getLastAt()).isEqualTo(lastAt);
-        assertThat(jobState.getImportCount()).isEqualTo(1);
-        assertThat(jobState.migrateCodes()).isTrue();
```
agorina: This is not related to the schema change. It is about stopping the migrate when all data have been processed.

PR #295, opened 2017-08-02
set client character encoding

malaffoon:
There are similar changes in reporting and in sbac-config-repo.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcItemTraitRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcItemTraitRepository.java
@@ -32,7 +32,7 @@ class JdbcItemTraitRepository implements ItemTraitRepository {
     private static class ItemTraitRowMapper implements RowMapper<ItemTrait> {
         @Override
         public ItemTrait mapRow(final ResultSet rs, final int rowNum) throws SQLException {
-            return new ItemTrait(rs.getInt("id"), rs.getNString("dimension"));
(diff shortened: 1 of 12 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcItemTraitRepository.java
```diff
@@ -32,7 +32,7 @@
     private static class ItemTraitRowMapper implements RowMapper<ItemTrait> {
         @Override
         public ItemTrait mapRow(final ResultSet rs, final int rowNum) throws SQLException {
-            return new ItemTrait(rs.getInt("id"), rs.getNString("dimension"));
+            return new ItemTrait(rs.getInt("id"), rs.getString("dimension"));
```
malaffoon: This was just wrong since the field is VARCHAR not NVARCHAR. It is interesting that it didn't fail until the client encoding was explicitly set.

import-service/src/test/java/org/opentestsystem/rdw/ingest/repository/DataSourceIT.java
```diff
@@ -28,4 +32,13 @@ public void itShouldUseTomcatPooling() {
         assertThat(tomcat.getMaxActive()).isEqualTo(10);
         assertThat(tomcat.getInitialSize()).isEqualTo(4);
     }
+
+    @Test
+    public void characterSetShouldBeUtf8() {
+        template.query("SELECT @@character_set_client, @@character_set_connection, @@collation_connection;", rs -> {
```
malaffoon: This was failing when i pointed at any of the aurora instances.

PR #300, opened 2017-08-08
reduce connections/threads for apps that only have management end-points

malaffoon:
Near as i can tell, for apps that have both a real endpoint and management endpoints, the same tomcat connection/thread management is shared. So i left import-service alone.

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -5,6 +5,9 @@ management:
     enabled: false
 server:
   port: 8008
+  tomcat:
+    max-connections: 2
+    max-threads: 2
 
 spring:
   cache:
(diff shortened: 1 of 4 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -30,6 +33,11 @@ spring:
     initialSize: 4
     maxWait: 10000
 
+  jackson:
```
malaffoon: unrelated, this was missing, needs to be in every app

exam-processor/src/main/resources/application.yml
```diff
@@ -5,6 +5,9 @@ management:
     enabled: false
 server:
   port: 8008
+  tomcat:
```
malaffoon: Running locally this did reduce the number of threads allocated by a few.

PR #301, opened 2017-08-09
Bug/status

malaffoon:


task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/UpdateOrganizationsConfiguration.java
```diff
@@ -8,6 +8,7 @@
 import org.opentestsystem.rdw.ingest.tasking.service.impl.DefaultUpdateOrganizationsService;
```
malaffoon: most of the changes are refactoring things so unit testing is possible

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/RestImportServiceClient.java
```diff
@@ -49,36 +37,15 @@ public void postOrganizationPayload(final byte[] payload) {
         final HttpEntity<byte[]> request = new HttpEntity<>(payload, headers);
         final RdwImport rdwImport = restTemplate.postForObject(properties.getOrganizationsImportsUrl(), request, RdwImport.class);
         if (rdwImport.getStatus().getValue() < 0) {
-            throw new RestClientException("organization import not accepted, id: " + rdwImport.getId() + ", status: " + rdwImport.getStatus());
+            logger.warn("organization import not accepted, id: {}, status: {}", rdwImport.getId(), rdwImport.getStatus());
```
malaffoon: This is the real change: don't throw just cause the system returns a non-positive response. That situation occurs when there is bad data in the payload. Sure, that's not good ... but it is not exceptional, it is a perfectly valid data flow.

PR #305, opened 2017-08-21
clean up; use exists instead of count

malaffoon:
when i was in there investigating an "unprocessed import" report, i noticed that we're counting imports when we really just need to check that at least one exists. And we know that counting things is a really bad idea.


task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/JdbcReportRepository.java
```diff
@@ -65,8 +64,8 @@ public void writeReconciliationReport(final RdwImportQuery query, final OutputSt
     }
 
     @Override
-    public long countExamImports(final RdwImportQuery query) {
-        return jdbcTemplate.queryForObject(sqlImportCount, mapQueryToSqlParameterSource(query), Long.class);
+    public boolean exists(final RdwImportQuery query) {
+        return jdbcTemplate.queryForObject(sqlImportExists, mapQueryToSqlParameterSource(query), Boolean.class);
```
malaffoon: Hmm, all my tests passed but does this work? Mapping a 0/1 response to a Boolean.

PR #306, opened 2017-08-21
ignore schools where parent is explicitly set to STATE

malaffoon:


malaffoon:
I want this for production deployment so i'm going to merge and then revisit any comments.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/JsonOrganizationParser.java
```diff
@@ -118,8 +118,11 @@ private void processNode(final JsonNode node) {
                         groupToDistrict.put(entityId, parentId);
                         break;
                     case "INSTITUTION":
-                        schools.put(entityId, School.builder().naturalId(entityId).name(entityName).build());
-                        schoolToParent.put(entityId, parentId);
+                        // completely ignore schools that are parented by STATE
+                        if (!"STATE".equalsIgnoreCase(nodeTextOrNull("parentEntityType", node))) {
```
chdietz: should we log something?  Because this is a non normal data state?

malaffoon: Oh. Yeah. That's a good idea.

PR #308, opened 2017-08-22
Add user-import to the group processor.

wtritch:
This is Step 1 of a two-step PR for group ingest processing.
Most of this SQL already existed in the hack-script, but has been broken into a sql.yml file with logical breakout into services.

In this step I've assumed that CSV data has been loaded into the upload_student_group table.
The general processing steps are:
1. standardize the uploaded values, converting empty strings into NULLs
2. bind existing students, schools, groups to their DB ids in the upload_student_group table
3. create and register an import for new student SSIDs called out in the CSV
4. un-delete and register an import for deleted student SSIDs called out in the CSV
5. cleanup staging/temp tables/references
6. abort any created imports if there is an issue

Step 2 is going to perform the actual group creation/update.


common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
```diff
@@ -0,0 +1,25 @@
...
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+import java.util.Map;
+
+/**
+ * This model contains SQL statements keyed by name.
+ */
+public class EntitySqls {
```
wtritch: I moved this and the associated SqlListBuilder utility and SqlListExecutionRepository interface into the common package for re-use.  This caused a lot of change ripple.  I'll call out the classes of interest.

common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/SqlListExecutionRepository.java
```diff
@@ -0,0 +1,25 @@
...
+     */
+    void execute(Iterable<String> sqls);
+
+    /**
+     * Execute the given sql statements in order, using the given parameters for each execution.
+     *
+     * @param sqls          SQL statements
+     * @param parameters    Parameters for all SQL statements
+     */
+    void execute(Iterable<String> sqls, Map<String, ?> parameters);
```
wtritch: I added a method to be able to inject a parameter map that will be used for the entire list of SQL strings.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/configuration/GroupProcessingSqlConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.validation.Valid;
+import java.util.Map;
+
+/**
+ * Configuration for the entities and corresponding SQLs (with an action and sql statement)
+ * for processing a group batch.
+ */
+@Component
+@ConfigurationProperties(prefix = "sql.process-batch")
+public class GroupProcessingSqlConfiguration {
```
wtritch: This configuration holds the SQL statements used by the group processor.  It's modeled after the existing MigrateWarehouseToStagingSqlConfiguration-style configuration classes.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
```diff
@@ -0,0 +1,25 @@
...
+    @Valid
+    @NotNull
+    @Size(min=1)
```
wtritch: Since I don't have access to the hibernate convenience validation annotations like @NotEmpty I fell back on a non-null and min-size 1 check.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Map;
+
+/**
+ * JDBC implementation of a SqlListExecutionRepository.
+ */
+@Repository
+public class JdbcSqlListExecutionRepository implements SqlListExecutionRepository {
```
wtritch: basic sql list execution repository

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/ProcessingReferenceService.java
```diff
@@ -0,0 +1,17 @@
+package org.opentestsystem.rdw.ingest.group.service;
+
+/**
+ * Implementations of this interface are responsible for initializing a batch
+ * import with known foreign key ids.
+ */
+public interface ProcessingReferenceService {
```
wtritch: This service wraps one of the steps called out in the PR overview.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
```diff
@@ -0,0 +1,78 @@
...
+                .build();
+
+        try {
+            repository.execute(sql, ImmutableMap.of("batch_id", batchId));
+            triggerImport(batchId);
+        } catch (final Exception e) {
+            logger.error("Failure importing Users", e);
+            abortImport(batchId);
+            throw e;
+        }
```
wtritch: I don't want to leave an import record in the 0 status if an exception is thrown during this process.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -0,0 +1,174 @@
...
+            WHERE batch_id = :batch_id
+
+
+      # ------------ Validate ------------------------------------------------------------------
+      validate:
+        sql:
+          invalid-school-exists: >-
+            SELECT 1 FROM upload_student_group WHERE school_id IS NULL AND batch_id = :batch_id LIMIT 1;
+          invalid-subject-exists: >-
+            SELECT 1 FROM upload_student_group WHERE subject_id IS NULL AND batch_id = :batch_id LIMIT 1;
```
wtritch: With front-end validation I'm not sure if/when/what I'm going to validate... but I've left these in here as a note until I hit the end

group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/RepositoryBackedIT.java
```diff
@@ -0,0 +1,30 @@
...
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@Transactional
+@Import({
+        DataSourceConfiguration.class,
+        GroupProcessingSqlConfiguration.class,
+        JdbcSqlListExecutionRepository.class
+})
+@ActiveProfiles("test")
+public abstract class RepositoryBackedIT {
```
wtritch: Since the services are basically thin shims over the sql list repository, I've had to create repository-backed integration tests for each of them.

group-processor/src/test/resources/WarehouseCodesSetup.sql
```diff
@@ -0,0 +1,19 @@
+--
+-- Used from warehouse.
+-- Need to make sure constraints to not conflict with initial DML
```
wtritch: I shamelessly stole these setup sql files from the migration module and made some minor mods to facilitate my testing.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
```diff
@@ -0,0 +1,25 @@
...
+    public Map<String, String> sql;
+
+    @NotNull
+    public Map<String, String> getSql() {
+        return sql;
+    }
+
+    public void setSql(final Map<String, String> sql) {
```
malaffoon: i haven't read far enough to know how this is used but it doesn't follow our pattern of immutable objects with immutable internals ... perhaps a constructor where you copy the map, and no setter?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/ProcessingReferenceService.java
malaffoon: Hmm, they all seem like just Function<Long, Void>, why give them special names?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
malaffoon: What if the other steps throw an exception? Don't they have the same issue?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
```diff
@@ -0,0 +1,78 @@
...
+                "batch_id", batchId,
+                "import_status", 1));
+    }
+
+    private void abortImport(final long batchId) {
+        final SqlListBuilder sqlListBuilder = new SqlListBuilder(sqlConfiguration.getEntities());
+        final List<String> sql = sqlListBuilder
+                .addNext("import-students", "update-migration")
+                .build();
+        repository.execute(sql, ImmutableMap.of(
```
malaffoon: Can we get more information? Perhaps pass the exception getMessage() or something and record it in import message?

group-processor/src/test/resources/WarehouseCodesSetup.sql
malaffoon: I disagree with the premise: duplication in test data is not a bad thing. It keeps the tests independent. Since my reason isn't in the list i'll take option #2, too much architecture for 3 files.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Autowired
+    private NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Transactional
+    @Override
+    public void execute(final Iterable<String> sqls) {
```
malaffoon: Hmm, i am on the fence as to whether this constitutes a sufficient about of code to factor out a common repo. But, meh, now that it is done.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
```diff
@@ -0,0 +1,39 @@
...
+        for (final String sql : sqls) {
+            jdbcTemplate
+                    .getJdbcOperations()
```
malaffoon: i do know that this is not an expensive operation but getJdbcOperations() should be moved out of the loop.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    @Transactional
+    @Override
+    public void execute(final Iterable<String> sqls, final Map<String, ?> parameters) {
+
+        for (final String sql : sqls) {
+            jdbcTemplate
+                    .update(sql, parameters);
```
malaffoon: This is converting the Map into a parameter source with every call. It would be more efficient to convert it once outside the loop and then pass it in. Personally, i'd take it a step further and steal the innards of NamedParameterJdbcTemplate getting rid of it completely and just using a plain ol' JdbcTemplate. As long as we're going to have this helper repo, it might as well do some optimization for us.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
malaffoon: Since this is being constructed for us by Spring injection, do we need all that? The default is that everything is required, right?

group-processor/src/main/resources/application.sql.yml
```diff
@@ -0,0 +1,174 @@
...
+              SELECT DISTINCT
+                0                           AS status,
+                5                           AS content,
+                'group batch student'       AS contentType,
+                cast(school_id as CHAR(32)) AS digest,
+                cast(:batch_id AS CHAR(50)) AS batch
+              FROM upload_student_group_import sgl
+              WHERE ref_type IN (0, 1)
+                    AND batch_id = :batch_id
+
```
agorina: There was a TODO in the 'hack' script to alter import table to better support this ingest. Did you choose not to do it for now? Or not to do it at all?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
```diff
@@ -1,19 +1,29 @@
...
-    @Autowired
-    public DefaultGroupProcessor() {
+    private final ProcessingStandardizationService standardizationService;
+    private final ProcessingReferenceService referenceService;
+    private final ProcessingUserImportService userImportService;
 
+    @Autowired
+    public DefaultGroupProcessor(final ProcessingStandardizationService standardizationService,
+                                 final ProcessingReferenceService referenceService,
+                                 final ProcessingUserImportService userImportService) {
```
agorina: Should there be a 'validation' service too: 
- does the user have access to the schools in the batch? 
- can we identify the schools?
-  is the group size < 200? and so on. 

I admit I have not finished the review and maybe it is done somewhere. In this case, just ignore this comment.

group-processor/src/main/resources/application.sql.yml
agorina: This answers my question above.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
```diff
@@ -0,0 +1,78 @@
...
+    public void importStudents(final long batchId) {
+        final SqlListBuilder sqlListBuilder = new SqlListBuilder(sqlConfiguration.getEntities());
+        final List<String> sql = sqlListBuilder
+                .addNext("import-students", "stage-new")
+                .addNext("import-students", "stage-deleted")
+                .addNext("import-students", "create-imports")
+                .addNext("import-students", "assign-imports")
+                .addNext("import-students", "propagate-imports")
+                .addNext("import-students", "insert-missing")
+                .addNext("import-students", "update-deleted")
```
agorina: Do we care about left-over data in a case of a failure?

Since there is no transaction wrapping (specifically around the last two statements), we may end up with a partially ingested and failed student group file. It is not going to break anything, so probably alright. But I want to see what other people think.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
wtritch: I don't think configuration properties are required by default, and this has caught a couple of typos for me.

common/src/main/java/org/opentestsystem/rdw/ingest/common/model/EntitySqls.java
wtritch: This was just moved unchanged from a nested class to the common lib.  It's a configuration property rather than a true model, and is only instantiated by Spring as part of a @ConfigurationProperties class. I wonder if this belongs in either the config or util package instead.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
wtritch: This impl is in the group-processor, I only moved the interface to common

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
wtritch: good call, done

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcSqlListExecutionRepository.java
wtritch: Changed to create the SqlParameterSource once outside the loop.  I played with ripping the guts out of the NamedParameterJdbcTemplate, but it's doing some stuff like caching sql-strings to PreparedStatements that I don't want to lose or re-write.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/ProcessingReferenceService.java
wtritch: lol, I broke it out because when I was reading through the hack-script it took me a while to figure out where the sections were, what each section was responsible for, how it was accomplishing it's goal, etc.  I figured grouping the SQL statements by objective would make for more readable code and easier testing.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
wtritch: We're doing the bulk of the validation in in the RDW_Admin front-end where we can give immediate feedback to the user during the upload process.  Check out the [DefaultCsvValidationService](https://github.com/SmarterApp/RDW_Admin/blob/develop/webapp/src/main/java/org/opentestsystem/rdw/admin/service/impl/DefaultCsvValidationService.java) 
I'm still debating what/where we want to validate in the processor, should we double-up on all the validations?  Is that a problem for later?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
wtritch: I believe the repository.execution method is annotated as @Transactional ...which means I probably don't need that abortImport/triggerImport method breakout at all...

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
wtritch: My thought is that each service cleans up after itself, and we have a general exception handler at the top somewhere in DefaultGroupProcessorService that flags the batch as "FAILED" or invalid with a message.  TODO =)

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
chdietz: Validation was moved to the API layer and Will tested the performance.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingUserImportService.java
wtritch: My thought is that's handled farther up the chain where we do the same error handling for any exception during group processing.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
agorina: Thank you for the update. In that case we are good.

group-processor/src/main/resources/application.sql.yml
wtritch: I wasn't sure exactly what the alteration would be.  Did you mean add two additional columns to avoid the casting?  Since there should be a relatively small number of imports I don't think the casting will kill us in the short-term.  Thoughts?

group-processor/src/test/resources/WarehouseCodesSetup.sql
wtritch: Sweet, will leave as-is.

chdietz: I agree that the test files for migrate do not need to be common or shared.  They are very specific individual step and then combined step end to end tests that check counts.

wtritch:
After discussion with @agorina I removed the @Transactional annotation from JdbcSqlListExecutionRepository and pulled transactional annotations up to the service layer.  I also broke the ProcessingUserImportService interface from one method into four with transactional annotations on inserting new students and updating deleted students.

PR #311, opened 2017-08-24
Feature/groups module

rcordeau:
There is quite a bit in here that needs to be changed. Not as elegant as I had hoped. Tests are lacking - need assistance.


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/ProcessingLoadService.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Implementations of this interface are responsible for initializing a group
+ * batch data load into the batch upload table
+ *
+ */
+public interface ProcessingLoadService {
+
+    /**
+     * Load the batch upload table with data located via the digest of the data
+     * @param digest
```
malaffoon: nit: javadocs incomplete

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public void loadBatch(final String digest, final long batchId) {
+        final String location = new LocationStrategy.GroupUploadContentLocationStrategy().location(digest);
+
+        //Check if file already exists in ArchiveService
+        if (archiveService.exists(location)) {
+            Properties props = archiveService.readProperties(location);
+            List<String> sql;
+            final String uri = props.getProperty("rooturi") + "/" + location;
+            logger.info(uri);
```
malaffoon: this will be annoying, please remove or add additional verbiage and drop to debug()

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -0,0 +1,74 @@
...
+     * Load the batch upload table with data located via the digest of the data
+     * @param digest
+     * @param batchId
+     */
+    @Override
+    public void loadBatch(final String digest, final long batchId) {
+        final String location = new LocationStrategy.GroupUploadContentLocationStrategy().location(digest);
+
+        //Check if file already exists in ArchiveService
+        if (archiveService.exists(location)) {
```
malaffoon: What if it doesn't exist? Should we be logging something, is that an internal error?

Also, guard-code style is a little nicer to avoid indenting the whole method body, i.e.:
```
if (!archiveService.exists(location)) {
    logger.info(...);
    return;
}
```

group-processor/src/main/resources/application.sql.yml
```diff
@@ -398,3 +398,28 @@ sql:
...
+            IGNORE 1 LINES
+            (group_name, school_natural_id, school_year, @vsubject_code, @vstudent_ssid, @vgroup_user_login)
+            SET
+              batch_id = :batch_id,
+              subject_code = nullif(@vsubject_code,''),
+              student_ssid = nullif(@vstudent_ssid,''),
+              group_user_login = nullif(@vgroup_user_login,'')
+
+          fromLocal: >-
+            LOAD DATA LOCAL INFILE :uri INTO TABLE upload_student_group FIELDS TERMINATED BY ',' LINES TERMINATED BY '\r'
```
malaffoon: If everything else is exactly the same except `S3` and `INFILE` you could put a placeholder in here and then substitute it at runtime. Not worth changing if that is the only change but if you're making other changes ...

group-processor/src/test/resources/sample-group-001.csv
```diff
@@ -0,0 +1,13 @@
+group_name,school_natural_id,school_year,subject_code,student_ssid,group_user_login
```
malaffoon: perhaps some error csv's with appropriate unit tests showing failure path

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -0,0 +1,74 @@
...
+            Properties props = archiveService.readProperties(location);
+            List<String> sql;
+            final String uri = props.getProperty("rooturi") + "/" + location;
```
malaffoon: Yeah, regarding comment in common changes ... we don't want to be doing this logic here. This should be just `final String uri = props.getProperty(RawURI);`

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
```diff
@@ -37,7 +39,8 @@ public void process(final Message<?> message) {
         logger.info("Received payload: " + (payload.length() > 80 ? payload.substring(0, 80) + "..." : payload));
 
         try {
-            groupProcessor.process(payload);
+            GroupMessage gm = new Gson().fromJson(payload, GroupMessage.class);
```
wtritch: Recommend using standard ObjectMapper here

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
wtritch: I would recommend throwing to stop processing if it doesn't exist.  No reason to run the rest of the import logic, and it looks like it'll get caught and logged by the caller.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -398,3 +398,28 @@ sql:
...
+        sql:
+          fromS3: >-
+            LOAD DATA FROM S3 :uri INTO TABLE upload_student_group FIELDS TERMINATED BY ',' LINES TERMINATED BY '\r'
+            IGNORE 1 LINES
+            (group_name, school_natural_id, school_year, @vsubject_code, @vstudent_ssid, @vgroup_user_login)
+            SET
+              batch_id = :batch_id,
+              subject_code = nullif(@vsubject_code,''),
+              student_ssid = nullif(@vstudent_ssid,''),
+              group_user_login = nullif(@vgroup_user_login,'')
```
wtritch: Neat, this will let us get rid of those '' -> null statements above.  We'll still need the subject's 'All' -> null unfortunately.

group-processor/src/test/resources/sample-group-001.csv
wtritch: The CSV's will have been validated before reaching this path.

PR #312, opened 2017-08-24
Feature/add group upload status update

wtritch:
Currently, once the user has uploaded a CSV they are not sure if or when the upload has been processed.  This PR updates the group-processor to update the batch status from ACCEPTED to either INVALID on failure or PROCESSED on completion.
If INVALID, the exception message is set as the batch message.

### DIFF
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
@@ -2,6 +2,8 @@ package org.opentestsystem.rdw.ingest.group;
 
 import com.google.gson.Gson;
 import org.opentestsystem.rdw.ingest.common.model.GroupMessage;
+import org.opentestsystem.rdw.ingest.common.model.ImportStatus;
+import org.opentestsystem.rdw.ingest.group.repository.StudentGroupBatchRepository;
 import org.opentestsystem.rdw.ingest.group.service.GroupProcessor;
 import org.opentestsystem.rdw.messaging.RdwMessageHeaderAccessor;
 import org.slf4j.Logger;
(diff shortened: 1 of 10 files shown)


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
```diff
@@ -21,28 +23,38 @@
...
+    public GroupProcessorConfiguration(final GroupProcessor groupProcessor,
+                                       final StudentGroupBatchRepository groupBatchRepository) {
         this.groupProcessor = groupProcessor;
+        this.groupBatchRepository = groupBatchRepository;
     }
 
     @ServiceActivator(inputChannel = Processor.INPUT)
     public void process(final Message<?> message) {
+        RdwMessageHeaderAccessor accessor = null;
+        try {
```
wtritch: I moved the try block up to also encompass basic message parsing in case the payload has been corrupted.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
```diff
@@ -21,28 +23,38 @@
...
+            final GroupMessage gm = new Gson().fromJson(payload, GroupMessage.class);
             groupProcessor.process(gm);
         } catch (final Exception e) {
             logger.error("failed with an unexpected import error: " + e.getMessage());
+            if (accessor != null) {
+                try {
+                    groupBatchRepository.updateBatchStatus(accessor.getImportId(), ImportStatus.INVALID, e.getMessage());
+                } catch (final Exception e2) {
+                    logger.error("Failed to update batch status: {}", accessor.getImportId());
+                }
```
wtritch: If I can at least get the batch id out of the message header, I update the batch status to INVALID on failure.
Unfortunately, there is no FAILED ImportStatus, INVALID was the closest I could get without adding an additional ImportStatus (which started to rabbit-hole a bit.)

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
```diff
@@ -50,12 +55,6 @@ public void process(final GroupMessage message) {
         final long batchId = message.getUploadId();
         try {
             loadService.loadBatch(message.getDigest(), batchId);
-        } catch (final Exception e) {
-            logger.error("Problem loading data for: {}", message.getDigest(), e);
-            return;
-        }
-
-        try {
```
wtritch: I combined the load call with the other service calls in a single try statement.

wtritch:
Merging in the interested of speed, will resolve future comments as they come in

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
malaffoon: i'm about 98% sure that wrapping the message and getting the payload can't throw but okay ...

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
malaffoon: All statuses that are negative can be considered to be failures. They just provide more granularity on the failure reason.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/StudentGroupBatchRepository.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.ingest.group.repository;
+
+import org.opentestsystem.rdw.ingest.common.model.ImportStatus;
+
+/**
+ * Implementations of this interface are responsible for working with student group
+ * upload batches.
+ */
+public interface StudentGroupBatchRepository {
```
malaffoon: what's wrong with org.opentestsystem.rdw.ingest.common.ImportRepository which has this exact method in it?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
```diff
@@ -69,9 +68,15 @@ public void process(final GroupMessage message) {
             groupImportService.updateModifiedGroupUsers(batchId);
             groupImportService.updateModifiedGroupStudents(batchId);
             groupImportService.triggerImport(batchId);
+            updateBatch(batchId, ImportStatus.PROCESSED, "");
```
malaffoon: I prefer the pattern established elsewhere (e.g. package processor) where the processors (the impl details) just throw an ImportException and the overall handler updates the import record. That way the processors don't have to know about import explicitly.

wtritch: Ah, I see the pattern now.  Will file a PR with the changes.

PR #313, opened 2017-08-24
DWR-920 fix group uploads with multiple entries for the same user/group

wtritch:
This PR fixes uploads that register the same user to the same group in multiple rows.
I've also made changes to handle registering the same student to the same group in multiple rows.

### DIFF
--- a/group-processor/src/main/resources/application.sql.yml
+++ b/group-processor/src/main/resources/application.sql.yml
@@ -361,7 +361,7 @@ sql:
             WHERE users.student_group_id IS NOT NULL
 
           add-updated-users: >-
-            INSERT INTO user_student_group (student_group_id, user_login)
+            INSERT IGNORE INTO user_student_group (student_group_id, user_login)
               SELECT
                 upload.group_id,
                 upload.group_user_login
@@ -381,7 +381,7 @@ sql:
             WHERE students.student_group_id IS NOT NULL
 
           add-updated-students: >-
-            INSERT INTO student_group_membership (student_group_id, student_id)
+            INSERT IGNORE INTO student_group_membership (student_group_id, student_id)
               SELECT
                 upload.group_id,
                 upload.student_id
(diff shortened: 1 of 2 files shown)


group-processor/src/main/resources/application.sql.yml
```diff
@@ -361,7 +361,7 @@ sql:
             WHERE users.student_group_id IS NOT NULL
 
           add-updated-users: >-
-            INSERT INTO user_student_group (student_group_id, user_login)
+            INSERT IGNORE INTO user_student_group (student_group_id, user_login)
```
wtritch: I believe @agorina called this out as a potential issue in the first PR... and I didn't understand exactly what she was warning about.

group-processor/src/test/resources/WarehouseGroupImportServiceSetup.sql
```diff
@@ -9,6 +9,7 @@ INSERT INTO upload_student_group (batch_id, group_name, school_year, school_natu
   (33, 'Test Student Group 7', 2017, 'natural_id-98', -98, '88', -88, -7, null, null),
   -- User change groups
   (33, 'Test Student Group 8', 2017, 'natural_id-1', -1, null, null, -8, 'user1@somewhere.com', null),
+  (33, 'Test Student Group 8', 2017, 'natural_id-1', -1, null, null, -8, 'user1@somewhere.com', null),
```
wtritch: Updated test SQL to validate the fix.

wtritch:
Merging in the interests of speed, will continue to resolve comments as they come in.

group-processor/src/main/resources/application.sql.yml
chdietz: Hmm, so insert ignore will not add the duplicate entry for the user.  I guess if we wanted to be more strict we could fail this on validation to give them a waring the CSV has duplicate entries, doing so may help them find an error in their logic used to generate the file?

PR #315, opened 2017-08-24
groups modifications

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/315)


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -50,25 +60,35 @@ public void loadBatch(final String digest, final long batchId) {
...
-            if(uri.startsWith(s3Prefix)) {
-                sql = sqlListBuilder.addNext("load", "fromS3").build();
-            }
-            else {
-                sql = sqlListBuilder.addNext("load", "fromLocal").build();
-            }
-            repository.execute(sql, ImmutableMap.of("uri", uri,
-                    "batch_id", batchId));
+        if (!archiveService.exists(location)) {
+            logger.debug("Unable to locate " + location + " via the archive service");
```
jtreuting: if this is throwing an exception, shouldn't that be a logger.error?

rcordeau: Yes, good call.

wtritch: per @malaffoon logger.warn might be more appropriate?

malaffoon: log-and-throw is an antipattern; will the exception be logged further up?

rcordeau: I made it as .error for now.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -50,25 +60,35 @@ public void loadBatch(final String digest, final long batchId) {
...
+        }
+        else if(uri.startsWith(localPrefix)){
+            sqlSource = "LOCAL INFILE";
+            uri = uri.substring(localPrefix.length());
+        }
+        else {
+            throw new IllegalArgumentException("Invalid URI for location " + location);
+        }
+
+        List<String> sql = sqlListBuilder.addNext(String.format(rawSql, sqlSource)).build();
```
malaffoon: This'll work but i have two concerns: String.format can do unexpected things, and %s placeholder is pretty non-specific. I'd lean towards putting `%%sqlSource%%` in the sql string resource and doing a substr() here.

rcordeau: Will fix.

rcordeau: How about replace(s1, s2)?

PR #318, opened 2017-08-25
Stop modifying imports as csv upload batches

wtritch:
Fix a bug where we were modifying Imports as upload batches.


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
```diff
@@ -48,12 +48,12 @@ public void process(final Message<?> message) {
 
             final GroupMessage gm = new Gson().fromJson(payload, GroupMessage.class);
             groupProcessor.process(gm);
-            importRepository.updateStatusAndMessageById(gm.getUploadId(), ImportStatus.PROCESSED, "");
+            batchRepository.updateBatchStatus(gm.getUploadId(), ImportStatus.PROCESSED, "");
```
wtritch: I'm not sure what kind of havoc I've been causing by modifying random Import statuses, but it stops here.

PR #319, opened 2017-08-25
DWR-940 Migrate all modified groups

wtritch:
This PR updates all groups with their assigned import_id during group processing rather than just groups that have been created, deleted, or had a subject change.

### DIFF
--- a/group-processor/src/main/resources/application.sql.yml
+++ b/group-processor/src/main/resources/application.sql.yml
@@ -348,7 +348,6 @@ sql:
                 ON sg.id = loading.group_id
             SET sg.subject_id = loading.subject_id,
               sg.update_import_id = loading.import_id
-            WHERE NOT sg.subject_id <=> loading.subject_id
 
           clear-updated-users: >-
             DELETE users FROM user_student_group users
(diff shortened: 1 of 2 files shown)


group-processor/src/main/resources/application.sql.yml
```diff
@@ -348,7 +348,6 @@ sql:
                 ON sg.id = loading.group_id
             SET sg.subject_id = loading.subject_id,
               sg.update_import_id = loading.import_id
-            WHERE NOT sg.subject_id <=> loading.subject_id
```
wtritch: This is the big fix, we want to bind all groups that have been modified to their import_ids, not just groups with a subject change.

group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
```diff
@@ -225,20 +228,35 @@ public void itShouldUpdateModifiedGroupStudents() {
...
                 "SELECT * from student_group_membership WHERE student_group_id = -8",
                 new HashMap<>());
         assertThat(group8Students).hasSize(2);
         assertThat(group8Students.stream().map(row -> row.get("student_id")))
                 .containsOnly(-88, -86);
 
+        group = template.queryForMap(
+                "SELECT * from student_group WHERE id = -7",
+                new HashMap<>());
+        assertThat(((Timestamp) group.get("updated")).toInstant()).isGreaterThan(Instant.now().minus(Duration.ofMinutes(1)));
```
wtritch: This was previously failing, since group "-7" only had a membership change, it wasn't being bound to an import id and it's "updated" column wasn't being bumped.

wtritch:
Merging this in the name of speed, will resolve comments as they come in.

PR #320, opened 2017-08-25
updating to deploy docker images to new spot in docker hub

rcordeau:
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_ingest/320)


malaffoon:
same comment as reporting: change docker-compose and README to get rid of all reference to fwsbac.

PR #321, opened 2017-08-26
add actuator end-points for triggering tasks

malaffoon:
During deployment there were a number of times that i wanted to trigger the tasks to test configuration. I had to resort to changing the configuration file (setting cron to every 5 minutes or something), pushing the config changes, waiting, reverting the config changes. 

This lets a client post to actuator end-points to trigger a task.

### DIFF
--- a/task-service/src/main/resources/application.yml
+++ b/task-service/src/main/resources/application.yml
@@ -66,7 +66,7 @@ task:
 #    cron: 0 0 4 * * *
 #    state: CA
 #    art-client:
-#      districts-url: https://art-deployment.sbtds.org/rest/district?stateAbbreviation={state}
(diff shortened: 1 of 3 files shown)


task-service/src/main/resources/application.yml
```diff
@@ -66,7 +66,7 @@ task:
 #    cron: 0 0 4 * * *
 #    state: CA
 #    art-client:
-#      districts-url: https://art-deployment.sbtds.org/rest/district?stateAbbreviation={state}
+#      districts-url: https://art-deployment.sbtds.org/rest/district?stateAbbreviation={state}&pageSize=5000
```
malaffoon: unrelated; i noticed during deployment that this was missing in my copy&paste versions of the configuration.

PR #322, opened 2017-08-28
DWR-932 Spread group processor imports over time

wtritch:
Previously while processing a group batch upload we were creating:
* An import record per school for processing student creation/modification
* An import record per school for processing group creation/modification

However, each student import for each school had the exact same "updated" timestamp, and each group import for each school had the exact same "updated" timestamp.  This means that although we may only *intend* to process half of the imports, *all* of the student or *all* of the group data would be migrated at once, do to being within the migration start-time/end-time boundaries.

This PR uses java to loop over each school individually, naturally spreading the imports in time and allowing bite-sized migration chunks.

### DIFF
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/SchoolBatchingGroupImportService.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/SchoolBatchingGroupImportService.java
@@ -1,0 +1,44 @@
+package org.opentestsystem.rdw.ingest.group.service.impl;
+
+import org.opentestsystem.rdw.ingest.group.repository.SchoolBatchRepository;
+import org.opentestsystem.rdw.ingest.group.service.BatchingGroupImportService;
+import org.opentestsystem.rdw.ingest.group.service.ProcessingGroupImportService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.Set;
+
+/**
+ * This service processes group imports and modifications for a Group upload
+ * while grouping imports together by school.
+ */
+@Service
+public class SchoolBatchingGroupImportService implements BatchingGroupImportService {
+
+    private final ProcessingGroupImportService importService;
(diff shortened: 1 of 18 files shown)


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/SchoolBatchingGroupImportService.java
```diff
@@ -0,0 +1,44 @@
...
+            importService.insertMissingGroups(batchId, schoolId);
+            importService.updateDeletedGroups(batchId, schoolId);
+            importService.updateModifiedGroups(batchId, schoolId);
+            importService.updateModifiedGroupUsers(batchId, schoolId);
+            importService.updateModifiedGroupStudents(batchId, schoolId);
+            importService.triggerImport(batchId, schoolId);
+        }
+
+        importService.cleanupBatch(batchId);
+    }
```
wtritch: This is the new implementation.
1. get the schools for this batch
2. for each school, perform the group staging/modifying/importing steps
3. after looping, clean up (remove interim table records)

chdietz:
Just to be clear on this: "This means that although we may only intend to process half of the imports, all of the student or all of the group data would be migrated at once..." 

- If there were 5000 imports with the exact same time stamp and a migrate.batch.size of 2000, migrate would do 2000, 2000 then 1000 ( if there were no more) correct or not?

wtritch:
Correct, even though we may INTEND to only process 2000 imports, if there are 10000 with the same "updated" date, all of the data for those 10000 imports would be processed in a single migration, since it's timestamp-based, not import-id-based.

malaffoon:
I have to admit, i was expecting more of a hack that just did something for that last update-migration step (i was hopeful there was a way to partition the update by school right at that last step). But this will work too.

wtritch:
@malaffoon I considered that, but the actual student and student_group records need their update date-times to be spread out as well.  Otherwise, we'd have nice import spread, but their contents would all still be migrated with the first import.

PR #324, opened 2017-08-29
DWR-758 Datamart: add school_year table to datamart

agorina:


migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -132,6 +132,24 @@ sql:
                DELETE ra FROM reporting.accommodation ra
                 WHERE NOT EXISTS(SELECT sa.id FROM reporting.staging_accommodation sa WHERE sa.id = ra.id);
 
+      # ------------ School Year ------------------------------------------------------------------------
+        school_year:
+          sql:
+            update:  >-
+              SELECT 1;
```
agorina: There is no update for school_year, but the generic task code expects it for all code.
Another option is to put a conditional execution into the steps, but I think this is cleaner.

migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -115,6 +116,16 @@ sql:
               INSERT INTO reporting.staging_accommodation (id, code) VALUES
                 (:id, :code)
 
+        # ------------ accommodation  -------------------------------------------------------------
```
malaffoon: shouldn't this be `school_year`?

agorina: yes, it should. Thank you!

PR #325, opened 2017-08-31
working version is 1.0.0

malaffoon:
The CI process is such that `develop` should have the version number set to what we're working towards, in this case 1.0.0.

I've already bumped common and schema.

Note that i am not resetting the build number cause it really should've been 1.0.0 for a while.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ git checkout develop
 Then to use those new changes, you can specify the SNAPSHOT version of RDW_Common
 ```bash
 //In RDW_Ingest
-./gradlew build it -Pcommon=0.0.1-SNAPSHOT
+./gradlew build it -Pcommon=1.0.0-SNAPSHOT
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -129,8 +129,8 @@ Running the applications locally depends on the local database being configured
 ```bash
 To completely clean out any existing data you might have and start fresh:
 ./gradlew cleanallprod migrateallprod
-or, if you want to use a different version of the schema, say version 0.0.1-68 of RDW_Schema
-./gradlew -Pschema=0.0.1-68 cleanallprod migrateallprod
+or, if you want to use a different version of the schema, say version 1.0.0-68 of RDW_Schema
+./gradlew -Pschema=1.0.0-68 cleanallprod migrateallprod
```
malaffoon: @rcordeau are these instructions still right?

rcordeau: Yes, those are still correct - cleanAllProd and migrateAllProd still exist to clean out the non-test database.

PR #328, opened 2017-09-01
remove school caches

malaffoon:
Pending the results of the investigation of how much the school cache helps, this simply removes it. The risk of rejecting test results outweighs the performance improvements. At least until the test result volume picks up.

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -12,7 +12,7 @@ server:
 spring:
   cache:
     type: SIMPLE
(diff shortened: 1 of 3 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -29,8 +29,8 @@ spring:
     validationQuery: SELECT 1
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
-    maxActive: 10
-    initialSize: 4
+    maxActive: 12
+    initialSize: 6
```
malaffoon: I ran into an issue where the system appeared to need more connections in the db pool when i disabled all caching. Pretty sure it won't be an issue just removing two of the caches but, just in case ...

PR #329, opened 2017-09-01
DWR-969 Allow adding users/students to group with no users/students

wtritch:
Allow clearing users/students from group with users/students

This PR updates the SQL that detects user/student membership changes for groups to account for empty user/student membership lists in both the existing group and uploaded CSV.

### DIFF
--- a/group-processor/src/main/resources/application.sql.yml
+++ b/group-processor/src/main/resources/application.sql.yml
@@ -195,63 +195,75 @@ sql:
               SELECT
                 batch_id,
                 school_id,
-                concat(loading.group_name, loading.school_id, loading.school_year),
+                concat(name, school_id, school_year),
                 3
               FROM (
-                     SELECT
-                       batch_id,
-                       school_id,
-                       group_id,
-                       group_name,
-                       GROUP_CONCAT(student_id ORDER BY student_id) AS students,
-                       school_year
-                     FROM upload_student_group
-                     WHERE student_id IS NOT NULL
-                           AND batch_id = :batch_id
-                           AND school_id = :school_id
-                     GROUP BY group_id) AS loading
-                JOIN
-                (
                   SELECT
-                    student_group_id,
-                    GROUP_CONCAT(student_id ORDER BY student_id) AS students
-                  FROM student_group_membership
-                  GROUP BY student_group_id
+                    grp.id,
+                    grp.school_id,
+                    grp.name,
+                    grp.school_year,
+                    upload.batch_id,
+                    GROUP_CONCAT(membership.student_id ORDER BY membership.student_id) AS students
+                  FROM student_group grp
+                    JOIN upload_student_group upload
+                      ON grp.id = upload.group_id
+                    LEFT JOIN student_group_membership membership
+                      ON membership.student_group_id = grp.id
+                  WHERE upload.batch_id = :batch_id
+                    AND upload.school_id = :school_id
+                  GROUP BY grp.id
                 ) AS existing
-                  ON existing.student_group_id = loading.group_id
-              WHERE existing.students <> loading.students;
+              LEFT JOIN (
+                  SELECT
+                    group_id,
+                    GROUP_CONCAT(student_id ORDER BY student_id) AS students
+                  FROM upload_student_group
+                  WHERE student_id IS NOT NULL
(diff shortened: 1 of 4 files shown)


group-processor/src/main/resources/application.sql.yml
```diff
@@ -195,63 +195,75 @@ sql:
...
+                    group_id,
+                    GROUP_CONCAT(student_id ORDER BY student_id) AS students
+                  FROM upload_student_group
+                  WHERE student_id IS NOT NULL
+                        AND batch_id = :batch_id
+                        AND school_id = :school_id
+                  GROUP BY group_id
+                ) AS loading
+                ON existing.id = loading.group_id
+              WHERE NOT existing.students <=> loading.students
```
wtritch: I had to invert this query a bit to account for missing student_group_membership records on existing groups and group upload records that don't have any students for the uploaded group.
Now it:
1. finds all groups referenced by the uploaded csv
2. concats all existing member student ids into a string (null if no students are members)
3. left-joins to a table that contains all students in the uploaded group (I need the left join in case the uploaded group contains no students)
4. registers an import for the group if the two concatenated student membership strings are different using null-safe comparison.

I needed similar processing for modifying users assigned to a group below.

group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
```diff
@@ -253,8 +253,48 @@ public void itShouldUpdateModifiedGroupStudents() {
...
+
+    @Test
+    public void itShouldClearUsersFromPopulatedGroup() {
+        createImports(37);
+
+        final Map<String, Object> groupWithClearedStudents = template.queryForMap(
+                "SELECT * from upload_student_group where group_id = -2",
+                new HashMap<>());
+        assertThat(groupWithClearedStudents.get("import_id")).isNotNull();
+    }
```
wtritch: initially-failing tests that now work with the changes above.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -195,63 +195,75 @@ sql:
               SELECT
                 batch_id,
                 school_id,
-                concat(loading.group_name, loading.school_id, loading.school_year),
+                concat(name, school_id, school_year),
```
millerts: I personally like to call out the alias in the select but I understand it isn't necessary here.

group-processor/src/main/resources/application.sql.yml
agorina: Just curious: have you tried on a largish db to see the response time?

wtritch: I have not.  I'm hoping that scoping via school and batch will keep times reasonable.

agorina: I think we should try it. Not urgent, maybe create a JIRA to test groups process for performance?

wtritch: query_plan
This is the query plan for the stage-membership-change.  It looks like accessing existing table data (student_group, student_group_membership) will be nice and indexed so time shouldn't scale with the number of records in the system.  
Unfortunately, we have 2 full-table-scans on the upload_student_group table (where one record represents one CSV row.)
As long as our CSVs are relatively small this shouldn't be a performance problem.  Performance will drop with large CSV uploads.
The only solution I see is to add an index to the upload_student_group table on (batch_id, school_id)  but I wonder if that's pre-emptive optimization?

agorina: I agree with your comment. I did try to index the upload table when I was doing a hack version of the scripts and I did not see a pay back. The load time was slower, and the JOIN did not improve. Just to check - we are deleting the data from the loading table once it is loaded, is it correct?

And thank you for checking this. What tool do you use for this nice visualization of the execution plan?

wtritch: Yup, the data is being removed from the upload_student_group table on SUCCESSFUL processing.  It's left in there for debugging on ingest failure.

For the visualization, I just used the standard MySQL Workbench Query->Explain Current Statement command.

And np, I probably should have done an "explain plan" before posting the PR...

PR #331, opened 2017-09-04
slight efficiency improvements

malaffoon:
The code before was doing the query into a `List<Map<key,value>>` and then converting that into a `List<MapSqlParameterSource>`. This eliminates that middle collection, converting directly to the parameter sources. Not sure how much it will really reduce memory consumption but it won't hurt.

I wanted to take it a step further, eliminating the collection of results completely. But the mysql driver doesn't support client-side cursors so i found it wasn't actually streaming the results: it collected the entire result set no matter what i did. After reading about it, there is a workaround to that but it involves server-side cursors which uses temporary tables and ... (ugh).

### DIFF
--- a/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlCopyWarehouseToStagingRepository.java
+++ b/migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlCopyWarehouseToStagingRepository.java
@@ -3,15 +3,21 @@ package org.opentestsystem.rdw.ingest.migrate.reporting.repository.impl;
 import org.opentestsystem.rdw.ingest.migrate.reporting.repository.SqlCopyWarehouseToStagingRepository;
 import org.opentestsystem.rdw.ingest.migrate.reporting.step.SqlCopyStatements;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.ColumnMapRowMapper;
+import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.core.namedparam.SqlParameterSource;
+import org.springframework.jdbc.support.JdbcUtils;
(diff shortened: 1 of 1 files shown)


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlCopyWarehouseToStagingRepository.java
```diff
@@ -32,38 +38,59 @@ public JdbcSqlCopyWarehouseToStagingRepository(final NamedParameterJdbcTemplate
     @Override
     @Transactional
     public void execute(final List<SqlCopyStatements> copyStatementsList, final Instant firstAt, final Instant lastAt, final long migrateId) {
+        final MapSqlParameterSource queryParams = new MapSqlParameterSource()
```
malaffoon: Another slight improvement: no need to create the query params for every iteration of the loop.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlCopyWarehouseToStagingRepository.java
```diff
@@ -32,38 +38,59 @@ public JdbcSqlCopyWarehouseToStagingRepository(final NamedParameterJdbcTemplate
...
+                .addValue("first_at", Timestamp.from(firstAt))
+                .addValue("last_at", Timestamp.from(lastAt));
         for (final SqlCopyStatements sqlCopyStatements : copyStatementsList) {
-            final List<Map<String, Object>> result = warehouseJdbcTemplate.queryForList(sqlCopyStatements.getWarehouseRead(),
```
malaffoon: This was creating List<Map<key, value>> ...

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcSqlCopyWarehouseToStagingRepository.java
wtritch: ...I still think paging at 50k or so is still the only way to protect the server =P

PR #332, opened 2017-09-05
DWR-953 Performance: adjust reconciliation report query

agorina:


task-service/src/main/resources/application.sql.yml
```diff
@@ -2,17 +2,22 @@ sql:
...
         JOIN exam e ON e.import_id = i.id
         JOIN exam_student es ON es.id = e.exam_student_id
         JOIN student s ON s.id = es.student_id
-        JOIN asmt a ON a.id = e.asmt_id
-      WHERE i.content = 1
-        AND (:batch IS NULL OR i.batch = :batch)
-        AND (:creator IS NULL OR i.creator = :creator)
-        AND (:before IS NULL OR i.created < :before)
-        AND (:after IS NULL OR i.updated > :after)
-      ORDER BY i.id
```
agorina: This came up in the slow query logs. Before the change it was taking 2m 42s. 
After the change it is under 2 seconds.

PR #336, opened 2017-09-11
refactor import service stuff a little bit

malaffoon:
I'm not 100% sure i'll merge this but having an abstract class in the properties hierarchy was bugging me. This condenses down everything into a single class ... but some of the properties are optional.


common/src/test/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQueryTest.java
```diff
@@ -157,4 +157,11 @@ public void itShouldIgnoreExtraParamsAndLeadingQuestionMarkInParams() {
         assertThat(query.getContent()).isEqualTo(ImportContent.PROBE);
         assertThat(query.asParamString()).isEqualTo("content=PROBE");
     }
+
+    @Test
+    public void itShouldParseProductionSetting() {
+        final RdwImportQuery query = builder().params("creator=eseitembanksupport@ets.org&after=-PT25H").build();
```
malaffoon: unrelated, just me testing something

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/ResubmitUnprocessedImportsConfiguration.java
```diff
@@ -51,12 +53,19 @@ public ResubmitUnprocessedImportsTask resubmitUnprocessedImportsTask(
...
     @Bean
-    public ResubmitUnprocessedImportsService defaultResubmitUnprocessedImportsService(final ImportServiceClient importServiceClient) {
+    public ResubmitUnprocessedImportsService defaultResubmitUnprocessedImportsService(
+            @Qualifier("exams") final ImportServiceClient importServiceClient) {
         return new DefaultResubmitUnprocessedImportsService(importServiceClient);
     }
 
-    @Bean
-    public ImportServiceClient restImportServiceClient(final ExamsImportServiceClientProperties properties) {
+    @Bean(name = "exams")
```
malaffoon: Since ImportServiceClient is now generic, each instance needs a name.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientProperties.java
```diff
@@ -14,8 +12,6 @@
  *     <li>accessTokenUri or access-token-uri</li>
  * </ul>
  */
-@Component
-@ConfigurationProperties(prefix = "task.update-organizations.art-client")
```
malaffoon: We don't want to instantiate these properties if the task isn't configured. This was doing that, but it felt better to stuff it all in the conditional configuration class where it is used.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/ResubmitUnprocessedImportsConfiguration.java
wtritch: maybe a more descriptive name like examsImportClient?

malaffoon: good point, will tweak

PR #337, opened 2017-09-12
fix reconcilation report query evaluation

malaffoon:
I will merge this after approval and smoke testing in awsxxx environment.

### DIFF
--- a/task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/ReconciliationReportTaskTest.java
+++ b/task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/ReconciliationReportTaskTest.java
@@ -1,0 +1,50 @@
+package org.opentestsystem.rdw.ingest.tasking;
+
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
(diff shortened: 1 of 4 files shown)


task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/ReconciliationReportTaskTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+        task.sendReconciliationReport();
+        sleep(200);
+        task.sendReconciliationReport();
+
+        final ArgumentCaptor<RdwImportQuery> captor = ArgumentCaptor.forClass(RdwImportQuery.class);
+        verify(service, times(2)).sendReport(captor.capture(), eq(senders));
+        final Instant after1 = captor.getAllValues().get(0).getAfter();
+        final Instant after2 = captor.getAllValues().get(1).getAfter();
+        assertThat(after1).isNotEqualTo(after2);
```
malaffoon: Before the fix, this would fail because the query was evaluated only once during instantiation of the task.

PR #338, opened 2017-09-13
use property placeholders for url-schema and url-server

malaffoon:
Sometimes Spring makes things easy ...

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -22,9 +22,14 @@ spring:
           group: default
 
   datasource:
-    url-server: localhost:3306
-    url-schema: warehouse
-    url: jdbc:mysql://${spring.datasource.url-server}/${spring.datasource.url-schema}?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true&noAccessToProcedureBodies=true
+    url: "\
+      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}\
+      ?useSSL=${spring.datasource.use-ssl:false}\
+      &useLegacyDatetimeCode=${spring.datasource.use-legacy-datetime-code:false}\
+      &characterEncoding=${spring.datasource.character-encoding:utf8}\
+      &rewriteBatchedStatements=${spring.datasource.rewrite-batched-statements:true}\
+      &noAccessToProcedureBodies=${spring.datasource.no-access-to-procedure-bodies:true}\
+      "
     username: root
     password:
     testWhileIdle: true
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -22,7 +22,9 @@ spring:
           group: default
 
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true&noAccessToProcedureBodies=true
+    url-server: localhost:3306
+    url-schema: warehouse
+    url: jdbc:mysql://${spring.datasource.url-server}/${spring.datasource.url-schema}?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true&noAccessToProcedureBodies=true
     username: root
     password:
     testWhileIdle: true
--- a/exam-processor/src/test/resources/application-test.yml
+++ b/exam-processor/src/test/resources/application-test.yml
@@ -1,3 +1,3 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse_test?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true
+    url-schema: warehouse_test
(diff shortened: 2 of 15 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -22,7 +22,9 @@ spring:
           group: default
 
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true&noAccessToProcedureBodies=true
+    url-server: localhost:3306
+    url-schema: warehouse
+    url: jdbc:mysql://${spring.datasource.url-server}/${spring.datasource.url-schema}?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true&noAccessToProcedureBodies=true
```
malaffoon: I thought about using default values instead of explicitly setting the properties but chose this instead of:
```
url: jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}?useSSL=fa...
```
Thoughts?

Also, are you all good with `url-server` and `url-schema`? I put the `url-` prefix because there is already a `schema` property for datasource properties. And i chose `url-server` over `url-host` because it is intended to include the port.

exam-processor/src/test/resources/application-test.yml
```diff
@@ -1,3 +1,3 @@
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse_test?useSSL=false&useLegacyDatetimeCode=false&characterEncoding=utf8&rewriteBatchedStatements=true
+    url-schema: warehouse_test
```
malaffoon: this demonstrates the whole point: no more copying and pasting all the parameters

exam-processor/src/main/resources/application.yml
agorina: The answer is probably no, but I would like to ask: does spring support a driver specific properties? In other words, adding all the properties after the db URL to the list below along with the username and password would not work?

malaffoon: Are you asking whether parameter-specific overrides like this would be supported, in application-test.yml:
```
    useSSL: true
    useLegacyDatetimeCode: false
...
```
I don't think so. We could add a placeholder every time we add a new parameter i suppose. Using useSSL as an example:
```
url: jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}?useSSL=${spring.datasource.use-ssl:false} ...
```

Know that the overriding config *can* specify the entire url if they want to customize the parameters but then, yeah, they are back in copy&paste / maintenance hell.

agorina: Yes, that was my question. Thank you.

esotrope: Wouldn't putting all the params in one prop allow us to dynamically add/remove them without adding/removing props?
```
url: jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}?${spring.datasource.url-params:}
```

malaffoon: That's a thought. Not sure how to convince Spring to then inject them delimited by &'s.

task-service/src/main/resources/application.yml
```diff
@@ -24,7 +24,12 @@ cloud:
 
 spring:
   datasource:
-    url: jdbc:mysql://localhost:3306/warehouse?useSSL=false&characterEncoding=utf8&useLegacyDatetimeCode=false
+    url: "\
+      jdbc:mysql://${spring.datasource.url-server:localhost:3306}/${spring.datasource.url-schema:warehouse}\
+      ?useSSL=${spring.datasource.use-ssl:false}\
+      &characterEncoding=${spring.datasource.character-encoding:utf8}\
+      &useLegacyDatetimeCode=${spring.datasource.use-legacy-datetime-code:false}\
+      "
```
malaffoon: @agorina, @esotrope: here is the best way i could get it to allow any and all parameters to be individually overridden. The `"\` stuff is necessary so yaml parsing doesn't put spaces into the string.

esotrope: Passes my approval criteria 👍

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/DataSourceConfiguration.java
```diff
@@ -74,8 +74,8 @@ public DataSource warehouseDataSource() {
      * @return DataSource
      */
     @Bean(name = "warehouseJdbcTemplate")
-    public NamedParameterJdbcTemplate warehouseJdbcTemplate() {
-        return new NamedParameterJdbcTemplate(warehouseDataSource());
+    public NamedParameterJdbcTemplate warehouseJdbcTemplate(@Qualifier("warehouseDatasource") final DataSource dataSource) {
+        return new NamedParameterJdbcTemplate(dataSource);
```
malaffoon: unrelated, just making it internally consistent

task-service/src/main/resources/application.yml
agorina: good with me too

PR #339, opened 2017-09-15
make components in common public so AoP can interact with them properly

malaffoon:
Apparently `gradle bootRun` does some AoP magick that doesn't like components in shared libraries to be non-public.


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorApplication.java
```diff
@@ -19,7 +19,7 @@
 })
 public class GroupProcessorApplication {
 
-    public static void main(String[] args) {
+    public static void main(final String[] args) {
```
malaffoon: unrelated, fix warning

PR #340, opened 2017-09-18
Created an overload for updateBatch in which a message is not passed …

bgsavage:
…and therefor is left unchanged.

Previously the message was being cleared when updating the status from ACCEPTED to PROCESSED.

So I started down the path of actually trying to track how many groups were created / modified by the group processor.  However, the more I got into it, the more of a rabbit hole it became requiring me to modify common interfaces & models that a lot of places were using.  After discussion with @wtritch , we opted to just keep the existing Accept message as good enough for now, and circle back to having a more detailed later message if this becomes a bigger requirement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.class
 build/
+out/
 
 *.iml
 .idea/
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcStudentGroupBatchRepository.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcStudentGroupBatchRepository.java
@@ -1,9 +1,9 @@
 package org.opentestsystem.rdw.ingest.group.repository;
 
-import com.google.common.collect.ImmutableMap;
 import org.opentestsystem.rdw.ingest.common.model.ImportStatus;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 *.class
 build/
+out/
```
bgsavage: unrelated, Is IntelliJ creating these files in the out directories?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/repository/JdbcStudentGroupBatchRepository.java
```diff
@@ -24,12 +24,17 @@ public JdbcStudentGroupBatchRepository(final NamedParameterJdbcTemplate template
...
         template.update(
                 updateBatch,
-                ImmutableMap.of(
-                        "batch_id", batchId,
-                        "status", status.getValue(),
-                        "message", message));
+                new MapSqlParameterSource()
+                    .addValue("batch_id", batchId)
+                    .addValue("status", status.getValue())
+                    .addValue("message", message));
```
bgsavage: Using `MapSqlParameterSource` now so that `null` can be an acceptable value.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -3,7 +3,7 @@ sql:
   update-batch: >-
     UPDATE upload_student_group_batch
     SET status = :status,
-      message = :message
+      message = coalesce(:message, message)
```
malaffoon: Do you want a separator in there?  Are the messages just text or are they json?

malaffoon: ignore previous comment, i misread this

PR #342, opened 2017-09-22
add to API documentation

malaffoon:
I've combined the API documentation for all the ingest apps. Apart from the diagnostic end-point, really only the import service has any interesting end-points but it seemed easier to combine them all anyway.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,12 +1,16 @@
-## Import Service
+## Ingest Services
+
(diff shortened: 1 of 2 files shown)


malaffoon:
There is a troubleshooting document; currently it is checked in to the private deployment repo because it includes env-specific secrets. I was going to extract that and put a sanitized version in our deployment repo.

PR #345, opened 2017-09-26
tighten down data load role

malaffoon:
After chatting with Will yesterday, i decided to better check the grant role in the import service. Most of the changes involve bean fiddling.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
@@ -19,6 +19,7 @@ import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Sets.newHashSet;
 import static org.opentestsystem.rdw.ingest.auth.SecurityConfigurer.DataLoadAuthority;
+import static org.springframework.util.StringUtils.hasText;
 
 /**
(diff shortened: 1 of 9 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -53,7 +75,8 @@ public OAuth2Authentication extractAuthentication(final Map<String, ?> map) {
             authentication = null;
         } else {
             final TenancyChain tenancyChain = TenancyChain.fromString((String) map.get("sbacTenancyChain"));
-            if (tenancyChain.hasRole(DataLoadAuthority)) {
+            if ((hasText(client) && tenancyChain.hasRoleForClient(DataLoadAuthority, client))
+             || (hasText(state) && tenancyChain.hasRoleForState(DataLoadAuthority, state))) {
```
malaffoon: This is the real change: it's no longer good enough to have ASMTDATALOAD in your tenancy chain, you need to have it for the configured client or state.

import-service/src/main/resources/application.yml
```diff
@@ -23,8 +23,10 @@ cloud:
 archive:
   root: file:///tmp/
 
-# cause StubTokenServices to be used
 security:
+  client: SBAC
+  state: CA
+  # cause StubTokenServices to be used
```
malaffoon: following lead in reporting, default to SBAC and CA.

PR #346, opened 2017-09-27
Feature/support exam upsert

wtritch:
This PR supports "upserting" an Exam based upon the key values: asmt_id, student_id, opp_id.
If those three values match an existing Exam, the existing exam is modified with the submitted exam's values, only updating/removing/creating the records necessary to match the submitted exam.

If any exam or sub-table values are modified, the exam is updated with the given update_import_id in order to participate in the next migration.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -49,17 +55,6 @@ public int getTypeId() {
...
-    /**
-     * This is the only setter because it is available after an object is persisted
-     */
-    public void setStudentId(final int studentId) {
-        this.studentId = studentId;
-    }
-
-    public int getStudentId() {
-        return studentId;
-    }
```
wtritch: This didn't technically belong here, since the student_id is only recorded in the exam_student table, populated by the StudentExamAttributes.  I've moved the property into that model.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -29,7 +32,10 @@
     private Double scaleScoreStdErr;
     private List<Integer> accommodationIds;
     private StudentExamAttributes studentExamAttributes;
-    private Integer studentId;
+
+    public Long getId() {
+        return id;
+    }
```
wtritch: I had to add some "id" properties to the models to facilitate updating.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -119,6 +114,29 @@ public String getSessionId() {
...
+                Objects.equals(completedAt, other.completedAt) &&
+                Objects.equals(asmtVersion, other.asmtVersion) &&
+                Objects.equals(opportunity, other.opportunity) &&
+                Objects.equals(oppId, other.oppId) &&
+                Objects.equals(completenessId, other.completenessId) &&
+                Objects.equals(administrationConditionId, other.administrationConditionId) &&
+                Objects.equals(sessionId, other.sessionId) &&
+                Objects.equals(scaleScore, other.scaleScore) &&
+                Objects.equals(scaleScoreStdErr, other.scaleScoreStdErr);
+    }
```
wtritch: Many of the models have been given an "equivalent" test method to detect if any intrinsic property values are different between two model instances.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamClaim.java
```diff
@@ -1,23 +1,24 @@
...
-        this.claimId = claimId;
-        this.scaleScore = scaleScore;
-        this.scaleScoreStdErr = scaleScoreStdErr;
-        this.category = category;
+
+    private Long id;
+    private int claimId;
+    private Double scaleScore;
+    private Double scaleScoreStdErr;
+    private Integer category;
```
wtritch: switched to a Builder implementation

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
```diff
@@ -9,17 +10,22 @@
  */
 public class ExamItem {
     // the exam item sub-scores db id (aka trait scores), these values must matched the one in the db
-    private final static int EvidenceAndElaboration = 1;
-    private final static int OrganizationAndPurpose = 2;
-    private final static int Conventions = 3;
+    public final static int EvidenceAndElaboration = 1;
+    public final static int OrganizationAndPurpose = 2;
+    public final static int Conventions = 3;
```
wtritch: these had to be made public for access while reading from the database.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -29,7 +32,10 @@
     private Double scaleScoreStdErr;
```
esotrope: Possible to port something like reporting's ScaleScore abstraction here?
It's nice because these props have some cohesion - if one is null the other is null.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -17,9 +18,18 @@
...
+import java.util.stream.Collectors;
 
+import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.ingest.common.repository.JdbcUtils.safeText;
+import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.Conventions;
+import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.EvidenceAndElaboration;
+import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.OrganizationAndPurpose;
 
 @Repository
 class JdbcExamRepository implements ExamRepository {
```
wtritch: This class contains all of the upsert functionality.  To the outside world, the only change is that the single advertised `create(final Exam exam, final long importId)` method was renamed to `upsert(final Exam exam, final long importId)`
This is in keeping with existing implementations that also hide the ugly inside the Repository layer.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
     @Transactional
-    public long create(final Exam exam, final long importId) {
+    public long upsert(final Exam exam, final long importId) {
+        final Exam existing = findOne(exam.getAssessmentId(), exam.getStudentExamAttributes().getStudentId(), exam.getOppId());
+        if (existing == null) {
+            return createExam(exam, importId);
+        } else {
+            return updateExamAndChildren(exam, existing, importId);
+        }
+    }
```
wtritch: If there is an existing exam, update it.  Otherwise, create the exam as before.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
+            updated = true;
+        }
+        if (batchUpsertAccommodations(submitted.getAccommodationIds(), existing.getId())) {
+            updated = true;
+        }
+
+        if (!existing.equivalent(submitted) || updated) {
+            updateExam(submitted, existing.getId(), importId);
+        }
+        return existing.getId();
```
wtritch: This is the root update method.  Child objects are upserted, and if any modifications are made to them OR the exam has updated intrinsic property values then the exam itself is updated.  If the submitted exam is identical to the existing exam then the exam is not modified.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
+                                    .ideaIndicator(row.getBoolean("iep"))
+                                    .lep(row.getBoolean("lep"))
+                                    .section504(row.getBoolean("section504"))
+                                    .economicDisadvantage(row.getBoolean("economic_disadvantage"))
+                                    .migrantStatus(row.getBoolean("migrant_status"))
+                                    .engProfLvl(row.getString("eng_prof_lvl"))
+                                    .t3ProgramType(row.getString("t3_program_type"))
+                                    .languageCode(row.getString("language_code"))
+                                    .primDisabilityType(row.getString("prim_disability_type"))
+                                    .build();
```
wtritch: Since there is only one StudentExamAttributes value per exam, I'm able to fetch its properties at the same time as I fetch the intrinsic Exam properties.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -69,35 +210,130 @@ public long create(final Exam exam, final long importId) {
...
+            }
+        });
+
+        batchUpdateExamClaims(updated);
+        batchCreateExamClaims(newClaims, examId);
+        deleteExamClaims(existingClaimsById.values().stream()
+                .map(ExamClaim::getId)
+                .collect(Collectors.toList()));
+        return !updated.isEmpty() || !newClaims.isEmpty() || !existingClaimsById.isEmpty();
+    }
```
wtritch: This upserts a collection of sub-models that are resolved based upon a primary key.  I'm doing a standard create/update/delete on the existing and submitted collections to make the persisted collection match the submitted collection at the end of this method.

exam-processor/src/main/resources/application.sql.yml
```diff
@@ -70,12 +70,91 @@ sql:
   # JdbcExamRepository
   exam:
```
wtritch: There's a whole lot of new SQL in here to support CRUDing the sub-models.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -234,13 +242,352 @@ public void itShouldCreateExams() {
...
+        final Map<String, Object> values = namedParameterTemplate.queryForMap(
+                "select update_import_id, completeness_id from exam where id = :id",
+                new MapSqlParameterSource().addValue("id", updatedExamId));
+        assertThat(values.get("update_import_id")).isEqualTo(-2L);
+        assertThat(values.get("completeness_id")).isEqualTo(updated.getCompletenessId());
+    }
+
+    @Test
+    @Ignore("Set a breakpoint in JdbcExamRepository#upsertExamStudent to test")
+    public void itShouldNotModifyAnUnchangedExamStudent() {
```
wtritch: By hiding all the functionality behind a single exposed "upsert" method, I can't think of a test that can distinguish between not modifying a sub-model and modifying it to the same property values. Any suggestions?

exam-processor/src/test/resources/SBAC-IAB-FIXED-G4M-G-MATH-4.xml
```diff
@@ -76,7 +76,7 @@
                  reportingVersion="0" abnormalStarts="0" ftCount="0" dateCompleted="2016-10-12T22:00:49.658"
                  assessmentParticipantSessionPlatformUserAgent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36"
                  itemCount="11" pauseCount="0" completeStatus="Complete" status="completed" completeness="Complete"
-                 statusDate="2016-10-12T22:00:49.658" administrationCondition="" startDate="2016-10-12T21:52:34.065" oppId="100149"
+                 statusDate="2016-10-12T22:00:49.658" administrationCondition="" startDate="2016-10-12T21:52:34.065" oppId="100150"
```
wtritch: Soooo, upsert works.  It looks like some of these xml files were copy-pasted with the same asmt_id, student_id, and oppId. =)  I had to change some of the oppId values to avoid breaking or invalidating existing tests.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
wtritch: This change is already larger than I'd like.  I'd prefer to save nice-to-have model refactoring for another PR.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
```diff
@@ -64,6 +70,25 @@ private ExamItemSubScore getByTraitId(final int traitId) {
...
+     * Test for intrinsic-value equivalence with the given StudentExamAttributes.
+     *
+     * @param other A StudentExamAttributes
+     * @return True if the other StudentExamAttributes has identical intrinsic values
+     */
+    public boolean equivalent(final ExamItem other) {
+        if (subScoreList.size() != other.subScoreList.size()) return false;
+        for (final ExamItemSubScore subScore : subScoreList) {
+            if (!other.subScoreList.contains(subScore)) return false;
+        }
```
malaffoon: Hmm, isn't the order of the subScoreList significant? Why isn't this just a deep compare of the two lists? And if the order isn't significant, why is the collection a List?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
+        if (existing == null) {
+            return createExam(exam, importId);
+        } else {
+            return updateExamAndChildren(exam, existing, importId);
+        }
+    }
+
+    private long updateExamAndChildren(final Exam submitted, final Exam existing, final long importId) {
+        boolean updated = upsertExamStudent(submitted.getStudentExamAttributes(), existing.getStudentExamAttributes());
+        if (batchUpsertExamClaims(submitted.getExamClaims(), existing.getId())) {
```
malaffoon: nit: is there a non-batch upsertExamClaims? Just don't think we need to throw the implementation details into the method name.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
     }
 
     @Override
     @Transactional
-    public long create(final Exam exam, final long importId) {
+    public long upsert(final Exam exam, final long importId) {
+        final Exam existing = findOne(exam.getAssessmentId(), exam.getStudentExamAttributes().getStudentId(), exam.getOppId());
+        if (existing == null) {
+            return createExam(exam, importId);
+        } else {
```
agorina: Theoretically this could fail because another thread may have created an exam in between of calls. I am not sure if this is going to happen in practice, but I do not think it is hard to solve it. You just need to catch an exception (not sure which one though) and if it happens execute update. 

You can find the error code raised in the student upsert stored procedure.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -29,36 +39,167 @@
...
+        boolean updated = upsertExamStudent(submitted.getStudentExamAttributes(), existing.getStudentExamAttributes());
+        if (batchUpsertExamClaims(submitted.getExamClaims(), existing.getId())) {
+            updated = true;
+        }
+        if (batchUpsertExamItems(submitted.getExamItems(), existing.getId())) {
+            updated = true;
+        }
+        if (batchUpsertAccommodations(submitted.getAccommodationIds(), existing.getId())) {
+            updated = true;
+        }
```
malaffoon: nit: it would look prettier if you just inlined them all?
```
final boolean updated = upsertExamStudent(...)
    || upsertExamClaims(...)
    || upsertExamItems(...)
    || upsertAccommodations(...);
```
or ...
```
    boolean updated = upsertExamStudent(...);
    updated |= upsertExamClaims(...);
    updated |= upsertExamItems(...);
    updated |= upsertAccommodations(...);

File: exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java [hunk 19]
```diff
@@ -69,35 +210,130 @@ public long create(final Exam exam, final long importId) {
...
+                ));
+
+        final Map<Long, ExamClaim> updated = new HashMap<>();
+        final List<ExamClaim> newClaims = new ArrayList<>();
+
+        submittedClaims.forEach(claim -> {
+            final ExamClaim existing = existingClaimsById.get(claim.getClaimId());
+            if (existing == null) {
+                newClaims.add(claim);
+                return;
```
malaffoon: i have to say the return in the middle of this lambda is confusing; only suggestion i have is to have an else block which might not be much better

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -69,35 +210,130 @@ public long create(final Exam exam, final long importId) {
...
+
+            existingClaimsById.remove(claim.getClaimId());
+            if (!existing.equivalent(claim)) {
+                updated.put(existing.getId(), claim);
+            }
+        });
+
+        batchUpdateExamClaims(updated);
+        batchCreateExamClaims(newClaims, examId);
+        deleteExamClaims(existingClaimsById.values().stream()
```
malaffoon: couldn't this just be `deleteExamClaims(existingClaimsById.keySet())`?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -107,10 +343,85 @@ private long create(final StudentExamAttributes studentExamAttributes, final lon
...
-                .addValue("prim_disability_type", studentExamAttributes.getPrimDisabilityType());
+                .addValue("prim_disability_type", studentExamAttributes.getPrimDisabilityType())
+                .getValues();
+    }
 
-        jdbcTemplate.update(sqlExamStudentCreate, parameterSource, keyHolder);
-        return keyHolder.getKey().longValue();
+    private boolean batchUpsertExamItems(final List<ExamItem> submittedItems, final long examId) {
+        final Map<Integer, ExamItem> existingItemsByItemId = findExamItemsByExamId(examId).stream()
+                .collect(Collectors.toMap(
```
malaffoon: a couple dittos here ...

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
malaffoon: put a spy on the jdbcTemplate?

wtritch: Unfortunately, spying the jdbcTemplate will only affect the property inside the test, it won't give me access to the instance that's already been injected into the @Autowired ExamRepository =(

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
wtritch: I don't believe the order is significant.  It looks like it could be changed to a Set without any loss of functionality.  Think I should roll that into this PR?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: hmm... good point.  Think I should remove "batch" from all of the method names?  (There's only one "Create Exam Claims" method named batchCreateExamClaims as well.)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: I considered the first, but I believe Java is smart enough to short-circuit an OR-chain on the first True value it finds: `boolean myBool = true || someMethod()` shouldn't execute someMethod()
I'll try the second and see if it will short-circuit that as well.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
malaffoon: Hmm. Although the order is not significant (at least i can find no evidence of that in any documentation), the XML does represent this with an ordered sequence. Personally, i'd just do a deep compare on the two lists: if they submit the exact same exam item XML with those sub scores in a different order then, yeah, it would count as an update. But that isn't going to happen.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: Unfortunately it won't fail, since we don't have a UNIQUE index on our proposed key values.  Which record will be updated with future submissions is indeterminate.
I discussed the idea of adding a UNIQUE index on our proposed key values to @malaffoon but he was worried we may already have duplicates and adding that unique index may blow up in production.  Thoughts?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
malaffoon: Doh, good point. The second approach will work fine though.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
malaffoon: I think we should discuss adding the constraint. That wasn't exactly my concern (although it is a valid one).

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: ahh, the |= impl executes the method regardless of the current value of "updated" and looks a lot cleaner.  Thank you!

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: hmm... maybe I'm just old-skool, but it looks cleaner as a standard `for claim in submittedClaims` loop with a "continue;" rather than a forEach with a lambda and "return;"

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: I suppose it could be, but I'm a purist and would prefer to delete by primary key rather than claimId...  Even though there is an assumption that claimId is unique per examId

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: Same for loop change here, same preference to delete by primary key, even though we assume itemId is unique per examId

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
wtritch: hmm... I'm a little iffy about that.  I'd have to make sure that the "read" operation from the DB loads the list in the same assumed order as the XML.  (Since the "list" is persisted as a set of 6 columns in the exam_item table.)

wtritch:
Merging for now.  Fixing the timing issue between update and create will require an additional PR and work to merge the exam_student table into the exam table.

PR #348, opened 2017-09-28
Handle duplicate exam insertion

wtritch:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -84,6 +86,7 @@
     }
 
     @Override
+    @Retryable(maxAttemptsExpression = "${jdbc.retry.default.max-attempts}", backoff = @Backoff(delayExpression = "${jdbc.retry.default.delay}"))
```
wtritch: This is the big change: Make the upsert retry-able to handle DuplicateKeyExceptions thrown if we attempt to create the same exam multiple times at the "same time" in a multi-threaded context.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/StudentExamWriter.java
```diff
@@ -16,5 +16,5 @@
      * @param student  the {@link Student}
      * @param importId the import id associated with the exam
      */
-    void saveNewExam(Student student, Exam exam, long importId);
+    void upsertExam(Student student, Exam exam, long importId);
```
wtritch: unrelated, just some naming for accuracy.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
jtreuting: What's the need for having this here where we don't have it in most places?

wtritch: We have @Retryable on a few of our methods, although I haven't done a deep-dive on exactly why they are where they are.  It's basically a way to handle transitory failure due to race conditions/optimistic-locking where one request wins and another fails.

bgsavage: I heard there's a @WillRetry annotation in Spring 5 which implements the exponential backoff.

wtritch: I do like exponential back-off... but not here =P

malaffoon: No default values for the properties?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/StudentExamWriter.java
malaffoon: Is it upserting the student too?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
wtritch: Looks like they're set in the application.yml:
```
jdbc:
  retry:
    default:
      max-attempts: 4
      delay: 30
```

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/StudentExamWriter.java
wtritch: looks like it *is* upserting the student as well.  Will change to "upsert"

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
chdietz: Is the use case here an exam and an update to the same exam potentially being processed at the same time by multiple exam processors?

chdietz: I see this in exam-processor and I think it is there for the race condition creating/updating a student or student group based on an exam

malaffoon: I'm fine with this approach but i will say that if i were doing it i probably would've caught a particular SQL exception so it wasn't retrying on random failures. Odds of that are pretty darn low though.

malaffoon: oh okay then; didn't see it in the review so i didn't think it would be in there ... thanks

wtritch: There is, but they're already covered via unique indexes and this same technique

wtritch: I'd agree that the odds are pretty low.  Also, this is the same technique used in other race conditions, such as upserting students.  So if we want to discuss an alternate approach it would be a larger discussion. =)

PR #350, opened 2017-10-03
Add a DELETE API to the import-service exams endpoints

wtritch:
This PR adds a DELETE `/exams/imports/{importId}' endpoint to the ExamController in the import-service.  The endpoint verifies the existence of a RDWImport with content EXAM and the given Id, then submits a message with the importId to the new EXAM_DELETE channel.
This will be followed up by a PR in the exam-processor that is responsible for soft-deleting the Exam identified by the importId.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
@@ -16,6 +16,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Properties;
 
@@ -137,6 +138,15 @@ class DefaultImportService implements ImportService {
             return Optional.empty();
         }
         return Optional.of(archiveService.readResource(location(rdwImport)));
+    }
+
+    @Override
+    public void deleteImport(final long importId, final ImportContent content) {
+        final RdwImport rdwImport = repository.findOne(importId);
+        if (rdwImport == null || rdwImport.getContent() != content) {
+            throw new NoSuchElementException("Import: " + importId + " Type: " + content.name() + " not found.");
+        }
+        source.submitDelete(importId, content);
     }
 
     private static String location(final RdwImport rdwImport) {
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
@@ -141,12 +141,20 @@ class DefaultImportService implements ImportService {
     }
 
     @Override
-    public void deleteImport(final long importId, final ImportContent content) {
-        final RdwImport rdwImport = repository.findOne(importId);
-        if (rdwImport == null || rdwImport.getContent() != content) {
+    public void deleteImport(final SbacUserDetails user,
+                             final long importId,
+                             final ImportContent content) {
+        final RdwImport importToDelete = repository.findOne(importId);
+        if (importToDelete == null || importToDelete.getContent() != content) {
             throw new NoSuchElementException("Import: " + importId + " Type: " + content.name() + " not found.");
         }
-        source.submitDelete(importId, content);
+
+        final RdwImport deleteImport = repository.create(RdwImport.builder()
+                .content(content)
+                .status(ImportStatus.ACCEPTED)
+                .creator(user.getUsername())
+                .build());
+        source.submitDelete(deleteImport.getId(), importToDelete.getId(), content);
     }
 
     private static String location(final RdwImport rdwImport) {
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
@@ -32,6 +32,7 @@ class DefaultImportService implements ImportService {
     private static final String ContentTypeProperty = "Content-Type";
     private static final String UsernameProperty = "username";
     private static final String TenancyChainProperty = "tenancy-chain";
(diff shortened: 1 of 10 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -139,6 +140,15 @@ public long resubmitImports(final RdwImportQuery query) {
...
     }
 
+    @Override
+    public void deleteImport(final long importId, final ImportContent content) {
+        final RdwImport rdwImport = repository.findOne(importId);
+        if (rdwImport == null || rdwImport.getContent() != content) {
+            throw new NoSuchElementException("Import: " + importId + " Type: " + content.name() + " not found.");
+        }
+        source.submitDelete(importId, content);
+    }
```
wtritch: Check that an EXAM import with the given importId actually exists.  (I noticed there weren't any permission/ownership checks on the GET endpoint so I wasn't sure if we should/could filter on permissions here?)

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportSource.java
```diff
@@ -57,6 +61,16 @@ public void submitContent(final byte[] body,
...
 
+    @Override
+    public void submitDelete(final long importId, final ImportContent content) {
+        final RdwMessageHeaderAccessor accessor = wrap(null)
+                .setReceivedNow()
+                .setContent(content.toString())
+                .setImportId(importId);
+        accessor.setHeader(OperationHeader, OperationDelete);
+        outputChannel.send(MessageBuilder.createMessage(importId, accessor.getMessageHeaders()));
+    }
```
wtritch: Add an "X-Operation: DELETE" header to the message and submit it to the same output channel.  The router will make sure it gets to the EXAM_DELETE channel.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportSource.java
```diff
@@ -69,7 +83,8 @@ public AbstractMappingMessageRouter router() {
             @Override
             protected List<Object> getChannelKeys(final Message<?> message) {
                 final RdwMessageHeaderAccessor accessor = wrap(message);
-                return Collections.singletonList(accessor.getContent());
+                final String operation = (String) accessor.getHeader(OperationHeader);
+                return Collections.singletonList(ChannelJoiner.join(accessor.getContent(), operation));
```
wtritch: Route based upon the value/absence of an "X-Operation" message header.

import-service/src/test/java/org/opentestsystem/rdw/ingest/service/DefaultImportSourceTest.java
```diff
@@ -43,5 +56,26 @@ public void submitExamShouldSetHeadersAndSendMessage() {
...
+        final ArgumentCaptor<Message> messageArgumentCaptor = ArgumentCaptor.forClass(Message.class);
+        verify(channel).send(messageArgumentCaptor.capture());
+        final Message message = messageArgumentCaptor.getValue();
+        final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
+        assertThat(message.getPayload()).isEqualTo(importId);
+        assertThat(accessor.getHeader(OperationHeader)).isEqualTo(OperationDelete);
+        assertThat(accessor.getImportId()).isEqualTo(importId);
+
+        importSource.router().handleMessage(message);
+        verify(channelResolver).resolveDestination(EXAM.name() + "_" + OperationDelete);
```
wtritch: I added some router tests in here to validate that we're sending the payload where we think we're sending it =)

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
malaffoon: Unless something has changed, every end-point is protected requiring ASMTDATALOAD at the client or state level.

wtritch:
I won't merge until both parts are GTG =P

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -139,6 +140,23 @@ public long resubmitImports(final RdwImportQuery query) {
...
+        if (importToDelete == null || importToDelete.getContent() != content) {
+            throw new NoSuchElementException("Import: " + importId + " Type: " + content.name() + " not found.");
+        }
+
+        final RdwImport deleteImport = repository.create(RdwImport.builder()
+                .content(content)
+                .status(ImportStatus.ACCEPTED)
+                .creator(user.getUsername())
+                .build());
+        source.submitDelete(deleteImport.getId(), importToDelete.getId(), content);
```
wtritch: I had to update this to create an additional Import to bind the deletion to.  Otherwise the deletion would never be migrated.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportSource.java
```diff
@@ -57,6 +61,16 @@ public void submitContent(final byte[] body,
...
     }
 
+    @Override
+    public void submitDelete(final long importId, final long deletedImportId, final ImportContent content) {
+        final RdwMessageHeaderAccessor accessor = wrap(null)
+                .setReceivedNow()
+                .setContent(content.toString())
+                .setImportId(importId);
+        accessor.setHeader(OperationHeader, OperationDelete);
+        outputChannel.send(MessageBuilder.createMessage(deletedImportId, accessor.getMessageHeaders()));
```
wtritch: The message now references two importIds.  the RdwMessageHeaderAccessor.getImportId() references the new Import that will be used to perform the deletion, and the payload contains the importId to be deleted.

wtritch:
Abandoning in favor of #366

PR #352, opened 2017-10-04
Feature/delete exam by import

wtritch:
This PR is the second half of deleting exams by import id.
See PR #350 for the API endpoint that publishes a delete message to the EXAM_DELETE channel.

This PR accepts the EXAM_DELETE message, updates the exam.deleted column to 1, and marks the processing Import record as PROCESSED.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -96,7 +96,7 @@ public class ExamProcessorConfiguration {
 
     private void processDelete(final long importId, final long deletedImportId) {
         try {
-            reportProcessor.processDelete(importId, deletedImportId);
+            reportProcessor.processDelete(deletedImportId, importId);
             importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
 
         } catch (final ImportException e) {
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -1,11 +1,12 @@
 package org.opentestsystem.rdw.ingest.processor;
 
+import org.opentestsystem.rdw.common.model.trt.TDSReport;
 import org.opentestsystem.rdw.ingest.common.model.ImportException;
 import org.opentestsystem.rdw.ingest.common.model.ImportStatus;
 import org.opentestsystem.rdw.ingest.common.repository.ImportRepository;
 import org.opentestsystem.rdw.ingest.processor.service.TDSReportProcessor;
+import org.opentestsystem.rdw.ingest.processor.stream.ExamProcessorSink;
 import org.opentestsystem.rdw.messaging.RdwMessageHeaderAccessor;
-import org.opentestsystem.rdw.common.model.trt.TDSReport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -19,7 +20,7 @@ import java.io.ByteArrayInputStream;
 import static org.opentestsystem.rdw.common.model.trt.XmlUtils.tdsReportFromXml;
 
 
-@EnableBinding(Sink.class)
+@EnableBinding(ExamProcessorSink.class)
 public class ExamProcessorConfiguration {
     private static final Logger logger = LoggerFactory.getLogger(ExamProcessorConfiguration.class);
 
@@ -34,12 +35,11 @@ public class ExamProcessorConfiguration {
     }
 
     @ServiceActivator(inputChannel = Sink.INPUT)
-    public void process(final Message<?> message) {
+    public void processUpsert(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         final byte[] payload = (byte[]) message.getPayload();
         final long importId = accessor.getImportId();
-
-        logger.debug("received {} import {}", accessor.getContent(), importId);
+        logger.debug("received {} upsert, import {}", accessor.getContent(), importId);
 
         try {
             final TDSReport report = toTDSReport(new ByteArrayInputStream(payload), importId);
@@ -48,16 +48,31 @@ public class ExamProcessorConfiguration {
                 return;
             }
 
-            process(report, importId);
+            processUpsert(report, importId);
         } catch (final Exception e) {
             logger.warn("failed with an unexpected import error: " + e.getMessage());
             importRepository.updateStatusAndMessageById(importId, ImportStatus.INVALID, e.getMessage());
         }
     }
 
-    private void process(final TDSReport tdsReport, final long importId) {
+    @ServiceActivator(inputChannel = ExamProcessorSink.DELETE)
+    public void processDelete(final Message<?> message) {
+        final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
+        final long deletedImportId = (Long) message.getPayload();
+        final long importId = accessor.getImportId();
+        logger.debug("received {} delete, import {}", accessor.getContent(), importId);
+
         try {
-            reportProcessor.process(tdsReport, importId);
+            processDelete(importId, deletedImportId);
+        } catch (final Exception e) {
+            logger.warn("failed with an unexpected import error: " + e.getMessage());
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.INVALID, e.getMessage());
+        }
+    }
+
+    private void processUpsert(final TDSReport tdsReport, final long importId) {
+        try {
+            reportProcessor.processUpsert(tdsReport, importId);
             importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
 
         } catch (final ImportException e) {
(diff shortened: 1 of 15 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -34,12 +35,11 @@ public ExamProcessorConfiguration(final TDSReportProcessor reportProcessor,
     }
 
     @ServiceActivator(inputChannel = Sink.INPUT)
```
wtritch: I left the upsert binding as "input" to avoid running into configuration issues by switching binding names.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -48,16 +48,31 @@ public void process(final Message<?> message) {
...
-            process(report, importId);
+            processUpsert(report, importId);
+        } catch (final Exception e) {
+            logger.warn("failed with an unexpected import error: " + e.getMessage());
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.INVALID, e.getMessage());
+        }
+    }
+
+    @ServiceActivator(inputChannel = ExamProcessorSink.DELETE)
+    public void processDelete(final Message<?> message) {
```
wtritch: This is the new delete channel binding.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -78,4 +93,18 @@ private TDSReport toTDSReport(final ByteArrayInputStream xml, final long importI
...
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
+
+        } catch (final ImportException e) {
+            logger.info("failed with an import error: " + e.getMessage());
+            importRepository.updateStatusAndMessageById(importId, e.getStatus(), e.getMessage());
+        } catch (final Exception e) {
+            logger.warn("failed with an unexpected import error: " + e.getMessage());
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
+        }
+    }
```
wtritch: The delete process follows the same path as the upsert process:
configuration -> processor -> writer -> repository

exam-processor/src/main/resources/application.sql.yml
```diff
@@ -101,7 +101,8 @@ sql:
         eng_prof_lvl = :eng_prof_lvl,
         t3_program_type = :t3_program_type,
         language_code = :language_code,
-        prim_disability_type = :prim_disability_type
+        prim_disability_type = :prim_disability_type,
+        deleted = 0
```
wtritch: Now, when updating an exam, we need to be sure to "undelete" it.

exam-processor/src/main/resources/application.sql.yml
```diff
@@ -137,6 +138,14 @@ sql:
         AND oppId = :opportunity_id
         AND student_id = :student_id
 
+    deleteByImportId: >-
+      UPDATE exam
+      SET
+        deleted = 1,
+        update_import_id = :import_id
+      WHERE
+        import_id = :deleted_import_id
```
wtritch: Note that I'm only deleting by the original creation import id, not by the update import id.  I waffled on this for a bit but decided that I wanted a consistent way to identify an exam.  If we also identified exams by update import id, then a deletion request that's valid *right now* may not be in a second, or a day, or a month, etc.

exam-processor/src/main/resources/application.sql.yml
agorina: This raises a question: can somebody update a deleted exam? 
Is this a valid use case?

exam-processor/src/main/resources/application.sql.yml
agorina: Hmmm..this is an interesting point. I am afraid that from the end user point of view the `delete` feature is hard to use. Maybe they would never use it.

The more I look into it, the more it seems that we should be deleting by the same composite key that we are using to identify update. I know we have talked about it, but should we support an explicit delete request with TRT? Would it be more user friendly approach?

exam-processor/src/main/resources/application.sql.yml
wtritch: Considering that all they're doing here is submitting a TRT, I don't think so.  I'm operating under the assumption that desired behavior is:
Submitting a TRT leads to an Exam record being created/updated.

exam-processor/src/main/resources/application.sql.yml
wtritch: I don't think either are user-friendly.  From what I understand the TRTs are just used in a flow.  I don't know that they end up anywhere an admin can grab and re-submit them with a "?delete=true" kind of flag.  But I have to admit I'm fuzzy on the whole TRT handling process.
If we want user-friendliness, I keep going back to needing a separate business-object-aware application/API.  Where a user can query for and work on exams, or students, or exam items, or etc, etc.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/stream/ExamProcessorSink.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.ingest.processor.stream;
+
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.messaging.SubscribableChannel;
+
+/**
+ *
```
malaffoon: Well this certainly illuminates things  ;)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/stream/ExamProcessorSink.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface ExamProcessorSink {
+
+    String DELETE = "delete";
```
malaffoon: final?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/stream/ExamProcessorSink.java
wtritch: oops

malaffoon:
I hate to say this (please don't hit me @wtritch) but ...

I think we should *not* make this change (nor the other related PR). We don't know the use-case / workflow for deletes. And, until we understand the workflow, we shouldn't be introducing special delete behavior. Especially since this delete behavior may be problematic or difficult to use (how do i get the import-id for a TRT i'm looking at). 

If we had a RESTful API for managing exams, etc. then we could just throw the DELETE method out there and be ready for the future. But what we've done here is a bit different, so there is no standard approach to fall back on. 

We can discuss it as a group and try to figure out a good approach (presumably by guessing how it is going to be used). Or we can push back on the requirements and ask, "exactly how will this be used?" Will this be AIR referencing a TRT file and asking that it be "revoked"? Is this an administrator wanting to delete a result based on the student id, grade, asmt id, and date? Is this ETS providing the asmt id, opp id, student id from their database?

FYI, @agorina, @wtritch, @jtreuting

exam-processor/src/main/resources/application.sql.yml
chdietz: I think I chatted or slacked with Will about this same thing.  I think it is reasonable that we accept a TRT via a post to /exams/imports and the response includes the import_id.  Any actions they want to take on that post include the import_id we gave them.

At the same time this feels different than how we perform an update.
But the use case for delete sounds more like - "I did not mean to submit this {import_id} please remove it"

wtritch:
I'm tempted to agree with you @malaffoon.  After hearing multiple concerns, alternative implementations, etc from multiple people I don't feel like we have this requirement well-defined enough to implement functionality.  I'll leave the PRs up overnight and barring any new information will:
1. merge this branch into the `feature/add-delete-exam-import-api`
2. close the PR from `feature/add-delete-exam-import-api` -> `develop`
3. will leave the branch alive until we can get a clear line of sight on use cases and well-defined requirements

agorina:
@wtritch, if you really need to be mad at somebody about it: I admit being the one who instigate this discussion. Sorry...but I think we should either do it right or not do it at all.

exam-processor/src/main/resources/application.sql.yml
chdietz: So delete, even though based on import_id should not be used to revert exam updates?
example:
* New TRT posted and exam created.
* Problem suspected in scoring, adjusted score TRT created and posted, exam is updated.
* Subsequent determination the update was not required and original posted TRT is the desired state.

What would be the remedy in this case?
* can't delete the update.
* can not re submit the original TRT because of duplicate digest.
* would have to be original TRT but with something changed so there is a new digest?

exam-processor/src/main/resources/application.sql.yml
chdietz: I would say no.  It's like can I modify a deleted file.
What if i want to revert a delete?
Would I have to re submit the TRT, but then the duplicate digest issue I already mentioned.

wtritch:
@agorina I'm not mad, it was a fun little exercise.  I think you're right that if we don't know who's going to use a feature, why they're going to use it, or how they're going to use it, then we have no business building it.  I thought this was a simple "oops" import reversion to fast-fix a quickly-discovered error.  If the requirement is instead more of a data management tool, this isn't it.  And if we don't even know which it is... we shouldn't just be guessing one =P

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -535,4 +535,20 @@ public void itShouldRemoveStaleAccommodationIds() {
...
+        repository.deleteByImportId(-1, -2);
+
+        final Map<String, Object> updatedValues = namedParameterTemplate.queryForMap(
+                "select * from exam where id = :id",
+                new MapSqlParameterSource().addValue("id", examId));
+
+        assertThat(updatedValues.get("deleted")).isEqualTo(1);
+        assertThat(updatedValues.get("update_import_id")).isEqualTo(-2L);
+    }
 }
```
chdietz: Should we have a test that deletes an updated exam?

exam-processor/src/main/resources/application.sql.yml
wtritch: Nope, I think delete should be delete.  If you want to revert to an earlier TRT, you should re-submit the "correct" TRT.  Something @malaffoon brought up w.r.t. exam upsert is that we may want to store and update based upon the latest Opportunity.statusDate value, latest one wins.

chdietz:
I agree with putting on hold and glad @wtritch thought it was fun.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
wtritch: eh, I could be talked into it if you feel strongly, but it wouldn't test any different functionality/branch cases.

wtritch:
Abandoning in favor of #366

PR #353, opened 2017-10-04
only build docker image if module needs it

malaffoon:


common/gradle.properties
```diff
@@ -0,0 +1 @@
+skipDocker=true
```
wtritch: hmmm... negative assertions vs positive assertions...

malaffoon: Yeah, thought about that. But the general case is to do Docker, so this is fewer files to change.

PR #354, opened 2017-10-05
DWR-1060 Migrate OLAP service - first draft

agorina:
This is not intended to be runnable yet.  And I hope it will not break anything.

There are multiple things missing: proper gradle config and CI/ Integration tests settings, config service, docker and so on.  But I am able to run it locally. It runs the full migrate cycle for students. From the aws dev Aurora to Redshift it processed all students (about 700K) while working through 25 mil imports within about 25 min.


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/status/MigrateJobStatusIndicator.java
```diff
@@ -0,0 +1,36 @@
+package org.opentestsystem.rdw.migrate.common.status;
```
agorina: Moved more classes to the migrate-common.

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/status/MigrateJobStatusIndicatorTest.java
```diff
@@ -0,0 +1,52 @@
+package org.opentestsystem.rdw.migrate.common.status;
```
agorina: This was moved to the migrate-common too.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+                                   final Step truncateAfterStageStep) {
+        try {
+
+            return jobBuilderFactory.get("Migrate OlAP Reporting Job")
+                    .preventRestart()
+                    .listener(listener())
+                    .start(cleanupLandingLocationStep)
+                    .next(truncateBeforeStageStep)
+                    .next(extractCodesStep)
+                    .next(extractEntitiesStep)
```
agorina: @malaffoon and @jtreuting , after I convinced myself and you to break the entities migrate into two steps (inserts vs. updates) , I ended up using one step and  DISTINCT and it seems to be just fine. Will need more testing with the EXAMs.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/CleanupLandingLocation.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Autowired
+    public CleanupLandingLocation(final ArchiveService archiveService) {
+        this.archiveService = archiveService;
+    }
+
+    @Override
+    public RepeatStatus execute(final StepContribution contribution, final ChunkContext chunkContext) throws Exception {
+        try {
+            archiveService.delete(location);
```
agorina: There are a few problems here:
1. While running the migrate from dev aurora I had about 4-5 exceptions here: `org.apache.http.NoHttpResponseException: The target server failed to respond` and `org.apache.http.conn.ConnectTimeoutException: Connect to rdw-dev-archive...`.
Quick Goggling found this: https://stackoverflow.com/questions/23912407/frequent-nohttpresponseexception-with-amazons3-getobjectrequest-getobjectconte.
I will look into it more.

2. The job did not stop the execution. And this is not right. I think it is because it stops only on the `UncheckedTransactionException`. Once again - need more research.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -0,0 +1,52 @@
...
+    entities:
+        # ------------ gender  -------------------------------------------------------------------
+        gender:
+          sql:
+            update:  >-
+               UPDATE gender
+               SET
+                 code = sg.code
+               FROM
+                  staging_gender sg where sg.id = gender.id;
```
agorina: It is interesting (or annoying) how the syntax is different from MySQL. Here is how it looks in the MySQL:
` UPDATE gender rg
                JOIN staging_gender sg ON sg.id = rg.id
              SET
                rg.code = sg.code;`

And there multiple details like this.

migrate-olap/src/main/resources/application.yml
```diff
@@ -0,0 +1,104 @@
...
+  port: 8008
+  tomcat:
+    max-connections: 2
+    max-threads: 2
+
+migrate:
+  aws:
+    location: MigrateOlap
+    redshift:
+      role: ${rdw.migrate.aws.redshift.role}
```
agorina: This will need to be defined.

migrate-olap/gradle.properties
```diff
@@ -0,0 +1 @@
+skipDocker=true
```
agorina: I am excluding it for now.

migrate-olap/src/main/resources/application.yml
malaffoon: I don't think that roles are considered sensitive data. So i don't think we need the abstraction of another property. @wtritch thoughts?

migrate-olap/src/main/resources/application.yml
```diff
@@ -0,0 +1,104 @@
...
+    testWhileIdle: true
+    validationQuery: SELECT 1
+    driverClassName: com.mysql.jdbc.Driver
+    initialize: false
+    maxActive: 10
+    initialSize: 4
+
+  reporting_datasource:
+    url: "\
+      jdbc:redshift://${spring.reporting_olap_datasource.url-server:localhost:3306?${spring.reporting_olap_datasource.url-schema:reporting_olap}
```
malaffoon: This isn't right: the placeholder properties should have the same scope. Instead of `spring.reporting_olap_datasource...` it should be `spring.reporting_datasource...`.

agorina: Oh, ok. Will change.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
malaffoon: Shoulda done everything with MongoDB ... it's web-scale!

migrate-olap/src/main/resources/bootstrap.yml
```diff
@@ -0,0 +1,17 @@
+# Spring Cloud applications have a bootstrap context so certain properties
+# should be defined here and not in application.yml/properties.
+# (https://github.com/spring-cloud/spring-cloud-commons/blob/master/docs/src/main/asciidoc/spring-cloud-commons.adoc#the-bootstrap-application-context)
+spring:
+  application:
+    name: rdw-ingest-migrate-olap
```
malaffoon: I have emailed Alex to get the docker hub repo created for this. He is out of the office today but we can hope that he creates it in the next few days.

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/status/MigrateJobStatusIndicatorTest.java
malaffoon: I would prefer creating a concrete instance of the status indicator and test the common functionality all within migrate-common as a real unit test. If the extensions add additional functionality, they should be testing just that.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapReportingJobStatusIndicator.java
```diff
@@ -0,0 +1,24 @@
...
+import org.opentestsystem.rdw.migrate.common.MigrateLifecycle;
+import org.opentestsystem.rdw.migrate.common.status.MigrateJobStatusIndicator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * A diagnostic {@link StatusIndicator} for the migrate job component.
+ */
+@Component
+public class MigrateOlapReportingJobStatusIndicator extends MigrateJobStatusIndicator {
```
malaffoon: (continuing comment from above somewhere...)

The unit test for this class should be testing *only* the name() method.

agorina: Which it does, so I think I am confused. Let's chat.

PR #355, opened 2017-10-05
DWR-1060 Migrate Olap - added student ethnicity

agorina:


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -105,4 +105,47 @@ sql:
...
+        # ------------ student_ethnicity  -------------------------------------------------------------------
+        student_ethnicity:
+          sql:
+            warehouseRead: >-
+              SELECT
+                ethnicity_id,
+                student_id
+              FROM student_ethnicity
+                JOIN student ws ON ws.id = student_ethnicity.student_id
+              WHERE ? IS NOT NULL AND ws.created > ? AND ws.created <= ? AND ws.deleted = 0
```
agorina: I would like to get some input on this: 

For some tables in redshift (master ones) I am including migrate_id, but others do not have it. There is a repo that takes Migrate as an input it uses its id, firstAt and lastAt as parameters to the SQL. To work around it I am including migrate id as '? IS NOT NULL'. Too ugly?

agorina: Talked offline - will add migrate_id to the staging table to be consistent.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -105,4 +105,51 @@ sql:
               FROM '${archive.root}/${migrate.aws.location}/student_update.part_00000'
               CREDENTIALS 'aws_iam_role=${migrate.aws.redshift.role}'
               FORMAT AS CSV
```
malaffoon: I know we're controlling the reading/writing of the CSV but what if a student name (or some other field) has a comma in it? Are we escaping things with double quotes?

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -105,4 +105,51 @@ sql:
...
+          sql:
+            warehouseRead: >-
+              SELECT
+                ethnicity_id,
+                student_id,
+                ? as migrate_id
+              FROM student_ethnicity
+                JOIN student ws ON ws.id = student_ethnicity.student_id
+              WHERE ws.created > ? AND ws.created <= ? AND ws.deleted = 0
+                INTO OUTFILE S3 '${archive.root}/${migrate.aws.location}/student_ethnicity'
```
bgsavage: Won't this fail on a second run?  Unless you have something coming behind you to clean up the file.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: I have it on my 'TODO' list to test this types of things. But thank you for reminding me.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: Yes, I clean up the file.

bgsavage: Any potential for a race condition here?

agorina: Both of our migrates (the reporting and this one) are expected to run as one threaded instance. There are multiple issues with making them scalable. If for some reason two instances are started, they will collide and cause migrate to fail. Once it fails it stays failed until the issues is resolved. This is our crude method of preventing it from being accidently scaled.

PR #357, opened 2017-10-09
Feature/dwr 1060 migrate olap

agorina:
Add asmt, school and district.

Note that I am only including ICA asmts, and do not include school/district groups. I want to discuss this with the team. I feel that until we have the requirements in place it is premature to know what we need, and for now I am doing the minimum based on the sample queries we have built so far. As long as we have the framework set up adding more should not be a big task.  Please let me know if you disagree.

### DIFF
--- a/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
@@ -168,3 +168,87 @@ sql:
               DELETE FROM student_ethnicity
                 USING staging_student ss
               WHERE ss.id = student_ethnicity.student_id AND ss.deleted = 1;
+
+      # ------------ asmt -----------------------------------------------------------------------
+        asmt:
+          sql:
+            insert: >-
+              INSERT INTO ica_asmt (id, grade_id, school_year, subject_id, name, migrate_id)
+                SELECT DISTINCT
+                  sa.id,
+                  sa.grade_id,
+                  sa.school_year,
+                  sa.subject_id,
+                  sa.name,
+                  sa.migrate_id
+                FROM staging_asmt sa
+                  LEFT JOIN ica_asmt ra ON ra.id = sa.id
+                WHERE ra.id IS NULL and sa.deleted = 0 AND sa.type_id = 1;
+
+            update: >-
+              UPDATE ica_asmt
+              SET
+                grade_id  = sa.grade_id,
+                school_year = sa.school_year,
+                subject_id = sa.subject_id,
+                name = sa.name,
+                migrate_id = sa.migrate_id
+              FROM staging_asmt sa WHERE sa.id = ica_asmt.id AND sa.type_id = 1;
+
+            delete: >-
+              DELETE FROM ica_asmt
+                USING staging_asmt sa
+              WHERE sa.id = ica_asmt.id AND sa.deleted = 1 AND sa.type_id = 1;
+
+      # ------------ school -----------------------------------------------------------------------
+        school:
+          sql:
+            insert: >-
+              INSERT INTO school (id, district_id, name, migrate_id)
+                SELECT DISTINCT
+                  ss.id,
+                  ss.district_id,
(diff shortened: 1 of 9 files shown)


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/ReportingMigrateJobIT.java
```diff
@@ -49,10 +49,7 @@ public void itShouldUpsertAndDeleteCodesIfBatchHasImportContentWithCodes() throw
         // collect counts of rows in each reporting code table before the step call
         final List<TableTestCountHelper> tableTestCounts = newArrayList();
         for (final String table : codesEntities) {
-            if ("administration_condition".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(reportingJdbcTemplate, table, -1, "id IN (-99)"));
-                tableTestCounts.add(new TableTestCountHelper(reportingJdbcTemplate, table, 1, "id IN (-98)"));
-            } else if ("math_practice".equals(table)) {
```
agorina: Unrelated clean up. I noticed that it seems to be a left over from the time when this table did not have an `id` column.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
```diff
@@ -4,24 +4,23 @@
...
 
     @Autowired
     private WarehouseToStageRepository repository;
 
     @Autowired
-    private JdbcTemplate warehouseJdbcTemplate;
+    private NamedParameterJdbcTemplate warehouseJdbcTemplate;
 
     @Autowired
-    private JdbcTemplate reportingJdbcTemplate;
```
agorina: This was injecting the same object. And it was pointing to the reporting DB. The only way it worked is because tables exist in both.

chdietz:
Only doing ICA first seems reasonable.  But why delay on groups of schools, are you waiting to make sure this is required?

agorina:
My main reason is that I am not sure how to correctly structure school/district groups in the star schema: do we do a 'star' or 'snowflake' schema.  I feel that this is something that requires a little more research and better understanding of the reports.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
malaffoon: It's not obvious to me how this changes anything? Does the DataSourceConfig have NamedParameterJdbcTemplate beans?

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -167,4 +167,88 @@ sql:
...
+                  sd.id,
+                  sd.name
+                FROM staging_district sd
+                  LEFT JOIN district rd ON rd.id = sd.id
+                WHERE rd.id IS NULL;
+
+            update: >-
+              UPDATE district
+              SET
+                name  = sd.name
```
malaffoon: I don't know what the next step of aggregate reporting is but an export will probably need the external id (NCES id).

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/repository/impl/JdbcWarehouseToStageRepositoryIT.java
agorina: I would like to run it by you offline (in the debugger). I am not sure why it makes the difference, suspect it is tests set up.

PR #358, opened 2017-10-09
Add integration test coverage

wtritch:


build.gradle
```diff
@@ -188,14 +187,26 @@ subprojects {
...
+        doFirst {
+            jacocoTestReport.executionData(test, IT)
+            jacocoTestReport.reports.html.destination = file("${reporting.baseDir}/${name}")
+
+            //Only include execution data that exists (handle projects without integration tests)
+            jacocoTestReport {
+                executionData = files(executionData.findAll {
+                    it.exists()
+                })
+            }
```
wtritch: Interesting twist required because RDW_Ingest/common doesn't contain any integration tests.

malaffoon: so it was complaining without this?

wtritch: Yup, it was dying on "common" because no IT.exec was being generated.

PR #359, opened 2017-10-10
update README regarding coverage

malaffoon:
As part of testing this i noticed the README needed updating. I also noticed a method in the project i was spot-checking had no coverage, so i added it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,8 +62,8 @@ java -jar exam-processor/build/libs/rdw-ingest-exam-processor-1.0.0-SNAPSHOT.jar
 ```
 
 #### FTP Server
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -62,8 +62,8 @@ java -jar exam-processor/build/libs/rdw-ingest-exam-processor-1.0.0-SNAPSHOT.jar
 ```
 
 #### FTP Server
-The task service likes to use FTP to send reconciliation reports. If you don't have an FTP server but want to test
-that functionality, you can use docker to run ProFTPD. Either directly run it and modify the task service 
+The task service can be configured to use FTP to send reconciliation reports. If you don't have an FTP server but want 
+to test that functionality, you can use docker to run ProFTPD. Either directly run it and modify the task service
```
malaffoon: unrelated tweak to documentation

PR #360, opened 2017-10-10
DWR-1060 Migrate Olap - added exams

agorina:


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/config/MigrateStagingToReportingSqlConfiguration.java
```diff
@@ -15,7 +15,7 @@
  */
 @Component
 @ConfigurationProperties(prefix = "sql.migrate")
-public class MigrateStagingToWarehouseSqlConfiguration {
+public class MigrateStagingToReportingSqlConfiguration {
```
agorina: I renamed it, because this is what it actually does.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -149,7 +149,7 @@ sql:
           sql:
             insert: >-
                INSERT INTO student_ethnicity (student_id, ethnicity_id)
-                 SELECT
+                 SELECT DISTINCT
```
agorina: Missed it before.

malaffoon: damn, i was looking for those too

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -376,3 +381,178 @@ sql:
...
+                  administration_condition_id,
+                  scale_score,
+                  scale_score_std_err,
+                  performance_level,
+                  completed_at,
+                  deleted,
+                 ? as migrate_id
+                FROM exam we
+                  WHERE we.created > ? AND we.created <= ? AND type_id = 1
+                INTO OUTFILE S3 '${archive.root}/${migrate.aws.location}/exam'
```
malaffoon: Another review made me think: have we tested round-tripping non-ascii UTF-8 for this stuff?

agorina: Not yet, but I will. Here is what I have to test in addition, please comment it you see something missing:
- dates 
- int to boolean (need to adjust schema for that)
- commas in string/text fields
- non-ascii UTF-8

chdietz: Can't think of anything else.
If there a way to may this an integration test that can be repeated?

agorina: I do not have integration tests set up yet, but the objective is to include it into the tests.

PR #361, opened 2017-10-10
DWR-1060 Minor clean up.

agorina:


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -33,6 +33,7 @@ sql:
                 CREDENTIALS 'aws_iam_role=${migrate.aws.redshift.role}'
                 FORMAT AS CSV
                 DELIMITER ','
+                COMPUPDATE OFF
```
agorina: I do not think it is going to make a big difference for us, but this would not hurt.

migrate-olap/src/main/resources/banner.txt
```diff
@@ -0,0 +1,6 @@
+spring-boot: ${spring-boot.formatted-version}
+  _   _                                       _           _     _
+ |_) | \ \    /   |\/| o  _  ._ _. _|_  _    / \ |   /\  |_)   |_)  _  ._   _  ._ _|_ o ._   _
+ | \ |_/  \/\/    |  | | (_| | (_|  |_ (/_   \_/ |_ /--\ |     | \ (/_ |_) (_) |   |_ | | | (_|
+                          _|                                           |                     _|
+© The Regents of the University of California – Smarter Balanced Assessment Consortium
```
agorina: I do not think it exactly matches the font in other banners, but close enough.

PR #366, opened 2017-10-12
Upsert/Delete exam based upon Opportunity.statusDate and Opportunity.…

wtritch:
…status values

This PR is blocked by the associated schema changes in RDW_Schema.

This PR has two objectives:
1. Delete an Exam for a TRT with an Opportunity.status value of "reset."
2. Only update an Exam if the submitted Opportunity.statusDate value is equal to or after an existing Exam's statusDate value.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -92,8 +95,13 @@ public long upsert(final Exam exam, final long importId) {
         final Exam existing = findOne(exam.getAssessmentId(), exam.getStudentExamAttributes().getStudentId(), exam.getOppId());
         if (existing == null) {
             return createExam(exam, importId);
-        } else {
+        } else if (existing.getStatusDate() == null ||
+                existing.getStatusDate().equals(exam.getStatusDate()) ||
+                existing.getStatusDate().isBefore(exam.getStatusDate())) {
```
wtritch: To handle legacy data, we will update any existing exam with a "null" statusDate value.  Otherwise, ignore a TRT that attempts to update an existing exam with a later statusDate.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -535,4 +538,35 @@ public void itShouldRemoveStaleAccommodationIds() {
...
+    public void itShouldInsertASoftDeletedExam() {
+        final Exam exam = builder
+                .deleted(true)
+                .build();
+        final long examId = repository.upsert(exam, -1);
+        final Map<String, Object> insertedValues = namedParameterTemplate.queryForMap(
+                "select * from exam where id = :id",
+                new MapSqlParameterSource().addValue("id", examId));
+        assertThat(insertedValues.get("deleted")).isEqualTo(1);
+    }
```
wtritch: After discussion, if a Create and a Delete message arrive at the same time and the Delete is processed first, we want to leave the system in the same state as if the Created had arrived first.  This means that attempting to Delete a non-existent exam should create a deleted exam.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -92,8 +95,13 @@ public long upsert(final Exam exam, final long importId) {
...
             return updateExamAndChildren(exam, existing, importId);
+        } else {
+            logger.info("Ignoring TDSReport with early statusDate. Import: {}", importId);
```
malaffoon: Can you emit the two dates in the log message?

wtritch: Good call, done

PR #172, opened 2017-05-19
Added scheduler to the migrate job.

agorina:

PR #369, opened 2017-10-16
Add import_id column to root entity tables during staging/migration o…

wtritch:
…f reporting_olap

This PR is related to https://github.com/SmarterApp/RDW_Schema/pull/172 which has associated schema changes.

### DIFF
--- a/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
@@ -120,11 +120,11 @@ sql:
         student:
           sql:
             insert: >-
-              INSERT INTO student (id,  gender_id, import_id, migrate_id)
+              INSERT INTO student (id,  gender_id, update_import_id, migrate_id)
                 SELECT DISTINCT
                   ss.id,
                   rg.id,
-                  ss.import_id,
+                  ss.update_import_id,
                   ss.migrate_id
                 FROM staging_student ss
                   LEFT JOIN student rs ON rs.id = ss.id
@@ -135,7 +135,7 @@ sql:
               UPDATE student
               SET
                 gender_id  = ss.gender_id,
-                import_id = ss.import_id,
+                update_import_id = ss.update_import_id,
                 migrate_id = ss.migrate_id
               FROM staging_student ss
                 JOIN gender rg ON ss.gender_id = rg.id
@@ -175,14 +175,14 @@ sql:
         asmt:
           sql:
             insert: >-
-              INSERT INTO ica_asmt (id, grade_id, school_year, subject_id, name, import_id, migrate_id)
+              INSERT INTO ica_asmt (id, grade_id, school_year, subject_id, name, update_import_id, migrate_id)
                 SELECT DISTINCT
                   sa.id,
                   sa.grade_id,
                   sa.school_year,
                   sa.subject_id,
                   sa.name,
-                  sa.import_id,
+                  sa.update_import_id,
                   sa.migrate_id
                 FROM staging_asmt sa
                   LEFT JOIN ica_asmt ra ON ra.id = sa.id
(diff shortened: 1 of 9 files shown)


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -124,6 +124,7 @@ sql:
                middle_name,
                gender_id,
                deleted,
+               import_id,
```
agorina: I know that reporting has the same issue, but I wonder if this should be `update_import_id' in both migrates (and all SQLs in this file). I think it was an oversight (a nicer way to call a bug). 

@malaffoon and @chdietz, would like to know what you think.

wtritch: huh... that's a good question.  I could see update_import_id being more useful... It would be easy to compare states between warehouse and reporting without needing to look up the migration date/time and warehouse record's updated date/time...

malaffoon: Since this is for debugging and tracking changes, i agree.

agorina: @wtritch, could you please change it in both places?

wtritch: You got it.  ...should I change the column names or just what value gets stuffed in there?

wtritch:
Added changes required to modify reporting and reporting_olap migration to record update_import_id rather than import_id

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
wtritch: Discussed offline and decided to rename "import_id" on reporting to "update_import_id."  I also changed the new column names on reporting_olap to "update_import_id."

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -120,11 +120,11 @@ sql:
         student:
           sql:
             insert: >-
-              INSERT INTO student (id,  gender_id, import_id, migrate_id)
+              INSERT INTO student (id,  gender_id, update_import_id, migrate_id)
                 SELECT DISTINCT
                   ss.id,
                   rg.id,
-                  ss.import_id,
+                  ss.update_import_id,
```
chdietz: should the same change be made to reporting(non-olap) staging?

wtritch: I did :-)

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -204,11 +208,12 @@ sql:
         school:
           sql:
             insert: >-
-              INSERT INTO school (id, district_id, name, migrate_id)
+              INSERT INTO school (id, district_id, name, update_import_id, migrate_id)
                 SELECT DISTINCT
```
chdietz: @agorina unrelated to this PR, but while in here I was wondering why all of the DISTINCT queries?

agorina: We load data into staging in two steps: by created timestamp and then by updated timestamp. Sometimes these loads duplicate data. In the reporting migrate we use 'INSERT IGNORE', but Redshift does not support it. That is why when moving from staging to the reporting we need DISTINCT.

agorina: We load data in two steps: by created timestamp and by updates timestamp. This may create duplicate entries. In the reporting migrate we manage it using `INSERT IGNORE`, but Redshift does not support this feature. In order to avoid duplicates in the reporting OLAP we need to use DISTINCT.

PR #370, opened 2017-10-17
Added integration tests for the ArtClient which only validates data i…

bgsavage:
…s returned.

### DIFF
--- a/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/JsonOrganizationParser.java
+++ b/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/JsonOrganizationParser.java
@@ -77,7 +77,7 @@ import static org.springframework.util.StringUtils.hasText;
  * </pre>
  * </p>
  */
-class JsonOrganizationParser implements OrganizationParser {
+public class JsonOrganizationParser implements OrganizationParser {
     private static final Logger logger = LoggerFactory.getLogger(JsonOrganizationParser.class);
 
     // NOTE: this state makes parse() completely NOT thread-safe
--- a/task-service/build.gradle
+++ b/task-service/build.gradle
@@ -41,3 +41,11 @@ bootRun {
         jvmArgs project.jvmArgs.split('\\s+')
     }
 }
+
+task AST(type: Test) {
+    include '**/*AST.*'
+    outputs.upToDateWhen { false }
+    doFirst {
+        println "Running ART System Tests..."
+    }
+}
--- a/task-service/build.gradle
+++ b/task-service/build.gradle
@@ -29,6 +29,7 @@ dependencies {
     compile project(':rdw-ingest-common')
 
     testCompile 'org.springframework.boot:spring-boot-starter-test'
+    testCompile project(':rdw-ingest-package-processor')
 
     // fixes java 8 time marshaling
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.8.7'
--- a/task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultUpdateOrganizationsService.java
+++ b/task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultUpdateOrganizationsService.java
@@ -39,7 +39,11 @@ public class DefaultUpdateOrganizationsService implements UpdateOrganizationsSer
 
     @Override
     public void doUpdate(final String state) {
-        final byte[] payload = Bytes.concat(
+        importServiceClient.postOrganizationPayload(getPayload(state));
+    }
+
+    public byte[] getPayload(final String state) {
+        return Bytes.concat(
                 OpenBrace,
                 DistrictGroups, organizationRepository.getGroupsOfDistrictsPayload(state),
                 Districts, organizationRepository.getDistrictsPayload(state),
@@ -47,7 +51,5 @@ public class DefaultUpdateOrganizationsService implements UpdateOrganizationsSer
                 Schools, organizationRepository.getSchoolsPayload(state),
                 CloseBrace
         );
-
-        importServiceClient.postOrganizationPayload(payload);
     }
 }
--- a/task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
+++ b/task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
@@ -4,7 +4,15 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.opentestsystem.rdw.ingest.common.model.District;
+import org.opentestsystem.rdw.ingest.common.model.DistrictGroup;
+import org.opentestsystem.rdw.ingest.common.model.School;
+import org.opentestsystem.rdw.ingest.common.model.SchoolGroup;
+import org.opentestsystem.rdw.ingest.common.util.DataElementErrorCollector;
+import org.opentestsystem.rdw.ingest.common.util.ParserHelper;
+import org.opentestsystem.rdw.ingest.processor.service.impl.JsonOrganizationParser;
 import org.opentestsystem.rdw.ingest.tasking.UpdateOrganizationsConfiguration;
+import org.opentestsystem.rdw.ingest.tasking.service.impl.DefaultUpdateOrganizationsService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.context.annotation.Import;
@@ -12,10 +20,12 @@ import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.Collection;
+
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * This test will run only if the oauth2 secret and password are provided at runtime:
+ * This test requires the oauth2 secret and password to be provided at runtime:
  * <ul>
  * <li>task.update-organizations.art-client.oauth2.client-secret</li>
  * <li>task.update-organizations.art-client.oauth2.password</li>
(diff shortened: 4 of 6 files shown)


task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
```diff
@@ -0,0 +1,76 @@
...
+    public void itShouldBeAvailable() {
+        assertThat(artClient.isAvailable()).isTrue();
+    }
+
+    @Test
+    public void itShouldGetGroupsOfDistrictsByPayload() throws JSONException {
+        JSONObject actual = new JSONObject(new String(artClient.getGroupsOfDistrictsPayload(State)));
+
+        assertThat(actual).isNotNull();
+        assertThat(actual.getJSONArray(SearchResults).length()).isGreaterThan(0);
```
bgsavage: I'm not sure if we want to do in depth contract-type testing, or we just care that the endpoint returns data?  For now, I'm just doing the latter.  

Thoughts?

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This test will run only if the oauth2 secret and password are provided at runtime:
+ * <ul>
+ * <li>task.update-organizations.art-client.oauth2.client-secret</li>
+ * <li>task.update-organizations.art-client.oauth2.password</li>
+ * </ul>
+ * For example,<pre>
+ * java -Dtask.update-organizations.art-client.oauth2.client-secret=MyClientSecret
+ *      -Dtask.update-organizations.art-client.oauth2.password=MyPassword
+ * </pre>
```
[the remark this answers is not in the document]
malaffoon: I would love to spend the time and effort to make these tests conditional on the correct properties being defined. We could have a special annotation, yada, yada. But we aren't going to do that (i think), so ...

I think perhaps we should discuss treating these more like the RSTs (redshift system tests): a separate group of tests that is explicitly invoked. As part of that, the invoker needs to reference an external properties file or pass lots of parameters, etc. If the invoker gets things wrong, they have to fix it. 

We might need categories. For example, these might be AST (ART system tests).

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
```diff
@@ -0,0 +1,76 @@
...
+        assertThat(actual).isNotNull();
+        assertThat(actual.getJSONArray(SearchResults).length()).isGreaterThan(0);
+    }
+
+    @Test
+    public void itShouldGetSchoolsByPayload() throws JSONException {
+        JSONObject actual = new JSONObject(new String(artClient.getSchoolsPayload(State)));
+
+        assertThat(actual).isNotNull();
+        assertThat(actual.getJSONArray(SearchResults).length()).isGreaterThan(0);
```
malaffoon: You could add expected payload snippets as properties. Then this could be a bit more rigorous, something like:
```
    @Value("ast.expected.schools")
    private String expectedSchools;
    ...
        assertThat(actual.toString()).startsWith(expectedSchools);
```

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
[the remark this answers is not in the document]
malaffoon: For now, perhaps `@IfProfileValue` to control it? Or just resort to the good ol' `@Ignore("need credentials")`

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
bgsavage: Then won't these tests fail every time someone changes something in ART?

malaffoon: That's why the expected payload has to be driven by properties. I couldn't think of another way to allow for changing system data.

bgsavage: If we automated how the properties are generated in a way that they were regenerated every time the underlying data changed, I'd be good with that.  Otherwise, it still sounds like it will be difficult to maintain.

If we are interested in just making sure the models don't change, I can write contract tests that just dig down into specific key properties to make sure the data isn't null.

malaffoon: That's a good point, instead of using the toString, look for particular key/values. Sure, somebody could still remove those organizations but that would be obnoxious  ;)

bgsavage: I can bring in the `JsonOrganizationParser` and see if I'm able to parse the payload from Art.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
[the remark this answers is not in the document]
bgsavage: Any issue if I just add these two props to the list of required environment variables required for the build machine and local dev?  The Jira requires that these be run as part of our CI builds.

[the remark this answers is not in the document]
malaffoon: Fine by me.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
jtreuting: I think we need this IT to alert us if there is a breaking change from the ART system we are trying to integrate with.  So I think we'd need a little more.

task-service/src/test/resources/application-organization.yml
```diff
@@ -0,0 +1,21 @@
...
+      districts-url: https://art-deployment.sbtds.org/rest/district?stateAbbreviation={state}&pageSize=1000
+      groups-of-schools-url: https://art-deployment.sbtds.org/rest/groupofinstitution?stateAbbreviation={state}&pageSize=1000
+      schools-url: https://art-deployment.sbtds.org/rest/institution?stateAbbreviation={state}&pageSize=20000
+      status-url: https://art-deployment.sbtds.org/rest/status
+      oauth2:
+        access-token-uri: https://sso-deployment.sbtds.org:443/auth/oauth2/access_token?realm=/sbac
+        client-id: sbacdw
+        client-secret: ${rdw.ingest.art.client.oauth2.client.secret}
+        username: ca.admin@example.com
+        password: ${rdw.ingest.art.client.oauth2.password}
```
bgsavage: New environments required to run integration tests in Ingest.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
```diff
@@ -0,0 +1,146 @@
...
+                                .build())
+                        .build())
+                .build();
+
+        // Under the hood, organization service will invoke all of the art client methods to get the payload.
+        final Collection<School> actualSchools = parser.parse(organizationsService.getPayload("CA"), new ParserHelper(errorCollector));
+        assertThat(errorCollector.isEmpty()).isTrue();
+
+        assertThat(actualSchools).contains(expectedSchoolWithDistrictGroup);
+        assertThat(actualSchools).contains(expectedSchoolWithSchoolGroup);
```
bgsavage: @jtreuting  and @malaffoon , thoughts on this test?  

I realize definitely test knows too much about the implementation details of how the payload gets used, but I think that's actually what we are going for right?  We want to make sure we can successfully parse the data from ART.  If a breaking change is introduced into ART that we can't parse, this test will fail.  Once we update the parser, this test should again pass.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/JsonOrganizationParser.java
```diff
@@ -77,7 +77,7 @@
  * </pre>
  * </p>
  */
-class JsonOrganizationParser implements OrganizationParser {
+public class JsonOrganizationParser implements OrganizationParser {
```
bgsavage: Made this public so that the IT can use it.  Better than having to create another parser just for the IT, and then having two parsers to keep in synch with ART, right?

task-service/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
     compile project(':rdw-ingest-common')
 
     testCompile 'org.springframework.boot:spring-boot-starter-test'
+    testCompile project(':rdw-ingest-package-processor')
```
bgsavage: testCompile new dependence for ART IT.

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/service/impl/DefaultUpdateOrganizationsService.java
```diff
@@ -39,15 +39,17 @@ public DefaultUpdateOrganizationsService(final OrganizationRepository organizati
 
     @Override
     public void doUpdate(final String state) {
-        final byte[] payload = Bytes.concat(
+        importServiceClient.postOrganizationPayload(getPayload(state));
+    }
+
+    public byte[] getPayload(final String state) {
```
bgsavage: extracted a method out so I can get the payload from the IT without having to mock `importServiceClient`.

task-service/src/test/java/org/opentestsystem/rdw/ingest/tasking/repository/ArtClientIT.java
malaffoon: Given the constraints we're working under, i think this is good.

task-service/build.gradle
```diff
@@ -40,3 +41,11 @@ bootRun {
...
     }
 }
+
+task AST(type: Test) {
+    include '**/*AST.*'
+    outputs.upToDateWhen { false }
+    doFirst {
+        println "Running ART System Tests..."
+    }
+}
```
bgsavage: New gradle task to run the art system tests.

PR #372, opened 2017-10-18
Add a status indicator to track DB connection health from the migrate…

wtritch:
…-olap service to the olap/redshift database.

This PR adds a status indicator for tracking the ability of the service to communicate with the olap/redshift database.

### DIFF
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapDatabaseStatusIndicator.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapDatabaseStatusIndicator.java
@@ -1,0 +1,77 @@
+package org.opentestsystem.rdw.ingest.migrate.olap.status;
+
+import com.google.common.collect.ImmutableMap;
+import org.opentestsystem.rdw.common.status.AbstractStatusIndicator;
+import org.opentestsystem.rdw.common.status.DiagnosticLevel;
+import org.opentestsystem.rdw.common.status.Status;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An {@link AbstractStatusIndicator} for the OLAP database status.
+ * <p>
+ * Slightly different than web diagnostic spec this adds "databaseOperations",
+ * each being a Status with "type" (READ/WRITE), and "responseTime".
+ * </p>
+ */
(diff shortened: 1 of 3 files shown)


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapDatabaseStatusIndicator.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public String name() {
+        return "OLAP-database-operations";
+    }
+
+    private Status getOlapReadStatus() {
+        final Status.Builder builder = Status.builder()
+                .detail("type", "READ");
+        responseTime(builder, 10000, () ->
```
wtritch: Unfortunately, I saw this read request taking upwards of 6 seconds when reading the single value from the status_indicator table.  My guess is that's just standard redshift query overhead?

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapDatabaseStatusIndicatorRST.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ActiveProfiles("redshift")
+@Import(MigrateOlapDatabaseStatusIndicator.class)
+public class MigrateOlapDatabaseStatusIndicatorRST extends SpringBatchIT {
```
wtritch: Added a new RST to test the indicator against a Real Live OLAP datasource.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/status/MigrateOlapDatabaseStatusIndicator.java
malaffoon: I suppose we could remove the timing of these calls instead of having a stupid-big time threshold.

wtritch: It's funny, when running this by standing up a the migrate-olap service locally and *not* as an integration test I end up with response times in the 200ms range...  I wonder if we're doing something silly with our integration tests...

agorina: Yeah, I have the same concern. I think I saw a sub-task on one of the JIRAs to look into it.

bgsavage: Didn't you see the `Thread.sleep(5800)` in there?

PR #373, opened 2017-10-18
fix inter-project dependencies with a library jar

wtritch:


build.gradle
```diff
@@ -164,7 +164,7 @@ subprojects {
     }
 
     // Skip running integration and system tests automatically, because they take a long time
-    test.exclude '**/*IT.*', '**/*RST.*'
+    test.exclude '**/*IT.*', '**/*RST.*', '**/*AST.*'
```
wtritch: unrelated, we don't want to run the AST tests when running the standard 'test' task

bgsavage: Thanks, I saw I missed this too while I was in here troubleshooting.

package-processor/build.gradle
```diff
@@ -29,4 +29,17 @@ dependencies {
...
+    from sourceSets.main.output
+}
+configurations {
+    library {
+        extendsFrom testRuntime
+    }
+}
+artifacts {
+    library libraryJar
+}
```
wtritch: It looks like the problem is that when we depend upon the standard artifact it's been re-packaged as a spring boot jar... which apparently causes some issues.  I took a page out of the common module and created a library jar without any spring boot processing/repackaging for cross-module use.

task-service/build.gradle
```diff
@@ -29,7 +29,7 @@ dependencies {
     compile project(':rdw-ingest-common')
 
     testCompile 'org.springframework.boot:spring-boot-starter-test'
-    testCompile project(':rdw-ingest-package-processor')
+    compile project(path: ':rdw-ingest-package-processor', configuration: 'library')
```
wtritch: Depending upon the "library" jar rather than the module artifact allows the tests to compile without issue.

bgsavage: So why don't we need to do this for `:rdw-ingest-common`?

package-processor/build.gradle
malaffoon: Ugh, really?

task-service/build.gradle
wtritch: I believe it has to do with ingest-common's build.gradle:
```
bootRepackage.enabled = false
bootRun.enabled = false
```

malaffoon: i *really* don't like this being `compile` vs `testCompile`

package-processor/build.gradle
wtritch: bleh >_<  what a mess.  @bgsavage thoughts?  Do you want to take this and run with it?

wtritch: @malaffoon In the meantime, the build is broken.  Should I merge this just to get us running again?  Or should we just remove the AST and start over?

malaffoon: Just checkin the exclude of **AST* and the build will work.

bgsavage: @malaffoon  just mentioned my thoughts.  If spring boot apps are meant to be used as libraries out of the box, I could just move the class I need to common.

bgsavage: Re: build will work, I think it'll still fail on the compileTestJava step (at least `gradle clean build` does for me locally)

task-service/build.gradle
wtritch: Agreed.  What are the alternatives?
1. create an extra level of hierarchy between the task-service and package-processor as yet another sub-module called shared-blah?
2. pull shared functionality up into the existing rdw-common sub-module?
3. ...?

package-processor/build.gradle
malaffoon: Ugh.

task-service/build.gradle
bgsavage: I'm good with merging this PR so people aren't blocked with a broken build for now while I work on moving this out to rdw-ingest-common

wtritch: Yeah, likewise.  I'll merge for now and hope to see this all go away in your future PR =)

malaffoon:
Note that my vote matters at this point but ...

I think we should revert all of these AST changes. If we want to do system tests, we should embrace them and do them properly. This test does a basic check against the ART deployed in the QA environment. I don't think that fulfills the spirit of the SOW. So why compromise and complicate the build system for it?

wtritch:
@malaffoon I like the idea of doing it cleaner... I'm just not sure exactly what that entails.  I like the idea of pulling the entire ART API validation concept out of the build process.  What kind of implementation are you thinking of?  A scheduled validation service of some kind that just queries and reports/logs?  An executable .jar that hits external resources and reports if things look (im)proper that's run on demand?

PR #374, opened 2017-10-18
Feature/dwr 1065 add art integration tests

bgsavage:
Moved to JsonOrganizationParser and it's interface out to rdw-ingest-common so that ArtClientAST can get it through common instead of depending on rdw-package-processor.


bgsavage:
This was pretty straight forward as IntelliJ did all the heavy lifting.

malaffoon:
As i noted in the other PR, i think we should revert all this stuff.

PR #376, opened 2017-10-19
DWR-1092 STEP 1- added more tests

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -269,7 +268,7 @@ public boolean equivalent(final Exam other) {
                 Objects.equals(handScoreProject, other.handScoreProject) &&
                 Objects.equals(contract, other.contract) &&
                 Objects.equals(testReason, other.testReason) &&
-                Objects.equals(assessmentAdminStartedAt, other.assessmentAdminStartedAt) &&
+//      TODO:          Objects.equals(assessmentAdminStartedAt, other.assessmentAdminStartedAt) &&
                 Objects.equals(startedAt, other.startedAt) &&
```
agorina: I will need help here, or more research.  

This is the only 'date' field in MySQL, but for some reason it still gets a time and 4 hours difference between what is stored and read back. I am not going to merge until it is resolved.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamItem.java
```diff
@@ -123,7 +123,7 @@ private ExamItemSubScore getByTraitId(final int traitId) {
                 return subScore;
             }
         }
-        return null;
+        return new ExamItemSubScore(traitId, null, null);
```
agorina: Since we are pivoting the data, we always have 3 trait score in the db table, and when we read the data back they all need to be present to compare the before and after for the update.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -241,8 +242,62 @@ public void itShouldCreateExams() {
     }
 
     @Test
+    public void itShouldNotModifyAnUnchangedExamWithNullValues() {
```
agorina: I have noticed in my last change that I accidently merged into develop without a PR that we were not handling nullable values when comparing before and after for the update. All null Long, Double and Int were converted to 0 and treated as an update. I fixed it in that accidental merge, but adding tests here to prove that it works.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -39,7 +38,7 @@
     private Long handScoreProject;
     private String contract;
     private String testReason;
-    private Date assessmentAdminStartedAt;
+    private LocalDate assessmentAdminStartedAt;
```
agorina: Does this look right for storing just a date?

bgsavage: I'm not trying to claim to be a java expert, but why not `Instant` ?

agorina: According to the data model this field has 'YYYY-MM-DD', Instant is more precise and I did not want to 'create' hours, min, sec.

agorina: It is getting more complicated than I wanted. Maybe I should change to `timestamp` in the db  and `Instant` here. 

Here is what is happening:
Data Model says that the format is 'YYYY-MM-DD'.
Our object model (generated from XSD) returns java.util.Date.
To convert java.until.Date to LocalDate, it looks like I need a ZoneId? 
If the above is true, then I am back to the time offset.

Oh, and I forgot to mention that working with java dates is my favorite thing!
Any help is very much appreciated.

chdietz: I think maybe the issue is the data model is wrong, or if the XSD has Date should we go with that because that may be the way ETS will send it.

malaffoon: I agree that the generated data model may be wrong. `java.util.Date` has a timezone associated with it, which we don't want. We should change the data model to be LocalDate. In lieu of that we can add a utility that converts the data model Date to a LocalDate assuming some timezone (not sure if it should be GMT, EST, PST and hopeful that it doesn't matter a lot).

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
malaffoon: LocalDate should be comparing only year,month,day?

PR #378, opened 2017-10-19
Reduce RST spring environment overhead

wtritch:
These changes make a minor improvement to RST execution speed, but the suite still takes ~15 minutes to run on my local machine.  After sampling CPU usage during one of the tests, it appears that the vast majority of time is spent waiting on Aurora and Redshift.  I'm not sure if that can be sped up at all.  At least, not without a drastic change like switching to a developer-local mock AWS stack: https://github.com/localstack/localstack

### DIFF
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingApplication.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingApplication.java
@@ -6,13 +6,20 @@ import org.opentestsystem.rdw.migrate.common.config.MigrateWarehouseToStagingSql
 import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.cloud.aws.autoconfigure.cache.ElastiCacheAutoConfiguration;
+import org.springframework.cloud.aws.autoconfigure.context.ContextStackAutoConfiguration;
+import org.springframework.cloud.aws.autoconfigure.mail.MailSenderAutoConfiguration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.PropertySource;
 
 /**
  * Spring boot application class
  */
-@SpringBootApplication
+@SpringBootApplication(exclude = {
+        ElastiCacheAutoConfiguration.class,
+        MailSenderAutoConfiguration.class,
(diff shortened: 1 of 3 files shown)


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingApplication.java
```diff
@@ -6,13 +6,20 @@
...
 
 /**
  * Spring boot application class
  */
-@SpringBootApplication
+@SpringBootApplication(exclude = {
+        ElastiCacheAutoConfiguration.class,
+        MailSenderAutoConfiguration.class,
+        ContextStackAutoConfiguration.class
+})
```
wtritch: I added the same exclusions that exist on the import-service.  During testing I found the ElastiCacheAutoConfiguration was consuming a fair number of cycles at startup.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/SpringBatchIT.java
```diff
@@ -14,7 +15,10 @@
 
 @RunWith(SpringRunner.class)
 @SpringBootTest
-@ContextConfiguration(classes = {SpringBatchIT.BatchTestConfig.class})
+@ContextConfiguration(classes = {
+        TestConfiguration.class,
+        SpringBatchIT.BatchTestConfig.class
+})
```
wtritch: I added the Spring-defined TestConfiguration to the context configuration.  It adds a few fast-aborts to internal spring autoconfiguration/setup.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/TestAppConfig.java
```diff
@@ -16,14 +13,12 @@
...
-@Import({MigrateOlapReportingConfiguration.class,
-        OlapReportingMigrateStepsConfig.class,
+@Import({
+        ConfigurationPropertiesAutoConfiguration.class,
         YamlPropertiesConfigurator.class,
+        MigrateOlapReportingConfiguration.class,
         DataSourceConfiguration.class,
-        MigrateWarehouseToStagingSqlConfiguration.class,
-        MigrateStagingToReportingSqlConfiguration.class,
         ArchiveServiceConfiguration.class
```
wtritch: I got rid of the general \@EnableAutoConfiguration annotation in favor of just including the imports we need.  I also removed some imports that were already coming in as transitive dependencies.

PR #380, opened 2017-10-19
Allow olap staging table truncation by manually starting a transaction.

wtritch:
This PR is more of an idea and talking point than a confirmed PR.

The issue with calling `TRUNCATE <tableName>;` in Redshift is that the session/connection is in auto-commit mode, which causes issues when the TRUNCATE command attempts to commit.
```
Migrate Job id 590] is complete with status [exitCode=FAILED;exitDescription=org.springframework.jdbc.UncategorizedSQLException: StatementCallback; 
uncategorized SQLException for SQL [truncate staging_administration_condition]; 
SQL state [HY000]; error code [10040]; 
[Amazon][JDBC](10040) Cannot use commit while Connection is in auto-commit mode.; 
nested exception is java.sql.SQLException: 
[Amazon][JDBC](10040) Cannot use commit while Connection is in auto-commit mode.
```
After googling around a bit, the best solution I found was to manually add a `BEGIN` transaction boundary before executing the TRUNCATE statement.  Thoughts?

### DIFF
--- a/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
+++ b/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
@@ -2,18 +2,21 @@ sql:
   warehouseToStage:
   # ------------ Truncate ------------------------------------------------------------------
     truncateStageList:
-      - delete from staging_administration_condition
-      - delete from staging_completeness
-      - delete from staging_ethnicity
-      - delete from staging_gender
-      - delete from staging_grade
-      - delete from staging_student
-      - delete from staging_student_ethnicity
-      - delete from staging_school
-      - delete from staging_district
-      - delete from staging_asmt
-      - delete from staging_exam
-      - delete from staging_exam_claim_score
+      # Turn off autocommit to allow TRUNCATE to process
+      - BEGIN
+      - truncate staging_administration_condition
+      - truncate staging_completeness
+      - truncate staging_ethnicity
+      - truncate staging_gender
+      - truncate staging_grade
+      - truncate staging_student
+      - truncate staging_student_ethnicity
+      - truncate staging_school
+      - truncate staging_district
+      - truncate staging_asmt
+      - truncate staging_exam
+      - truncate staging_exam_claim_score
+      - END
 
     entities:
         # ------------ administration_condition  -------------------------------------------------------------------


wtritch:
It looks like AWS once considered this a "bug" then decided it was a "feature."
https://forums.aws.amazon.com/message.jspa?messageID=805358

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -2,18 +2,21 @@ sql:
...
+      - truncate staging_gender
+      - truncate staging_grade
+      - truncate staging_student
+      - truncate staging_student_ethnicity
+      - truncate staging_school
+      - truncate staging_district
+      - truncate staging_asmt
+      - truncate staging_exam
+      - truncate staging_exam_claim_score
+      - END
```
malaffoon: Remind me why we want to use truncate vs. delete? Is it a performance thing? I know truncate is generally more efficient but do we know the impact?

agorina:
I have mixed feelings about it. What I am debating is whether we get benefits of using TRUNCATE vs DELETE. We do not have any indexes or things like that.. Do you have any thoughts?

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
agorina: Oh, just saw this. 
The main reason is "because reporting did" :)
And reporting did because it was easy and recommended over DELETE since it leaves no traces behind. 

But I am also questioning if it is needed here, considering the solution.
I am fine with using DELETE unless there are reasons against.

wtritch: In general, Redshift highly recommends truncate over unqualified deletes since no VACUUM or ANALYZE is required: http://docs.aws.amazon.com/redshift/latest/dg/r_TRUNCATE.html
However, how large is the performance gain?  I have no idea.  I'm going to try an export/import of 10k exam records and run delete vs truncate to see how large the difference is.

wtritch: It doesn't look like deleting is time-consuming enough for truncation to be worthwhile:
```
[2017-10-19 16:56:33] [01000] Load into table 'staging_exam' completed, 100000 record(s) loaded successfully.
[2017-10-19 16:56:33] completed in 2s 909ms
sql> delete from staging_exam
[2017-10-19 16:56:47] 100000 rows affected in 424ms
```
less than half a second to delete 100k exam records from the staging table.

wtritch:
Abandoning this review as "won't fix"

PR #382, opened 2017-10-20
DWR-1092 Refactor application.yml for storing sbac score ranges.

agorina:
SbacScoreConfiguration is defined in the `ingest-common` but is used by both `ingest-exam-processor` and `ingest-package-processor`. This refactoring moves the yml settings into one file in the `ingest-common` project.

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -45,19 +45,11 @@ spring:
     serialization:
       WRITE_DATES_AS_TIMESTAMPS: false
 
+  profiles:
+    include: ingest-common
+
 jdbc:
   retry:
     default:
       max-attempts: 4
       delay: 30
-
-sbac:
-  scaleScore:
-    min: 1500
-    max: 3500
-  thetaScore:
-    min: -5
-    max: 5
-  itemScore:
-    min: -1
-    max: 20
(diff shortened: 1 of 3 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -45,19 +45,11 @@ spring:
     serialization:
       WRITE_DATES_AS_TIMESTAMPS: false
 
+  profiles:
+    include: ingest-common
+
```
agorina: Does this look right? (seems to work)

malaffoon: I've never done this but, if it seems to work.

How does this interact with `spring.profiles.active` which we use in docker and kubernetes?

malaffoon: Could you eliminate the application-ingest-common.yml file and just use default values in the @Value annotations instead? If an app needed to override the values, they could (in the external yml file) but there would be no need to have the configuration properties in *any* yml file be default. Seems less messy.

agorina: To answer the first question: according to Spring docs "The spring.profiles.include property can be used to unconditionally add active profiles":https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html

And in IT tests we do use active profile and it includes `ingest-common` too.

As for @value, still trying to see how to do it since I have nested configurations.

malaffoon: I'm fine with this approach then.

PR #391, opened 2017-10-26
add exception message to import exception

malaffoon:
If you look at the jira (https://jira.fairwaytech.com/browse/DWR-1091) this came up when i was diagnosing a problem in production. It would've been a lot nicer to have the SQL exception text in the ImportException message.

Also some unrelated cleanup.

### DIFF
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/GroupProcessor.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/GroupProcessor.java
@@ -9,7 +9,9 @@ import org.opentestsystem.rdw.ingest.common.model.GroupMessage;
 public interface GroupProcessor {
 
     /**
-     * @param groupUploadMessage the upload to process
+     * Processes an upload message
+     *
+     * @param message a message sent after successful upload from RDW_Admin for a groups CSV
      */
-    void process(final GroupMessage groupUploadMessage);
+    void process(final GroupMessage message);
 }
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
@@ -37,15 +37,8 @@ public class DefaultGroupProcessor implements GroupProcessor {
         this.groupImportService = groupImportService;
     }
 
-    /**
-     * Processes an upload message
-     *
-     * @param message A message sent after successful upload from RDW_Admin for a groups CSV
-     */
     @Override
     public void process(final GroupMessage message) {
-
-        // todo: Need to design message that RDW_Admin will publish to queue
         final String jsonMessage = message.toJson();
         logger.debug("Received upload message: " + (jsonMessage.length() > 80 ? jsonMessage.substring(0, 80) + "..." : jsonMessage));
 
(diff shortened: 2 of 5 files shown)


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/GroupProcessor.java
```diff
@@ -9,7 +9,9 @@
 public interface GroupProcessor {
 
     /**
-     * @param groupUploadMessage the upload to process
+     * Processes an upload message
+     *
+     * @param message a message sent after successful upload from RDW_Admin for a groups CSV
```
malaffoon: IMO, the interface should have the useful javadocs while the implementation doesn't (unless the implementation introduces side effects or other unexpected issues).

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
```diff
@@ -58,7 +51,7 @@ public void process(final GroupMessage message) {
             groupImportService.processBatch(batchId);
         } catch (final Exception e) {
             logger.error("Problem processing batch: {}", batchId, e);
-            throw new ImportException(ImportStatus.INVALID, "Problem processing Batch: " + batchId);
+            throw new ImportException(ImportStatus.INVALID, "Problem processing batch " + batchId + ": " + e.getMessage());
```
malaffoon: This is the fix.

group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessorTest.java
```diff
@@ -60,22 +63,37 @@ public void itShouldCallServiceMethodsInOrder() {
...
+    }
+
+    @Test
+    public void itShouldThrowNullMessageOnProcessingError() {
+        doThrow(new RuntimeException()).when(loadService).loadBatch(anyString(), anyLong());
+        try {
+            processor.process(message);
+            fail("it should have thrown an ImportException");
+        } catch (final ImportException e) {
+            assertThat(e.getMessage()).contains("123").contains("null");
```
malaffoon: Yes, the message will be something like `Problem processing batch 123: null` if the cause exception has no message.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
```diff
@@ -37,15 +37,8 @@ public DefaultGroupProcessor(final ProcessingStandardizationService standardizat
...
 
-    /**
-     * Processes an upload message
-     *
-     * @param message A message sent after successful upload from RDW_Admin for a groups CSV
-     */
     @Override
     public void process(final GroupMessage message) {
-
-        // todo: Need to design message that RDW_Admin will publish to queue
```
malaffoon: @wtritch i'm assuming this TODO is TODONE?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
wtritch: Interesting... the logger.error logging statement on line 53 didn't include the message?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
wtritch: TODONE

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultGroupProcessor.java
malaffoon: The log file had the message. But that is an additional step to get that information. In this case, there was clearly a typo in the CSV but it took 10 minutes to get there instead of being obvious from the import record.

PR #393, opened 2017-10-27
make log level consistent

malaffoon:
Just some minor cleanup while i was documenting log messages

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
@@ -50,7 +50,7 @@ public class ExamProcessorConfiguration {
 
             process(report, importId);
         } catch (final Exception e) {
-            logger.warn("failed with an unexpected import error: " + e.getMessage());
+            logger.warn("failed with an unexpected error: " + e.getMessage());
             importRepository.updateStatusAndMessageById(importId, ImportStatus.INVALID, e.getMessage());
         }
     }
@@ -61,10 +61,10 @@ public class ExamProcessorConfiguration {
             importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
 
         } catch (final ImportException e) {
-            logger.info("failed with an import error: " + e.getMessage());
+            logger.warn("failed with an import error: " + e.getMessage());
             importRepository.updateStatusAndMessageById(importId, e.getStatus(), e.getMessage());
         } catch (final Exception e) {
-            logger.warn("failed with an unexpected import error: " + e.getMessage());
+            logger.warn("failed with an unexpected error: " + e.getMessage());
             importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
         }
     }
--- a/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
+++ b/group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
@@ -35,37 +35,30 @@ public class GroupProcessorConfiguration {
 
     @ServiceActivator(inputChannel = Processor.INPUT)
     public void process(final Message<?> message) {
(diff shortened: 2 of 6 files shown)


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessorConfiguration.java
```diff
@@ -35,37 +35,30 @@ public GroupProcessorConfiguration(final GroupProcessor groupProcessor,
 
     @ServiceActivator(inputChannel = Processor.INPUT)
     public void process(final Message<?> message) {
-        RdwMessageHeaderAccessor accessor = null;
+        final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         try {
             //payload should be a json object
-            accessor = RdwMessageHeaderAccessor.wrap(message);
```
malaffoon: This won't throw an exception; assigning it outside the try/catch eliminates a few lines of code.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseCheck.java
```diff
@@ -18,20 +17,17 @@
     private static final Logger logger = LoggerFactory.getLogger(WarehouseCheck.class);
 
     private final WarehouseToStageRepository warehouseToStageRepository;
-    private final MigrateWarehouseToStagingSqlConfiguration warehouseToStagingSqlConfiguration;
```
malaffoon: Unrelated, this wasn't being used anywhere in the code.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseCheck.java
```diff
@@ -75,10 +71,8 @@ private boolean checkSubjectClaimScores() {
         if (warehouseSubjectClaimScores.equals(reportingSubjectClaimScores)) {
             return true;
         } else {
-            logger.error("warehous subject_claim_score table does not match reporting");
+            logger.warn("warehouse subject_claim_score table does not match reporting");
```
chdietz: So this will require intervention to migrate but import can continue.  Still a pretty serious problem because will stop new imports from making it to the user facing UI.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
```diff
@@ -61,10 +61,10 @@ private void process(final TDSReport tdsReport, final long importId) {
             importRepository.updateStatusAndMessageById(importId, ImportStatus.PROCESSED, null);
 
         } catch (final ImportException e) {
-            logger.info("failed with an import error: " + e.getMessage());
+            logger.warn("failed with an import error: " + e.getMessage());
```
chdietz: Is this really an import error? Exam processor is getting an error updating the import record?
I guess what I am asking for consistency is should we just say we have an unexpected error and the message includes the class path and the exception?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/WarehouseCheck.java
malaffoon: Yes. Following the guidelines, does this warrant somebody being called in the middle of the night to fix it?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorConfiguration.java
malaffoon: I was trying to follow the pattern of: an `ImportException` is something that we "expect" during import processing and we use the verbiage "import error" when logging it. Any other exception is not "expected" and could be anything so, even though it happened during import, it isn't an "import error". 

And, yeah, i realize that being consistent with logging is almost as impossible has having a constructive discussion about politics or religion.

PR #394, opened 2017-10-27
make application configuration consistent

malaffoon:
After looking over Will's shoulder yesterday when he was pushing things around to make IT's work in reporting, it seemed worthwhile to make a similar change in RDW_Ingest: separate the application configuration from the main application class. The test runner will then not see the application configuration unless you explicitly tell it to. That makes the DI in the tests cleaner.

While i was in there i renamed a couple classes that had configuration in their name but aren't. That was left over from using Spring Stream's naming convention for processors and it has been bugging me for months.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
@@ -1,11 +1,8 @@
 package org.opentestsystem.rdw.ingest.processor;
 
-import org.opentestsystem.rdw.common.status.StatusConfiguration;
-import org.opentestsystem.rdw.utils.YamlPropertiesConfigurator;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cache.annotation.EnableCaching;
-import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.PropertySource;
 
 /**
@@ -14,7 +11,6 @@ import org.springframework.context.annotation.PropertySource;
 @SpringBootApplication
 @EnableCaching
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
-@Import( { ExamProcessorConfiguration.class, YamlPropertiesConfigurator.class, StatusConfiguration.class } )
 public class ExamProcessorApplication {
 
     public static void main(final String[] args) {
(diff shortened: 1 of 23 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
```diff
@@ -14,7 +11,6 @@
 @SpringBootApplication
 @EnableCaching
 @PropertySource(value = "classpath:build-info.properties", ignoreResourceNotFound = true)
-@Import( { ExamProcessorConfiguration.class, YamlPropertiesConfigurator.class, StatusConfiguration.class } )
 public class ExamProcessorApplication {
```
malaffoon: I chatted with Will about renaming all the applications to be just `Application`. That's the usual name in a standalone project. Will pointed out that, in a multi-module project like RDW_Ingest, that makes it a bit harder to find the application class you're looking for. So i left them alone.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessor.java
```diff
@@ -18,18 +18,17 @@
 
 import java.nio.charset.StandardCharsets;
 
-
 @EnableBinding(Sink.class)
-public class GroupProcessorConfiguration {
-    private static final Logger logger = LoggerFactory.getLogger(GroupProcessorConfiguration.class);
+public class GroupProcessor {
```
malaffoon: This one was a bit trickier since there was already a `GroupProcessor`. I renamed that to be `GroupMessageProcessor` and then continued to follow my pattern ...

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorApplication.java
chdietz: Makes sense

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessor.java
chdietz: This works for me.  This made me go and look at exam-processor module again.
Not sure TDSReportProcessor is such a great name. May be left over from when we started.
Maybe it should be TestResultProcessor ?

malaffoon: I don't disagree; but that'll have to be cleanup later.

PR #396, opened 2017-10-30
DWR-1067  Store additional item data

agorina:


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -79,7 +79,12 @@ public String validate(final String name, final String rawValue, final int maxLe
      */
     public static final Function<String, Boolean> toBoolean = value -> {
         if ("Yes".equalsIgnoreCase(value) || "y".equalsIgnoreCase(value)) return true;
+        if ("TRUE".equalsIgnoreCase(value) || "T".equalsIgnoreCase(value)) return true;
+        if ("1".equals(value)) return true;
+
         if ("No".equalsIgnoreCase(value) || "n".equalsIgnoreCase(value)) return false;
+        if ("FALSE".equalsIgnoreCase(value) || "F".equalsIgnoreCase(value)) return false;
+        if ("0".equals(value)) return false;
```
agorina: Tabulator output is not consistent with handling boolean values. Added support for different options.

package-processor/src/main/resources/application.sql.yml
```diff
@@ -92,12 +92,6 @@ sql:
          SELECT (SELECT id AS accommodation_id FROM accommodation WHERE code = :accCode), :langId, :label
            ON DUPLICATE KEY UPDATE label=:label
 
-    findByAccCodeAndLangCode: >-
-      SELECT id FROM accommodation_translation WHERE accommodation_id=:accId AND language_id=:langId
-
-    updateTranslation: >-
-      UPDATE accommodation_translation SET label=:label WHERE accommodation_id=:accId AND language_id=:langId
-
```
agorina: Unrelated: why looking at the accommodations I noticed that this is not used.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -21,6 +21,12 @@
     private final double maxPoints;
     private final Set<Integer> commonCoreStandards;
     private final Set<Integer> otherTargets;
+    private final Boolean isFieldTest;
+    private final Boolean isActive;
+    private final Integer optionsCount;
+    private final String type;
+    private final String answerKey;
```
malaffoon: How are we handling the fact that these fields aren't really optional but won't be present for loaded packages? Are they just going to be optional throughout the system?

agorina: The ingest enforces the rules (see below), but the Db does not enforce the data. 
Maybe we should have a JIRA to follow up and update Db constraints after we backfill the data?

 Here are the rules based on my understanding: 
- isActive and isFieldTest are optional since tabulator outputs an empty field for it. 
- optionsCount is a mandatory int.
- type is required to be not blank
- answerKey is conditionally required for type = 'MS' or 'MC'.

PR #397, opened 2017-10-31
use fuzzy compare with SQL float values

malaffoon:


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/MathUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     * Although the database should be storing 23 bits of precision, when it goes through the JDBC
+     * call stack it seems to be rounded aggressively at 6 significant decimal digits.
+     * </p>
+     *
+     * @param v1 value
+     * @param v2 value
+     * @return true if values are the same to SQL float precision
+     */
+    public static boolean isCloseSqlFloat(final double v1, final double v2) {
+        return isClose(v1, v2, 1e-5, 0);
```
malaffoon: I found this kind of surprising. I fully expected the SQL client to round decimal values but i really thought that the JDBC call stack would preserve more precision.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
```diff
@@ -239,56 +238,6 @@ public Double getThetaScoreStdErr() {
...
     }
 
-    /**
-     * Test for shallow intrinsic-value equivalence with the given `
-     * This does not test child models.
-     *
-     * @param other An Exam
-     * @return True if the other Exam has identical intrinsic values
-     */
-    public boolean equivalent(final Exam other) {
```
malaffoon: I chose to remove these helper methods from the POJO. They were being used only in the JDBC repo classes so i moved all the functionality there.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/MathUtilsIT.java
```diff
@@ -0,0 +1,33 @@
...
+            "INSERT INTO testFloat (scoref, scored) VALUES (2345.1234567, 2345.1234567), (.1234567, .1234567), (2.1234567, 2.1234567)"
+    })
+    @Sql(executionPhase = Sql.ExecutionPhase.AFTER_TEST_METHOD, statements = {
+            "DROP TABLE testFloat"
+    })
+    public void isCloseFloat() {
+        final List<Float> values = jdbcTemplate.queryForList("SELECT scoref FROM testFloat", Float.class);
+        assertThat(MathUtils.isCloseSqlFloat(2345.1234567, values.get(0))).isTrue();
+        assertThat(MathUtils.isCloseSqlFloat(.1234567, values.get(1))).isTrue();
+        assertThat(MathUtils.isCloseSqlFloat(2.1234567, values.get(2))).isTrue();
```
malaffoon: If i change these to MathUtils.isCloseFloat, they fail. That's why i made isCloseSqlFloat and reduced its precision.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/MathUtils.java
jtreuting: does this depend on the driver?  meaning if we switch that driver at some point (as we thought of doing based on suggestions for Aurora) that we might cause an unintended bug later on?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Exam.java
agorina: 👍

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/MathUtils.java
malaffoon: I doubt switching drivers would make a big difference, but i guess it could. At worst we would think two values were the same when they really do vary at < 1e-5 precision. Since the SB scoring system is really only good to 4-5 significant digits we'd still be okay.

It does (re)raise the issue @chdietz brought up ... should we consider switching all the float fields to double? Then we could probably compare down to 1e-12 or so.

agorina: What is the disadvantage of switching to double? 
I guess what I am trying to ask is why don't we do it?

chdietz: I think based on what Mark mentioned is after 4-5 precision the data passed in the TRT is not valuable.  I hat to say junk, but I just did.

chdietz: @malaffoon are you suggesting(or have you tested) if we switch to double the JDBC call stack will preserve more precision?

chdietz: is it JDBC ? eve from command line it looks the same:
```mysql
mysql> UPDATE exam e SET e.scale_score = 2345.1234567 WHERE e.id = 1;
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select e.scale_score from exam e where e.id = 1;
+-------------+
| scale_score |
+-------------+
|     2345.12 |
+-------------+
1 row in set (0.00 sec)

mysql> UPDATE exam e SET e.scale_score = 2.1234567 WHERE e.id = 1;
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select e.scale_score from exam e where e.id = 1;
+-------------+
| scale_score |
+-------------+
|     2.12346 |
+-------------+
1 row in set (0.00 sec)

```

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/MathUtilsIT.java
chdietz: Nice test!

PR #399, opened 2017-11-02
DWR-1131 Add findOneByNaturalId to AssessmentPackageRepository 

agorina:
This is done in preparation for supporting asmt updates/

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
@@ -1,0 +1,34 @@
+package org.opentestsystem.rdw.ingest.common.util;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+
+/**
+ * Holds common methods for when working with JDBC
+ */
+public class JdbcUtil {
+    /**
+     * Returns the given value or null if the last column lookup resulted in a null value.
+     * This is a workaround for two things: typecasting the return value of {@link ResultSet#getObject(String)} for getNullable columns
+     * <ol>
+     * <li>{@link ResultSet#getLong(String)} and other primitive type lookups will turn null into the default values
+     * (e.g. null boolean will be false, null long will be 0)</li>
+     * <li>Type casting the result of {@link ResultSet#getLong(String)} can result in type cast exceptions if the
+     * database driver returns an Integer and you cast to Long.</li>
+     * </ol>
+     *
+     * @param resultSet the row of data which will be checked for null
+     * @param value     the result of the row lookup (e.g. resultSet.getLong("column_name"))
+     * @param <T>       the type of the provided value
+     * @return the value or null if the column lookup was null
+     * @throws SQLException if a database access error occurs or this method is called on a closed result set
+     */
+    public static <T> T getNullable(final ResultSet resultSet, final T value) throws SQLException {
+        return resultSet.wasNull() ? null : value;
+    }
+
+
+    private JdbcUtil() {
+    }
+}
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
@@ -23,8 +23,6 @@ import org.springframework.retry.annotation.Retryable;
 import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Transactional;
 
-import java.sql.ResultSet;
-import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -37,6 +35,7 @@ import java.util.stream.Collectors;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.opentestsystem.rdw.ingest.common.repository.JdbcUtils.safeText;
+import static org.opentestsystem.rdw.ingest.common.util.JdbcUtil.getNullable;
 import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.Conventions;
 import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.EvidenceAndElaboration;
 import static org.opentestsystem.rdw.ingest.processor.model.ExamItem.OrganizationAndPurpose;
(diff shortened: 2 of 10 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
```diff
@@ -0,0 +1,34 @@
...
+     * @throws SQLException if a database access error occurs or this method is called on a closed result set
+     */
+    public static <T> T getNullable(final ResultSet resultSet, final T value) throws SQLException {
+        return resultSet.wasNull() ? null : value;
+    }
+
+
+    private JdbcUtil() {
+    }
+}
```
agorina: This probably belongs to RDW_Common, but...

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -674,24 +673,4 @@ private static Instant toInstantOrNull(final Timestamp timestamp) {
...
-     * @param resultSet the row of data which will be checked for null
-     * @param value     the result of the row lookup (e.g. resultSet.getLong("column_name"))
-     * @param <T>       the type of the provided value
-     * @return the value or null if the column lookup was null
-     * @throws SQLException if a database access error occurs or this method is called on a closed result set
-     */
-    private static <T> T getNullable(final ResultSet resultSet, final T value) throws SQLException {
-        return resultSet.wasNull() ? null : value;
-    }
 }
```
agorina: Move to the Ingest Common.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Accommodation.java
```diff
@@ -1,19 +1,15 @@
...
  * AccessibilityCode - model of an accessibility code and its language translations to assist in report generation
  */
 public class Accommodation {
-    private final String code;
-    private final Map<Integer, String> translations;
-
-    private Accommodation(Builder builder) {
-        this.code = builder.code;
-        this.translations = builder.translations;
-    }
```
agorina: No real change, just made this consistent with the rest of the system.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -100,21 +106,19 @@ public Double getCutPoint3() {
...
         private String naturalId;
         private String name;
         private String label;
         private Integer schoolYear;
         private String version;
-        //The following are validated against translation tables in the database
         private Integer gradeId;
         private Integer typeId;
         private Integer subjectId;
-        //Not checked against database, but required to exist
```
agorina: Since the validation is done in the service layer I did not feel that the comment is appropriate here (and above).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Item.java
```diff
@@ -10,45 +9,26 @@
...
-        this.commonCoreStandards = builder.commonCoreStandards;
-        this.otherTargets = builder.otherTargets;
-        this.isFieldTest = builder.isFieldTest;
-        this.isActive = builder.isActive;
-        this.optionsCount = builder.optionsCount;
-        this.type = builder.type;
-        this.answerKey = builder.answerKey;
-    }
-
-    public static Builder builder() {
```
agorina: Once again, no real change - just made it consistent with the rest of the system.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
```diff
@@ -38,6 +44,18 @@
     @Value("${sql.assessment.create}")
     private String sqlAssessmentCreate;
 
+    @Value("${sql.assessment.findOneByNaturalId}")
+    private String slqFindOneByNaturalId;
```
jtreuting: typo: slq

agorina: LOL - thank you, fixing.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
```diff
@@ -86,6 +104,110 @@ public void createPackage(final Collection<Assessment> assessments, final long i
         if (!assessmentNaturalIds.isEmpty()) jdbcTemplate.update(sqlUpdateDifficultyCode, new MapSqlParameterSource("asmt_natural_ids", assessmentNaturalIds));
     }
 
+    @Transactional(readOnly = true)
+    @Override
+    public Assessment findOneByNaturalId(final String id) {
+
+        final List<Assessment.Builder> assessmentList = jdbcTemplate.query(slqFindOneByNaturalId, new MapSqlParameterSource("natural_id", id), new AssessmentRowMapper());
+        if (assessmentList.size() == 1) {
```
jtreuting: should it throw if it gets more than 1 instead of returning null?  if there are 0 then of course null, but if there are 2?

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
jtreuting: agreed

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
agorina: Well, I guess I know that it is UNIQUE in the DB...And if it is not unique,  a lot of things would go wrong. But then from this code point of view I can check for more than 1 and throw an exception. Will do.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
malaffoon: 👍

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
```diff
@@ -86,6 +104,110 @@ public void createPackage(final Collection<Assessment> assessments, final long i
...
+                        final int itemId = results.getInt("item_id");
+                        if (idsByItemIds.containsKey(itemId)) idsByItemIds.get(itemId).add(results.getInt("common_core_standard_id"));
+                        else idsByItemIds.put(itemId, newArrayList(results.getInt("common_core_standard_id")));
+
+                    }
+                    return idsByItemIds;
+                });
+    }
+
+    private Map<Integer, List<Integer>> findOtherTargets(final Integer assessmentId) {
```
malaffoon: an unfortunate name; i don't know enough yet to suggest a better one

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
malaffoon: I waffled on this one too. I think it should stay as is. There is NO way to generate the error that you are checking for (without changing the schema).

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
malaffoon: If you do that, please make a new module in RDW_Common (i.e. don't roll it into the existing Utils module).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepository.java
jtreuting: For some reason GitHub was hiding this from me.  Since the DB has a unique constraint and we aren't worried about having 2, shouldn't the check be `if (!assessmentList.isEmpty())` (or check for empty and return null first).  Since you don't really care about the size of the list just that something exists.  It just feels odd to check for exactly 1 when if it's 2 not doing anything about it.  So seems like an existence check makes more since and is microscopically faster :)

Not a big deal though.

agorina: Makes sense. Will adjust with the next change.

PR #402, opened 2017-11-05
DWR-1131 Assessment upsert

agorina:
This is the final step for this JIRA.


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/JdbcUtil.java
```diff
@@ -28,6 +28,16 @@
...
+    /**
+     * Compares two Double values using {@link MathUtils#isCloseSqlFloat}
+     *
+     * @param v1 one value to compare
+     * @param v2 another value to compare
+     * @return true if equivalent, false otherwise
+     */
+    public static boolean equivalent(final Double v1, final Double v2) {
+        return (v1 == null && v2 == null) || (v1 != null && v2 != null && MathUtils.isCloseSqlFloat(v1, v2));
+    }
```
agorina: Since it is now needed in more than one place I moved it here. Not sure if this is the best home for it.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -300,56 +300,56 @@ private void updateExam(final Exam exam, final long examId, final long importId)
     // shallow equals check on Exam
     private static boolean equivalent(final Exam e1, final Exam e2) {
```
agorina: The formatting was not intentional. Hmmm..I wonder how it happened. I should have the same IntelliJ settings as the rest of the team.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,87 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+            final Assessment existing = packageRepository.findOneByNaturalId(toUpsert.getNaturalId());
+
+            if (existing == null) {
+                newAssessments.add(toUpsert);
+            } else if (validToUpdate(existing, toUpsert)) {
+                final Assessment assessmentToUpdate = processAssessmentForUpdate(existing, toUpsert);
+                if (assessmentToUpdate != null) updatedAssessments.add(assessmentToUpdate);
+            } else {
+                rejectedAssessments.add(toUpsert);
+                // TODO: not sure what to do here ?
```
agorina: Not sure if I should do anything else here. I am afraid that it may be a silent error that is easy to miss.

package-processor/src/main/resources/application.sql.yml
```diff
@@ -28,6 +28,8 @@ sql:
           a.label = :label,
           a.version = :version,
           a.update_import_id = :update_import_id
+        WHERE a.id = :id
```
agorina: This qualifies for a stupidest mistake of the year. I have no idea what I was thinking...but I am glad one of the tests discovered it.

I would think that JDBCTemplate should be smart enough to complain if a parameter is not used in the query, but it does not.

package-processor/src/main/resources/application.sql.yml
```diff
@@ -44,6 +46,7 @@ sql:
              s.cut_point_3 = :cutPoint3,
              s.min_score = :minScore,
              s.max_score = :maxScore
+          WHERE s.asmt_id = :asmt_id
```
agorina: And this one.

package-processor/src/main/resources/application.sql.yml
```diff
@@ -70,6 +73,7 @@ sql:
              i.type = :type,
              i.options_count= :optionsCount,
              i.answer_key = :answerKey
+          WHERE i.id = :id
```
agorina: And one more.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcAssessmentPackageRepositoryIT.java
```diff
@@ -257,7 +256,23 @@ public void itShouldUpdate() {
         assertThat(item990.getFieldTest()).isEqualTo(true);
         assertThat(item990.getAnswerKey()).isEqualTo("new answerKey");
         assertThat(item990.getAllowCalculator()).isNull();
+        assertThat(item990.getMathPractice()).isEqualTo(1);
         assertThat(item990.getCommonCoreStandards()).containsExactlyInAnyOrder(-99, -98);
         assertThat(item990.getOtherTargets()).containsExactlyInAnyOrder(-1111);
+
+        final Item anotherItem = items.stream().filter(i -> i.getId() != -990).findFirst().get();
```
agorina: Added more tests.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessorIT.java
```diff
@@ -33,14 +34,51 @@
...
-
+    public void itShouldCreate() throws IOException {
         final int beforeAssessmentCount = countRowsInTable(jdbcTemplate, "asmt") + countRowsInTable(jdbcTemplate, "asmt");
         final int beforeItemCount = countRowsInTable(jdbcTemplate, "item") + countRowsInTable(jdbcTemplate, "item");
 
-        assertThat(packageProcessor.process(ByteStreams.toByteArray(this.getClass().getResourceAsStream("/IAB_ICA_Combined.items.csv")), -99))
-                .isEqualTo("9 assessments processed");
+        assertThat(packageProcessor.process(ByteStreams.toByteArray(this.getClass().getResourceAsStream("/updateAsmt/IAB_ICA_Combined.items.csv")), -99))
+                .isEqualTo("Total assessments processed 9, newly created count 9, updated count 0, rejected count 0");
         assertThat(countRowsInTable(jdbcTemplate, "asmt")).isEqualTo(beforeAssessmentCount + 9);
```
agorina: Any concerns that I am changing the message here? 
Would we ever need to query this msg in the db?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+                if (!updatedItem.equals(existingItem)) {
+                    items.add(updatedItem);
+                    update = true;
+                }
+            } else {
+                items.add(itemToUpsert);
+                update = true;
+            }
+        }
+        return update ? asmtBuilder.items(items).build() : null;
```
agorina: Just to note: if there are 'removed' items they will not be in the built assessment. Since the repo does not delete, it works. I am not sure if it would be 'cleaner' to add the items here?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+        packageRepository.create(newAssessments, importId);
+        final String msg = "Total assessments processed " + assessments.size() + ", newly created count " + newAssessments.size() + ", updated count " + updatedAssessments.size() + ", rejected count " + rejectedAssessments.size();
         logger.info(msg);
         return msg;
     }
 
+    private boolean validToUpdate(final Assessment existing, final Assessment toUpsert) {
+        return existing.getSubjectId() == toUpsert.getSubjectId()
+                && existing.getGradeId() == toUpsert.getGradeId()
+                && existing.getTypeId() == toUpsert.getTypeId();
```
agorina: I added the asmt type to the validation rules(JIRA mentions subject and grade per asmt id). When we parse TRTs we take the type from the asmt so I felt it is right to include it here.

Any objections?

malaffoon: I guess the obvious question is: why is type_id denormalized in the exam table?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+    }
+
+    /**
+     * Combines existing assessment with the newly provided one.
+     * The rules for building the assessment are:
+     * - the assessment must be updated if at least one field has changed. This is done to update the import id and 'updated' timestamp for migrate to pick up the change.
```
malaffoon: nit: javadocs should use proper markup (i.e. <ul><li>); also, very long lines here

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+                }
+            } else {
+                items.add(itemToUpsert);
+                update = true;
+            }
+        }
+        return update ? asmtBuilder.items(items).build() : null;
+    }
+
+    private boolean asmWithoutItemsEqual(final Assessment one, final Assessment another) {
```
malaffoon: typo in name? `asmtWithoutItemsEqual` although maybe `equalsIgnoringItems` (since it obviously deals with asmts and i like the `equals` at the beginning of the name.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+     * - if an existing assessment has more items, they should be preserved, we do not allow for items to be deleted in a case there are exams that are linked to it
+     */
+    private Assessment processAssessmentForUpdate(final Assessment existing, final Assessment toUpsert) {
+        boolean update = !asmWithoutItemsEqual(existing, toUpsert);
+
+        final Assessment.Builder asmtBuilder = Assessment.builder()
+                .copyWithoutItems(toUpsert)
+                .id(existing.getId());
+
+        final Map<String, Item> existingItemsById = existing.getItems().stream().collect(toMap(Item::getNaturalId, i -> i));
```
malaffoon: I think building this map and using it will be less efficient (and certainly is less readable) than adding a method on Assessment like `findItem(final String itemNaturalId)`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -36,19 +40,82 @@ public String process(final byte[] assessmentPackage, final long importId) {
...
+        final List<Item> items = newArrayList();
+
+        for (final Item itemToUpsert : toUpsert.getItems()) {
+            if (existingItemsById.containsKey(itemToUpsert.getNaturalId())) {
+                final Item existingItem = existingItemsById.get(itemToUpsert.getNaturalId());
+                final Item updatedItem = Item.builder().copy(itemToUpsert).id(existingItem.getId()).build();
+                if (!updatedItem.equals(existingItem)) {
```
malaffoon: I think there may be too much knowledge of what the repository is doing here, maybe we can talk about it. I might feel better if the javadocs explicitly said something like: `This method constructs an assessment that has just the updated parts so the repository can efficiently update an existing assessment. Yes, it has deep knowledge of how the repository works and the efficiency logic is mostly here not in the repository.`

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessorIT.java
malaffoon: i don't think we need to worry about changing the message 

I'd request a slight change to the message to make it shorter, something like "Assessments processed: 9, created: 9, updated: 0, rejected: 0"

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
agorina: We used to have two tables - one of iab and another for ICA/Summative. At the time we merged two tables into one it looked like a good idea to have type_id as a discriminator. I will check how and where it is used and we can discuss changing it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
agorina: I like this idea, will change.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
agorina: I share the same concern. Let's talk tomorrow.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
agorina: Thanks.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessorIT.java
agorina: Will change.

PR #404, opened 2017-11-06
DWR-1132 Ingest: support accommodations updates, remove language rest…

agorina:
…rictions


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAccommodationsProcessor.java
```diff
@@ -99,16 +96,15 @@ public String process(final byte[] accessibilityConfig, final long importId) thr
...
-                accommodations.add(Accommodation.builder()
-                        .code(code)
-                        .addTranslation(langId, label)
-                        .build());
-                logger.debug(code + " = [" + lang + "] " + label);
+                builder.addTranslation(langCode, label);
+
+                logger.debug(code + " = [" + langCode + "] " + label);
             }
+            accommodations.add(builder.build());
```
agorina: This was really wrong: it caused the same accommodation to be inserted multiple times in a case when there is more than one translation. Which in turn caused the accommodation to change its auto-incremented id. I added tests to prove that it is fixed.

PR #405, opened 2017-11-08
fiddle with student groups with bad years

malaffoon:
As part of trying to track this down, i created this test. Doesn't really help with the overall issue but i thought i'd preserve it ...

### DIFF
--- a/group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
+++ b/group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
@@ -7,6 +7,8 @@ import org.opentestsystem.rdw.ingest.group.RepositoryBackedIT;
 import org.opentestsystem.rdw.ingest.group.service.ProcessingGroupImportService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Import;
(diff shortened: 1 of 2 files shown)


group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
```diff
@@ -305,6 +308,48 @@ public void itShouldTriggerAMigration() {
...
+        // This test plays with the situation where an invalid school year was set for a group
+        // (the CSV validation has been improved so this shouldn't happen any more but ...)
+        //
+        // The problem is that `insertMissingGroups` silently fails to insert the missing group
+        // because of a data integrity violation. After that, other steps no-op because they are
+        // joining to the student_group table which doesn't have an entry.
+        //
+        // All that said, the worst that will happen is the import will be marked PROCESSED which
+        // might be misleading since nothing was imported.
+
```
chdietz: So, one case where the import can be left ACCEPTED is this case and this exercises it but is covered by CSV validation?

malaffoon: Unfortunately, no. This should not leave the import in ACCEPTED state because, as this test demonstrates, it gets through the whole process without throwing an exception or otherwise veering off the happy path.

PR #408, opened 2017-11-14
Feature/cleanup group membership processing

wtritch:
This cleans up group membership processing to only clear removed students/users from membership tables when processing a group update.
The "INSERT IGNORE" and unique index will ensure that only new users/students are inserted into a membership table.


wtritch:
hm... looks like I borked the branching.  Aborting this PR and will file a clean one.

PR #409, opened 2017-11-14
Cleanup group membership processing to only clear removed students/us…

wtritch:
…ers from groups.
This cleans up group membership processing to only clear removed students/users from membership tables when processing a group update.
The "INSERT IGNORE" and unique index will ensure that only new users/students are inserted into a membership table.

### DIFF
--- a/group-processor/src/main/resources/application.sql.yml
+++ b/group-processor/src/main/resources/application.sql.yml
@@ -396,7 +396,11 @@ sql:
                           AND upload.batch_id = :batch_id
                           AND upload.school_id = :school_id) updated
                 ON updated.group_id = users.student_group_id
+              LEFT JOIN upload_student_group uploaded_user
+                ON uploaded_user.group_id = users.student_group_id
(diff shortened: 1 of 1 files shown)


group-processor/src/main/resources/application.sql.yml
```diff
@@ -396,7 +396,11 @@ sql:
                           AND upload.batch_id = :batch_id
                           AND upload.school_id = :school_id) updated
                 ON updated.group_id = users.student_group_id
+              LEFT JOIN upload_student_group uploaded_user
+                ON uploaded_user.group_id = users.student_group_id
+                  AND uploaded_user.group_user_login = users.user_login
             WHERE users.student_group_id IS NOT NULL
+              AND uploaded_user.id IS NULL
```
wtritch: This basically adds an additional lookup to only remove users that don't exist in the uploaded members list.

chdietz: @wtritch is thinking in sets

PR #415, opened 2017-11-20
DWR-1145, deal with nullable values (migrant status and section 504) 

malaffoon:
For the nullable attributes, the row mapper was using row.getBoolean() which returns false for a null value. During update, it was comparing the incoming value (null) against the existing value (null -> false) and null != false so it was always updating.

This fixes the row mapper.


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -433,7 +433,7 @@ public void itShouldModifyAChangedExamStudent() {
                         .ideaIndicator(true)
                         .languageCode("CDE")
                         .lep(false)
-                        .migrantStatus(false)
+                        .migrantStatus(null)
```
malaffoon: oops, i think this is an unintentional change ... sec

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryIT.java
```diff
@@ -449,6 +449,45 @@ public void itShouldModifyAChangedExamStudent() {
...
+                "SELECT * FROM exam WHERE id = :id",
+                new MapSqlParameterSource().addValue("id", examId));
+
+        final Exam updated = builder.build();
+        assertThat(repository.upsert(updated, -2)).isEqualTo(examId);
+        final Map<String, Object> updatedValues = namedParameterTemplate.queryForMap(
+                "SELECT * FROM exam WHERE id = :id",
+                new MapSqlParameterSource().addValue("id", examId));
+
+        assertThat(updatedValues.get("updated")).isEqualTo(originalValues.get("updated"));
```
malaffoon: before the fix, this assertion failed

PR #417, opened 2017-11-30
newline handling

malaffoon:
I'd like to push this and try it in awsqa before deploying to production configuration

### DIFF
--- a/group-processor/src/main/resources/application.sql.yml
+++ b/group-processor/src/main/resources/application.sql.yml
@@ -463,14 +463,23 @@ sql:
 
 
   # ------------ Load from S3 or local ---------------------------------------------------------
+  # Line termination. Recall ...
+  #   '\r' == CR == 0x0D : default for Mac
+  #   '\n' == LF == 0x0A : default for Linux
+  #   '\r\n' : default for Windows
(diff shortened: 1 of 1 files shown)


group-processor/src/main/resources/application.sql.yml
```diff
@@ -463,14 +463,24 @@ sql:
 
 
   # ------------ Load from S3 or local ---------------------------------------------------------
+  # Line termination. Recall ...
+  #   '\r' == CR == 0x0D : default for Mac
```
malaffoon: Hmm. I have found evidence that the default for Mac may be LF. Sigh. Anybody have any definitive answer?

FYI, Melissa's files have CR as the newline. Of course, that is Excel 2011 which might be from before the switch.

Grrrrr.

jtreuting: i can install Mac Office 2011 tonight and test that versus the 2016 version that I have and assuming you have

jtreuting: here is what i found with Mac Excel Office 365

Save As...
- MS-DOS CSV: CRLF
- Windows CSV: CRLF
- Generic CSV: CRLF
- CSV UTF8 (default CSV it seems): CR

PR #419, opened 2017-12-05
use delimiter and newline from file properties

malaffoon:
Read delimiter and newline from file properties. There is a matching PR in the RDW_Admin project.

I need to do more testing but i wanted eyes/brains on the approach for sanity check.


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -63,30 +62,39 @@ public void loadBatch(final String digest, final long batchId) {
...
+        String delimiter = props.getProperty("delimiter");
+        if (delimiter == null) {
+            logger.warn("Delimiter unspecified for {}, using default ','", location);
+            delimiter = ",";
+        }
 
+        String newline = props.getProperty("newline");
+        if (newline == null) {
+            logger.warn("Newline sequence unspecified for {}, using default LF", location);
+            newline = "\\n";
```
malaffoon: I chose to default to LF because it is the least messy of the two choices. I'm hopeful this will never happen but i wanted to code defensively.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -439,15 +439,17 @@ sql:
...
   load-csv: >-
-    LOAD DATA %s :uri INTO TABLE upload_student_group FIELDS TERMINATED BY ','
+    LOAD DATA %s :uri INTO TABLE upload_student_group
+    FIELDS TERMINATED BY '%s'
+    LINES TERMINATED BY '%s'
     IGNORE 1 LINES
     (@vgroup_name, @vschool_natural_id, school_year, @vsubject_code, @vstudent_ssid, @vgroup_user_login)
     SET
       batch_id = :batch_id,
       group_name = TRIM(REPLACE(@vgroup_name, '\r', '')),
```
malaffoon: Because it will default to LF, i'm leaving this REPLACE in so that a CRLF file will be handled okay.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
jtreuting: might be overkill, but should the delimiter and newline be whitelisted so we know they won't break the SQL?

malaffoon: @jtreuting you mean have a list of allowed values?

jtreuting: i don't have a list of all possible values but we know some values will break the SQL right?  it just feels like a SQL injection possibility here.  it might not in reality be possible because we are doing this based on the a file upload and setting properties, but if those properties got corrupted somehow i'm assuming there could be some harm done like classic SQL injection with ending the current statement and doing a delete all type thing.

that at least is what i was thinking.  we know we want to support CR, CRLF and LF, so should we allow it to be anything else?

group-processor/src/main/resources/application.sql.yml
malaffoon: Hmm, i think this is backwards ... adding tests.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -63,30 +65,39 @@ public void loadBatch(final String digest, final long batchId) {
...
+        // The sql statement is parameterized to allow S3 or local file syntax to be injected,
+        // as well as the delimiter and newline sequence. Fetch the sql statement, substitute
+        // the source, delimiter and newline, then pass it off to the repository to be executed.
+        final Properties props = archiveService.readProperties(location);
 
-        // The sql statement is string parameterized to choose between S3 syntax and local file syntax
-        // We want to get the sql statement, substitute in the right source, and then pass the sql statement
-        //  off to the repository to be executed.
+        String delimiter = props.getProperty("delimiter");
+        if (!ValidDelimiter.contains(delimiter)) {
```
malaffoon: @jtreuting

group-processor/src/main/resources/application.sql.yml
```diff
@@ -439,15 +439,17 @@ sql:
 
   # ------------ Load from S3 or local ---------------------------------------------------------
   load-csv: >-
-    LOAD DATA %s :uri INTO TABLE upload_student_group FIELDS TERMINATED BY ','
+    LOAD DATA %s :uri INTO TABLE upload_student_group
+    FIELDS TERMINATED BY '%s' OPTIONALLY ENCLOSED BY '"'
```
malaffoon: almost missed putting in the configuration fix we jacked in a while ago

group-processor/src/main/resources/application.sql.yml
```diff
@@ -439,15 +439,17 @@ sql:
...
       batch_id = :batch_id,
-      group_name = TRIM(REPLACE(@vgroup_name, '\r', '')),
-      school_natural_id = TRIM(REPLACE(@vschool_natural_id, '\r', '')),
-      subject_code = nullif(TRIM(REPLACE(@vsubject_code, '\r', '')), ''),
-      student_ssid = nullif(TRIM(REPLACE(@vstudent_ssid, '\r', '')), ''),
+      group_name = TRIM(@vgroup_name),
+      school_natural_id = TRIM(@vschool_natural_id),
+      subject_code = nullif(TRIM(@vsubject_code), ''),
+      student_ssid = nullif(TRIM(@vstudent_ssid), ''),
       group_user_login = nullif(TRIM(REPLACE(@vgroup_user_login, '\r', '')), '')
```
malaffoon: This is to catch CRLF if the default newline (LF) is used. In that case, the CR will tag along on the value from the last field so we have to replace it.

PR #425, opened 2017-12-20
a bug with safeText

malaffoon:


chdietz:
How many of these greater than 65K did we hit in prod?
How many close to this size?
I am wondering what is generating such a long response? somebody left their hand on the keyboard?

malaffoon:
In production there are currently 21.5 million item responses, 9 of which exceeded 64k. And 1 of those had the multibyte character. The top 20 response lengths get down to 30k and the top 100 get down to "normal" lengths.

I think there is something borked in the editor that allows this to happen but, yeah, i'm assuming this is a kid leaning on the spacebar or just intentionally holding it down.

malaffoon:
I spent a while trying to do something similar using only SQL. Can't figure it out since all SQL commands are multibyte aware and don't let us get at the underlying data buffer. For example if i do something like this: 
```
UPDATE table SET value = LEFT('really long string with multibyte characters ...', 65535);
```
where table.value is a TEXT field, it will blow up. LEFT is multibyte aware so it truncates the string to 65535 *characters* which is more than 65535 *bytes* (TEXT field limit). 

I really want to have MySQL treat TEXT as VARCHAR(65535) but haven't been able to figure out how to do that.

chdietz:
I thought I read there is a setting that issues a warning and truncates when TEXT is assigned a value too large

PR #426, opened 2017-12-21
Feature/dwr 1215

agorina:
Task list:
- [x] combine ica and summative into one fact table 
- [x] change ica_asmt to just asmt and include all asmt types 
- [x] add asmt labels 
- [x] add school_year 
- [x] convert active_asmt into a table and load it during the migrate 
- [ ] account for the latest exam per student/school year/asmt type
- [ ] handle non-scored tests: do not load them into redshift
- [ ] add school and district groups to support permissions
- [ ] migrate embargo data
- [ ] and anything else..?

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSqlListExecutionRepository.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSqlListExecutionRepository.java
@@ -18,7 +18,7 @@ public class JdbcSqlListExecutionRepository implements SqlListExecutionRepositor
         this.jdbcTemplate = jdbcTemplate;
     }
 
-   @Transactional
+    @Transactional
     @Override
     public void execute(final Iterable<String> sqls) {
         final JdbcOperations operations = jdbcTemplate.getJdbcOperations();
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
@@ -139,8 +139,8 @@ public class OlapReportingMigrateStepsConfig {
     @Bean
     public Step deriveActiveAsmtStep() {
         final SqlListBuilder sqlsBuilder = new SqlListBuilder(stagingToReportingSqlConfiguration.getEntities())
-                .addNext("active_asmt_per_year", "delete")
-                .addNext("active_asmt_per_year", "insert");
+                .addNext("asmt_active_year", "delete")
+                .addNext("asmt_active_year", "insert");
 
         return stepBuilderFactory.get(deriveActiveAsmtStep)
                 .tasklet(new SqlListCodesExecutionStep(
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
@@ -33,21 +33,22 @@ public class OlapReportingMigrateStepsConfig {
     private final MigrateStagingToReportingSqlConfiguration stagingToReportingSqlConfiguration;
     private final MigrateWarehouseToStagingSqlConfiguration warehouseToStagingSqlConfiguration;
 
-    public static final String stepTruncateStageName = "stepTruncateStage";
-    public static final String extractEntitiesStep = "extractEntitiesStep";
-    public static final String extractCodesStep = "extractCodesStep";
-    public static final String loadCodesToStagingStep = "loadCodesToStagingStep";
-    public static final String loadEntitiesToStagingStep = "loadEntitiesToStagingStep";
+    static final String stepTruncateStageName = "stepTruncateStage";
+    static final String extractEntitiesStep = "extractEntitiesStep";
+    static final String extractCodesStep = "extractCodesStep";
+    static final String loadCodesToStagingStep = "loadCodesToStagingStep";
+    static final String loadEntitiesToStagingStep = "loadEntitiesToStagingStep";
+    static final String deriveActiveAsmtStep = "deriveActiveAsmtStep";
 
-    public static final String deleteCodesStepName = "deleteCodesStepName";
-    public static final String upsertCodesStepName = "upsertCodesStepName";
-    public static final String deleteEntitiesStepName = "deleteEntitiesStep";
-    public static final String upsertStudentsStepName = "upsertStudentsStep";
-    public static final String upsertOrganizationsStepName = "upsertOrganizationsStep";
-    public static final String upsertAsmtsStepName = "upsertAsmtsStep";
-    public static final String upsertFactStudentIcaExamStepName = "upsertFactStudentIcaExam";
+    static final String deleteCodesStepName = "deleteCodesStep";
+    static final String upsertCodesStepName = "upsertCodesStep";
+    static final String deleteEntitiesStepName = "deleteEntitiesStep";
+    static final String upsertStudentsStepName = "upsertStudentsStep";
+    static final String upsertOrganizationsStepName = "upsertOrganizationsStep";
+    static final String upsertAsmtsStepName = "upsertAsmtsStep";
+    static final String upsertFactStudentExamStepName = "upsertFactStudentExamStep";
 
-    public static final List<String> codes = newArrayList("administration_condition", "completeness", "ethnicity", "gender", "grade");
+    public static final List<String> codes = newArrayList("administration_condition", "completeness", "ethnicity", "gender", "grade", "school_year");
     static final List<String> warehouseEntities = newArrayList("asmt", "school", "district", "student", "student_ethnicity", "exam", "exam_claim_score");
 
     @Autowired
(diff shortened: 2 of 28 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/repository/JdbcSqlListExecutionRepository.java
```diff
@@ -18,7 +18,7 @@ public JdbcSqlListExecutionRepository(final NamedParameterJdbcTemplate jdbcTempl
         this.jdbcTemplate = jdbcTemplate;
     }
 
-   @Transactional
+    @Transactional
```
agorina: unrelated.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -33,21 +33,22 @@
...
+    static final String loadCodesToStagingStep = "loadCodesToStagingStep";
+    static final String loadEntitiesToStagingStep = "loadEntitiesToStagingStep";
+    static final String deriveActiveAsmtStep = "deriveActiveAsmtStep";
+
+    static final String deleteCodesStepName = "deleteCodesStep";
+    static final String upsertCodesStepName = "upsertCodesStep";
+    static final String deleteEntitiesStepName = "deleteEntitiesStep";
+    static final String upsertStudentsStepName = "upsertStudentsStep";
+    static final String upsertOrganizationsStepName = "upsertOrganizationsStep";
+    static final String upsertAsmtsStepName = "upsertAsmtsStep";
```
agorina: Fixed IntelliJ warning.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/SqlListWithParamExecutionStep.java
```diff
@@ -35,9 +32,4 @@ public RepeatStatus execute(final StepContribution contribution, final ChunkCont
         repository.execute(sqls, migrate);
         return RepeatStatus.FINISHED;
     }
-
-    @VisibleForTesting
-    Iterable<String> getSqls() {
-        return newArrayList(sqls);
-    }
```
agorina: This was not used.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -175,19 +193,20 @@ sql:
...
                   sa.subject_id,
                   sa.name,
+                  sa.label,
                   sa.type_id,
                   sa.update_import_id,
                   sa.migrate_id
                 FROM staging_asmt sa
                   LEFT JOIN asmt ra ON ra.id = sa.id
-                WHERE ra.id IS NULL and sa.deleted = 0 AND sa.type_id = 1;
+                WHERE ra.id IS NULL and sa.deleted = 0;
```
agorina: I removed this `and` because it seemed redundant. The script that extracts data makes sue that it only extracts the supported types.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -117,6 +118,25 @@ sql:
...
                 COMPUPDATE OFF
 
+        # ------------ school_year  -------------------------------------------------------------------
+        school_year:
+          sql:
+            warehouseRead: >-
+                SELECT
+                  year
+                FROM school_year
+                INTO OUTFILE S3 '${archive.root}/${migrate.aws.location}/year'
```
agorina: I do not know why, but if i name this file 'schoo_year` it conflicts with the `school` file down below.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -252,11 +272,12 @@ sql:
                school_year,
                subject_id,
                name,
+               label,
                deleted,
                update_import_id,
                ? as migrate_id
              FROM asmt wa
-              WHERE wa.type_id = 1
+              WHERE wa.type_id in (1, 3)
```
agorina: Now support ICA and Summative.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -116,6 +116,24 @@ sql:
                DELETE FROM grade
                WHERE NOT EXISTS(SELECT sg.id FROM staging_grade sg WHERE sg.id = grade.id);
 
+       # ------------ school_year  -------------------------------------------------------------------
+        school_year:
+          sql:
+            update:  >-
+               SELECT 1;
```
jtreuting: is this just a placeholder, or does it have a purpose?

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -447,9 +467,24 @@ sql:
...
+      # ------------ active_asmt_per_year -----------------------------------------------------------------------
+        active_asmt_per_year:
+          sql:
+            insert: >-
+             INSERT INTO active_asmt_per_year (asmt_id, school_year)
+               SELECT asmt_id, school_year
+                 FROM
+                 ( SELECT DISTINCT f.asmt_id as asmt_id,  f.asmt_grade_id AS grade_id, f.school_year,  asmt.subject_id, asmt.type_id FROM fact_student_exam f JOIN asmt asmt ON asmt.id = f.asmt_id
+                   UNION
+                   SELECT id AS asmt_id, grade_id, school_year, subject_id, type_id FROM asmt
```
jtreuting: it seems like the part after the `UNION` would cover everything and maybe the first part isn't needed?

the first part is all assessments where a student has taken an exam for them, and the second part is all assessments in general, right?  if a student took an exam for an assessment doesn't that mean it's in the asmt table already so that covers both situations?

i think there is a distinct chance i'm missing some part of this equation though.

also small nit: seems odd to have the alias for the `asmt` table be `asmt`, couldn't that just be left off?

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -124,17 +125,29 @@ public Step upsertAsmtsStep() {
...
 
-        return stepBuilderFactory.get(upsertFactStudentIcaExamStepName)
+        return stepBuilderFactory.get(upsertFactStudentExamStepName)
                 .tasklet(new SqlListExecutionStep(
                         reportingSqlListExecutionRepository,
                         sqlsBuilder.build())).build();
     }
 
+    @Bean
+    public Step deriveActiveAsmtStep() {
```
malaffoon: nit: other steps have db-related verb in them (e.g. upsert, truncate); perhaps this could be `stageAsmtActiveYearStep` or some such. Yeah, i won't lose sleep over it either way.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
malaffoon: The union is needed because there may be a test result for an assessment for the "wrong" year. For example, the 2016-2017 assessment (school year = 2017) is given in 2017-2018; so we want to know that that assessment was "active" for school year 2018.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
agorina: There is one step to manage all CODEs and it expects all three statements (insert, update, delete) per each code. This is an empty filler for the 'update'.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
agorina: Will fix alias.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
malaffoon: And `SELECT 1` works fine in the execute method?

agorina: It seems to work. It is being used in the reporting (not olap version) this way too.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
jtreuting: Does it make sense (from a performance perspective) to narrow the first part to just the "off-year" results so that the distinct has less work to do?

agorina: Redshift seems to be very good with handling DISTINCT. 
The above statement with 1,217,410 exams takes about 2 sec. And with 24,215,377 records about 10 sec. 

Considering that it is going to be executed once per day, I am not sure it is worth optimizing. Let me know if you disagree.

jtreuting: i guess it depend on how fast it would be eliminating those :)  it's probably not an issue because it runs once a day

agorina: I am going to keep an eye on the overhead this step adds, and if it becomes significant - will look into optimizing.

agorina:
I am reloading the dev version of Redshift and I am seeing something puzzling:
Here are the execution time in seconds for each step:
Avg Time in sec | Step name
-|-
15.0862 | extractEntitiesStep
12.6379 | loadEntitiesToStagingStep
11.8596 |	upsertCodesStep
5.5088 |	stepTruncateStageAfter
5.1897 |	stepTruncateStageBefore
5.0000 |	upsertFactStudentExamStep
3.3860 |	upsertStudentsStep
3.2982 |	deleteEntitiesStep
2.6842 |	upsertOrganizationsStep
2.2241 |	cleanupLandingZone
1.7895 |	upsertAsmtsStep
1.2069 |	loadCodesToStagingStep
1.0526 |	deriveActiveAsmtStep
1.0526 |	deleteCodesStep
1.0000 |	extractCodesStep

Here is the total sum of all steps execution time:
```sql
select  sum(timestampdiff(SECOND, START_TIME, END_TIME)) as runtime_in_sec from BATCH_STEP_EXECUTION b ;
```
-- 4460

But the overall total JOB execution time (the same value from both `migrate` and `BATCH_JOB_EXECUTION`)

-- 6184

Where does the time go? My first though was that maybe finding the next batch takes that long, but the Aurora slow query log does not prove it.

So far I have migrates 1,465,442 exams and it has been running for 105 min. I think it is two times slower than I remember from the last run (but I may have increased the size of the batch chunk last time). I still think 105 min is reasonable, but I would like to understand the 'lost' time.

malaffoon:
My guess would be the overhead (in the batch framework) of instantiating and tearing down the steps. Imagine that getting a connection takes a couple seconds -- from the framework's point of view, does that happen during a step or while setting up the step?

agorina:
Just to note: I re-ran it with 1 mil as a batch size (double from the before) and the execution time was 62 min.

PR #427, opened 2017-12-21
DWR-1215  Do not load null scored exams into Redshift

agorina:


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -476,7 +476,7 @@ sql:
                   update_import_id,
                  ? as migrate_id
                 FROM exam we
-                  WHERE we.created > ? AND we.created <= ? AND type_id in (1, 3)
+                  WHERE we.created > ? AND we.created <= ? AND type_id in (1, 3) AND we.scale_score IS NOT NULL AND we.scale_score_std_err IS NOT NULL AND we.performance_level IS NOT NULL
```
agorina: I do not know if it is an overkill to check all three values, but since we are doing aggregate calculations on each of them this feels safer. (?)

malaffoon: 👍

PR #428, opened 2017-12-21
integration test to exercise percentile data

chdietz:
Not to commit but to test and discuss schema changes with same branch name.
- updated to work with recent schema changes


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
```diff
@@ -0,0 +1,132 @@
...
+                        "  (2,50,2420,2408,2420),\n" +
+                        "  (2,55,2433,2420,2433),\n" +
+                        "  (2,60,2445,2433,2445),\n" +
+                        "  (2,65,2458,2445,2458),\n" +
+                        "  (2,70,2471,2458,2471),\n" +
+                        "  (2,75,2485,2471,2485),\n" +
+                        "  (2,80,2500,2485,2500),\n" +
+                        "  (2,85,2517,2500,2517),\n" +
+                        "  (2,90,2538,2517,2538),\n" +
+                        "  (2,95,2566,2538,9999)\n" +
```
malaffoon: This doesn't seem correct. A score of, for example 2600, will end up being categorized as 95?

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
```diff
@@ -0,0 +1,132 @@
...
+    private NamedParameterJdbcTemplate namedParameterTemplate;
+
+    @Test
+    public void itShouldFindPercentileForOneExam() {
+        assertThat(namedParameterTemplate.queryForObject(
+                "SELECT ps.percent FROM exam e\n" +
+                        "LEFT JOIN percentile p ON e.asmt_id = p.asmt_id\n" +
+                        "LEFT JOIN percentile_score ps ON ps.percentile_id = p.id\n" +
+                        "WHERE e.oppId = :oppId\n" +
+                        "      AND e.scale_score BETWEEN (ps.min + 1) AND ps.max",
```
malaffoon: I don't think we should use BETWEEN because it forces us to do that +1 stuff. If we decide we want to use floats we will be stuck. I'd much prefer `... AND ps.min < e.scale_score AND e.scale_score <= ps.max`.

exam-processor/src/test/resources/PreloadAsmtOrgStudentExam.sql
```diff
@@ -0,0 +1,166 @@
...
+  (8, 2, 2018, 3, '11111', 5, '100000000020', 2, 3, 'LEE-3bb6', 2601, 44, 3, '2018-05-07 08:50:00.000000', -1, -1, 0, '2017-12-21 06:36:39.267195', '2017-12-21 06:36:39.267195', 11, 1, 2, 1, 0, 0, 1, NULL, NULL, NULL, NULL, NULL, '2018-05-07 08:50:00.000000', 100000000333, 'online', NULL, 'MOCK', NULL, '2018-05-07', '2018-05-07 08:05:00.000000', NULL, 'scored', 15, 0, 0, 0, 0, 'WINDOW_ID', NULL, 'SBAC', NULL, '', 'ip-10-113-148-45', 'session', NULL, NULL, NULL),
+  (9, 1, 2018, 1, '11111', 2, '100000000080', 2, 2, 'MAG-3818', 2600, 67, 3, '2018-05-07 08:50:00.000000', -1, -1, 0, '2017-12-21 06:36:39.303356', '2017-12-21 06:36:39.303356', 11, 4, 1, 1, 0, 0, 1, NULL, NULL, NULL, NULL, NULL, '2018-05-07 08:50:00.000000', 100000000333, 'online', NULL, 'MOCK', NULL, '2018-05-07', '2018-05-07 08:05:00.000000', NULL, 'scored', 44, 0, 0, 0, 0, 'WINDOW_ID', NULL, 'SBAC', NULL, '', 'ip-10-113-148-45', 'session', NULL, NULL, NULL);
+INSERT INTO exam_claim_score (id, exam_id, subject_claim_score_id, scale_score, scale_score_std_err, category, created, theta_score, theta_score_std_err)
+VALUES (1, 2, 7, 2501, 42, 3, '2017-12-21 06:36:38.815810', NULL, NULL),
```
malaffoon: This is a lot of set up data, can't you just skip the claim scores and item level data?

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
malaffoon: Hmm, thinking about this.

exam-processor/src/test/resources/PreloadAsmtOrgStudentExam.sql
chdietz: I used a dump of real exams being loaded.  I thought that it feels odd to remove child tables that would normally exist.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
chdietz: Have to admit I went back and forth on this one.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
chdietz: I thought the way this works is this:
If the highest percentile they give us is 95 then no matter how high the score is the highest percentile we can say you are in is 95

chdietz: I have seen other tables for datasets where they do it by 5 percent increments but after 95 they give a few more like 98, 99 etc.  So if they provide it we use it.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PercentileSchemaIT.java
```diff
@@ -0,0 +1,132 @@
...
+    private NamedParameterJdbcTemplate namedParameterTemplate;
+
+    @Test
+    public void itShouldFindPercentileForOneExam() {
+        assertThat(namedParameterTemplate.queryForObject(
+                "SELECT ps.percent FROM exam e\n" +
+                        "LEFT JOIN percentile p ON e.asmt_id = p.asmt_id\n" +
+                        "LEFT JOIN percentile_score ps ON ps.percentile_id = p.id\n" +
+                        "WHERE e.oppId = :oppId\n" +
+                        "      AND e.scale_score > ps.min_score AND e.scale_score <= ps.max_score",
```
chdietz: No longer using between

PR #429, opened 2017-12-26
fix warnings about non-validated beans

malaffoon:
I'm not thrilled with this but ...

The customer notices when there are warnings in the log. And there are a bunch of messages regarding beans with java validation annotations that don't have the `@Validate` annotation to go with them. I tried the obvious: added `@Validate` annotation. But then tests started failing and the running applications wouldn't start up due to validation failures. So, near as i can tell, the java validation annotations were never being honored. I couldn't get things working properly in my self-imposed time box of 4 hours so i decided to just remove the annotations.

If we ever get some free time perhaps we can revisit this to figure out how to convince Spring to do what we want here. I'm sure there's a way, just kind of fed up with it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,14 +101,14 @@ subprojects {
         }
     }
 
+    // this is for creating metadata
+    compileJava.dependsOn(processResources)
+
     /***********************************************************
      * Docker setup for all sub-projects that need it
      ***********************************************************/
     if (!it.hasProperty('skipDocker')) {
         apply plugin: 'com.bmuschko.docker-remote-api'
-
-        // this is for creating metadata
-        compileJava.dependsOn(processResources)
 
         classes.dependsOn(createBuildInfoFile)
 
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/config/SbacScoreConfiguration.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/config/SbacScoreConfiguration.java
@@ -2,20 +2,14 @@ package org.opentestsystem.rdw.ingest.common.config;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
-import javax.validation.constraints.NotNull;
-
 /**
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -101,15 +101,15 @@ subprojects {
         }
     }
 
+    // this is for creating metadata
+    compileJava.dependsOn(processResources)
+
```
malaffoon: Unrelated ... just noticed that the common modules (which skip docker) weren't being processed for configuration properties.

chdietz:
Interesting, will take a look, I have to agree on not wanting warnings

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/configuration/GroupProcessingSqlConfiguration.java
```diff
@@ -1,11 +1,9 @@
 package org.opentestsystem.rdw.ingest.group.configuration;
 
-import org.hibernate.validator.constraints.NotEmpty;
 import org.opentestsystem.rdw.ingest.common.model.EntitySqls;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.stereotype.Component;
 
-import javax.validation.Valid;
```
chdietz: I suspect this just got copied over from the migrate logic

common/src/main/java/org/opentestsystem/rdw/ingest/common/config/SbacScoreConfiguration.java
```diff
@@ -2,21 +2,15 @@
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
-import javax.validation.constraints.NotNull;
-
 /**
  * SBAC scores configuration
  */
 @ConfigurationProperties(prefix = "sbac")
```
chdietz: @malaffoon I think `@Validated` is missing to make it work ?

PR #431, opened 2017-12-27
DWR-1215 Migrate district/school groups

agorina:


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -442,13 +466,129 @@ sql:
...
+             warehouseRead: >-
+              SELECT
+                id,
+                name,
+                natural_id,
+                external_id,
+                ? as migrate_id
+              FROM district_group wdg
+               WHERE EXISTS( SELECT id FROM school  WHERE created > ? AND created <= ? AND deleted = 0 AND district_group_id = wdg.id)
+              INTO OUTFILE S3 '${archive.root}/${migrate.aws.location}/dstrgroup'
```
agorina: Using `district_group` as a file name causes a conflict with the '`district`. I am not sure why but suspect that it goes by the first X chars?

malaffoon: That sucks. But i'd vote for a different contraction, per just `dgroup` and `sgroup`?

chdietz: Is the conflict on the write or the load?
Wonder if something is matching only up to the underscore?

agorina: The conflict is on the write. And I have tried without the underscore, no luck.

PR #433, opened 2017-12-27
DWR-1215 Little refactoring of embargo migrate

agorina:


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlCopyCodesExecutionStep.java
```diff
@@ -26,8 +26,8 @@
      * @param repository        the repository
      * @param sqlCopyStatements the sqls to be executed by this step
      */
-    public SqlCopyCodesExecutionStep(final SqlCopyWarehouseToStagingRepository repository,
-                                     final List<SqlCopyStatements> sqlCopyStatements) {
+    SqlCopyCodesExecutionStep(final SqlCopyWarehouseToStagingRepository repository,
```
agorina: Unrelated clean up in this file.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/SqlCopyEmbargoExecutionStep.java
```diff
@@ -6,40 +6,35 @@
...
-import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 
 import java.util.List;
 
 /**
  * A step for executing {@link SqlCopyEmbargoExecutionStep#sqlCopyStatements} if
- * {@link Migrate#isMigrateEmbargo()} is true. It also passes in the current school year.
+ * {@link Migrate#isMigrateEmbargo()} is true.
  */
 class SqlCopyEmbargoExecutionStep extends StepTasklet {
```
malaffoon: I'd be tempted to merge the three SqlCopy*ExecutionStep: if there was either a functor or a method that answered the question "should migrate" they would all be exactly the same. The base/default method/functor would return true, while the other two would check their appropriate flag.

agorina: Yeah...I was looking at it and decided to leave it as is for now.

PR #434, opened 2017-12-28
Feature/dwr 1215 embargo

agorina:
Migrate a district embargo flag into a school table in Redshift.

Lots and lots of 'copy and paste' from https://github.com/SmarterApp/RDW_Ingest/pull/424/files.

I suspect that we may need a state embargo too, but will chat with people offline.


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -238,6 +238,7 @@ sql:
                   ss.external_id,
                   ss.school_group_id,
                   ss.district_group_id,
+                  1,
```
agorina: This is different from reporting since by default it should be embargoed.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -253,6 +254,7 @@ sql:
                 external_id  = ss.external_id,
                 school_group_id  = ss.school_group_id,
                 district_group_id  = ss.district_group_id,
+                embargo_enabled = 1,
```
agorina: I followed the logic in reporting, but not sure if this is right. 
What if the school changed but the embargo has not?
I think this may work since we migrate embargo with every school, but not completely sure.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/UpdateEmbargoIT.java
```diff
@@ -29,8 +29,8 @@
     private JdbcTemplate reportingJdbcTemplate;
 
     @SqlGroup({
-            @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:StagingEmbargoSetup.sql"}, config = @SqlConfig(dataSource = "reportingDatasource")),
-            @Sql(executionPhase = AFTER_TEST_METHOD, scripts = {"classpath:StagingEmbargoTeardown.sql"}, config = @SqlConfig(dataSource = "reportingDatasource"))
```
agorina: I renamed these files to reflect what they do more accurately.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
jtreuting: are you saying that this would always update the embargo to enabled and then in another step set it back if it had been released already?

migrate-olap/src/main/resources/application.yml
```diff
@@ -1,5 +1,8 @@
 # this app is expected to run inside the security perimeter so disable security for actuator end-points and,
 # since those are the only end-points, switch to a less standard port using the main server.port config
```
jtreuting: these comments are now getting further away from the settings, might want to reshuffle them to be next to the security info with this new setting in place

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
agorina: That is my understanding. I need @malaffoon to confirm or deny this.

I still think this may be unsafe since these two steps are wrapped into different transactions.

migrate-olap/src/main/resources/application.yml
agorina: Good point, will do.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
malaffoon: That is correct: the embargo migrate must be done if there is any change to schools.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/UpdateEmbargoIT.java
malaffoon: Hmm, i won't fight it but the old name meant "setup for testing the staging of embargo settings" which i think is a very accurate description of what they do.   :P

agorina: Sorry - it probably did not come right. Let me try to rephrase it: "since the SQL scripts set up the data in both reporting and staging tables for embargo testing it seemed more natural *to me* to rename them".

PR #436, opened 2017-12-28
Added state embargo to olap migrate

agorina:
I think that state aggregate queries could benefit from this flag. If not, we can remove it then.


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -254,7 +255,6 @@ sql:
                 external_id  = ss.external_id,
                 school_group_id  = ss.school_group_id,
                 district_group_id  = ss.district_group_id,
-                embargo_enabled = 1,
```
agorina: Unrelated: I removed it here and in the reporting migrate too. @malaffoon, please let me know if you disagree.

PR #437, opened 2017-12-29
Adjusted to work with the cleaned up schema.

agorina:


agorina:
I have one tests that fails intermittently, most likely related to the data not being clean up. Checking into it.

PR #438, opened 2018-01-02
DWR-1219 Adjusted to work with the latest Redshift schema changes

agorina:
Related to https://github.com/SmarterApp/RDW_Schema/pull/219


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -444,11 +444,11 @@ sql:
                   se.grade_id,
                   ra.grade_id,
                   se.school_year,
-                  se.iep,
-                  se.lep,
-                  se.section504,
-                  se.economic_disadvantage,
-                  se.migrant_status,
+                  CASE WHEN se.iep = true THEN 1 WHEN se.iep = false THEN 0 ELSE 2 END,
```
malaffoon: If i'm reading this correctly ... in Aurora, these fields in the reporting data mart are 0,1,null and in Redshift they are 0,1,2? That inconsistency isn't going to bite us?

agorina: Hmmm...maybe. 
Not sure how to address it though: change reporting? 
What are your thoughts?

malaffoon: or change the code lookup table to allow the key to be null, instead of forcing it to be 2

agorina: I need not NULL to be able to join on it (I think).

PR #441, opened 2018-01-10
DWR-1294 Fix to embargo migration

agorina:
The very first solution had a SQL (that I returned back), then I introduced a view to share the logic between two migrates. The view was not doing exactly what the SQL did. This put the query into a shared SQL file.

### DIFF
--- a/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
+++ b/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
@@ -782,11 +782,10 @@ sql:
           sql:
             warehouseRead:
               SELECT
-                district_id,
-                aggregate,
+                e.district_id,
+                e.aggregate,
                 ? as migrate_id
               FROM (${sql.embargo}) e
-               WHERE e.school_year = ${app.school-year}
               INTO OUTFILE S3 '${archive.root}/${migrate.aws.location}/embargo'
                               FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '"'
                               LINES TERMINATED BY '\n'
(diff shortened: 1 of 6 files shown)


migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/StageEmbargoIT.java
```diff
@@ -34,7 +34,7 @@
             @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:WarehouseCodesSetup.sql"}, config = @SqlConfig(dataSource = "warehouseDatasource")),
             @Sql(executionPhase = BEFORE_TEST_METHOD, scripts = {"classpath:WarehouseEntitiesSetup.sql"}, config = @SqlConfig(dataSource = "warehouseDatasource")),
             @Sql(executionPhase = BEFORE_TEST_METHOD, config = @SqlConfig(dataSource = "warehouseDatasource"), statements = {
-                    "INSERT INTO district_embargo (district_id, school_year, individual, aggregate) VALUES (-1,  1999, null, null), (-98, 1999, 0, 0), (-99, 1999, 1, 1)"
+                    "INSERT INTO district_embargo (district_id, school_year, individual, aggregate) VALUES (-98, 1999, 0, 0), (-99, 1999, 1, 1)"
```
agorina: This test with the change was failing before the change.

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -785,8 +785,8 @@ sql:
                 district_id,
                 aggregate,
                 ? as migrate_id
-              FROM embargo
-               WHERE school_year = ${app.school-year}
+              FROM (${sql.embargo}) e
```
agorina: This is a blind change since I cannot test RST for now. Will not merge until it is tested.

malaffoon: You can now run RSTs. It's really easy if you're using CI resources. Bit more work if you want to avoid that and have your own database.

PR #446, opened 2018-01-12
Fix a bug with olap embargo migrate.

agorina:
While working with Noehmi today we discovered this bug.


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -18,6 +18,7 @@ sql:
       - delete from staging_exam_claim_score
       - delete from staging_school_year
       - delete from staging_district_embargo
+      - delete from staging_state_embargo
```
agorina: Since this was not cleaned up, there are multiple entries in the final destination table.

PR #447, opened 2018-01-15
redshift timezone fix

malaffoon:
This change goes with schema change. I'll update the schema dependency when it finishes building and all that ... just wanted to get the review out.


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -168,6 +168,7 @@ sql:
               FROM '${archive.root}/${migrate.aws.location}/student.part_00000'
               CREDENTIALS 'aws_iam_role=${migrate.aws.redshift.role}'
               FORMAT AS CSV
+              TIMEFORMAT 'auto'
```
malaffoon: FYI, I added this to only those that have timestamp fields.

PR #448, opened 2018-01-16
DWR-1315 Fix a bug with olap active asmt migrate

agorina:


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -153,7 +153,7 @@ public Step deriveActiveAsmtStep() {
                 .addNext("asmt_active_year", "insert");
 
         return stepBuilderFactory.get(deriveActiveAsmtStep)
-                .tasklet(new SqlListCodesExecutionStep(
+                .tasklet(new SqlListExecutionStep(
```
agorina: A silly 'copy and paste' error.

PR #450, opened 2018-01-23
DWR-1230 Misc changes based on the load test

agorina:


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -168,10 +168,12 @@ public Step truncateBeforeStageStep() {
 
     @Bean
     public Step truncateAfterStageStep() {
+        final List<String> sqls = warehouseToStagingSqlConfiguration.getTruncateStageList();
+        sqls.addAll(stagingToReportingSqlConfiguration.getFinalizeList());
```
agorina: Logically this may belong to a separate step,but it seemed like an overkill to me. 
Let me know if you disagree.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -407,7 +410,6 @@ sql:
                 student_id,
                 asmt_id,
                 grade_id,
-                asmt_grade_id,
```
agorina: There is a corresponding schema change.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
malaffoon: Is this intended to invoke `analyze` and warm the redshift cache?  If so, i would vote for it being a separate step: i'd like to see the times associated with it, and generally be able to manage it separately. If it is something else or if you disagree then i would only ask that the step name be changed to reflect its wider scope.

agorina: Makes sense. Will change.

PR #451, opened 2018-01-23
Feature/percentile ingest

chdietz:
think this is ready to merge

* added update logic
* refactored exception handling in parse
* added check for overlapping dates in parse

Create / Update Logic
* tables are loaded in a batch, multiple in one import/CSV.
* if there are any errors, all are parsed and reported on, no creates or updates occur.
* note: validation errors include a table with identical `asmt_id` and overlapping dates.
* if keys match; `asmt_id`, `start_date`, `end_date`
  * it's the same percentile table
  * delete any `percentile_score` child records
  * update `percentile` set columns to trigger migrate
  * create `percentile_score` child records for updated `percentile` parent
* if keys do not match use create logic
* api will report tables (created and/or updated) or errors

We are not auditing these tables


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
```diff
@@ -0,0 +1,69 @@
...
+    private String sqlPercentileScoreCreate;
+
+    @Autowired
+    public JdbcPercentileRepository(final NamedParameterJdbcTemplate jdbcTemplate) {
+        this.jdbcTemplate = jdbcTemplate;
+    }
+
+    @Transactional
+    @Override
+    public void create(final Collection<Percentile> percentiles, final long importId) {
```
chdietz: create logic implemented, looking for feedback on update locic

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
```diff
@@ -0,0 +1,20 @@
...
+ * PercentileRepository - Interface for bulk inserting an percentile tables.
+ */
+public interface PercentileRepository {
+    /**
+     * Method to create new {@link Percentile}s
+     *
+     * @param percentiles collection of {@link Percentile}s
+     * @param importId    id of the import that is creating the {@link Percentile}s
+     */
+    void create(Collection<Percentile> percentiles, long importId);
```
malaffoon: Not quite sure where this will end up but perhaps this should just be `save` with the update logic buried in it? Or were you going to put update logic in the service layer so you'd need a `create` and `update` in the repo?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -102,23 +104,26 @@ public DefaultPercentileParser(final AssessmentPackageRepository assessmentPacka
             asmtNaturalId = record.get(AssessmentId).trim();
             startDate = record.get(StartDate).trim();
             endDate = record.get(EndDate).trim();
+            // The min and max score is used in the percentile rank.
+            minScore = parserHelper.validate(contextName(MinScore), record.get(MinScore), toScaleScoreMin);
```
malaffoon: not obvious to me why these are up here instead of inline like all the other ones

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -156,6 +162,44 @@ private void validatePercentileScoreOrder() {
...
         }
 
+        private List<PercentileScore> calculateMinAndMax(List<PercentileScore> inScores, Double minScore, Double maxScore) {
+            Double previousScore = null;
+            List<PercentileScore> outScores = new ArrayList<>();
+            PercentileScore.Builder builder = new PercentileScore.Builder();
+
+            ListIterator<PercentileScore> iterator = inScores.listIterator(inScores.size());
+
+            while (iterator.hasPrevious()) {
```
malaffoon: i know collection-wonks suggest you always use iterators and never use indexing but i think this is really hard to read and something simpler would be easier; pseudo-code:
```
    outScores.add(inScores[0].copy().minScore(minScore).build());
    for (int i = 1; i < inScores.size() -1; ++i) {
        outScores.add(inScores[i].copy().maxScore(inScores[i+1].getScore()).build());
    }
    outScores.add(inScores[inScores.size()-1].copy().maxScore(maxScore).build());
```

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
chdietz: I was leaning with the latter.  But will see how it flushes out.
At a minimum I would like all of the update in one transaction and all create in one.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
chdietz: there are 5 things that may be needed when evaluating subsequent records after initiating a new parent,
the 3 key parts and the minScore and maxScore for calculateMinAndMax()

If they are just in the parent builder that is partially complete then I need a getter on the builder members, ugly.  I plan to refactor this parser and add actual units tests.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
chdietz: thanks, i will keep in mind on refactor.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
```diff
@@ -0,0 +1,152 @@
...
+                        .addValue("percentile_id", percentile.getId())
+                        .addValue("percentile_rank", percentileScore.getPercentileRank())
+                        .addValue("score", percentileScore.getScore())
+                        .addValue("min_inclusive", percentileScore.getMinInclusive())
+                        .addValue("max_exclusive", percentileScore.getMaxExclusive()));
+            }
+        }
+        jdbcTemplate.batchUpdate(sqlPercentileUpdate, updatePercentileParams.toArray(new MapSqlParameterSource[updatePercentileParams.size()]));
+        jdbcTemplate.batchUpdate(sqlPercentileScoreDelete, deletePercentileScoreParams.toArray(new MapSqlParameterSource[deletePercentileScoreParams.size()]));
+        jdbcTemplate.batchUpdate(sqlPercentileScoreInsert, insertPercentileScoreParams.toArray(new MapSqlParameterSource[insertPercentileScoreParams.size()]));
```
chdietz: new update method as discussed, for update update parent, delete children and re add them.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
```diff
@@ -0,0 +1,152 @@
...
+    @Override
+    public List<Integer> findIdsByDateOverlap(final Integer assessmentId, final LocalDate startDate, final LocalDate endDate) {
+
+        return jdbcTemplate.queryForList(findIdsByDateOverlap,
+                new MapSqlParameterSource()
+                        .addValue("asmt_id", assessmentId)
+                        .addValue("start_date", startDate.toString())
+                        .addValue("end_date", endDate.toString()),
+                Integer.class);
+    }
```
chdietz: used to find overlapping dates if the composite key is the same

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
chdietz: @malaffoon When I was working in this section I was contemplating if there was a reasonable minimum number of ranks `percentile_score`'s per `percentile`  I would think 3 as in quartiles.  

But it is easy enough to just handle only one, and if there is only one use min_score and max_score for the min_inclusive and max_exclusive of the one rank record.

This table will not really be useful and could be due to an error in the creation of the CSV and thus why a minimum number may be a good idea?

malaffoon: Good point, we should code in a minimum. I think your idea of quartiles is a good limit, i've never seen tertiles or biniles/secondiles (??) listed or quoted.

chdietz: ok, i will code it for quartiles and that would be `3`, correct?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/PercentileScore.java
```diff
@@ -38,23 +38,32 @@ public Builder score(final Double score) {
...
         }
+
+        public Builder copy(final PercentileScore toCopy) {
+            this.percentileRank = toCopy.percentileRank;
+            this.score = toCopy.score;
+            this.minInclusive = toCopy.minInclusive;
+            this.maxExclusive = toCopy.maxExclusive;
+            return this;
+        }
+
```
chdietz: refactored to match changes in schema column names, added copy.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
```diff
@@ -0,0 +1,60 @@
...
+public interface PercentileRepository {
+
+
+    /**
+     * Method to save {@link Percentile}s as new or updates to existing.
+     *
+     * @param percentiles collection of {@link Percentile}s to be persisted
+     * @param importId    id of the import that is creating the {@link Percentile}s
+     */
+    void save(Collection<Percentile> percentiles, long importId);
```
chdietz: added a save method accepts list of new and updated.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
chdietz: yes, done.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -75,89 +78,170 @@ public DefaultPercentileParser(final AssessmentPackageRepository assessmentPacka
...
-        private final String startDate;
-        private final String endDate;
-
-        private List<PercentileScore> percentileScores = new ArrayList<>();
+    private List<Percentile> parseRecords(final Iterator<CSVRecord> recordIterator) {
+
+        final List<Percentile> percentiles = new ArrayList<>();
+        final DataElementErrorCollector parseErrorCollector = new DataElementErrorCollector();
+
+        CsvPercentile csvPercentile = new CsvPercentile(parseErrorCollector);
```
chdietz: just create once, and it holds the error collector.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -75,89 +78,170 @@ public DefaultPercentileParser(final AssessmentPackageRepository assessmentPacka
...
+                csvPercentile.addPercentile(record);
+            } else {
+                csvPercentile.addPercentileScore(record);
+            }
+        }
+        percentiles.add(csvPercentile.build());
+
+        // Errors are collected and thrown if not empty after parsing all rows for a comprehensive list
+        if (!parseErrorCollector.isEmpty()) throw new ImportException(BAD_DATA, parseErrorCollector.toJson());
+        return percentiles;
```
chdietz: No more return empty list, throw exception with errors found if any

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -75,89 +78,170 @@ public DefaultPercentileParser(final AssessmentPackageRepository assessmentPacka
...
-            if (errorCollector.isEmpty()) {
-                return percentileBuilder
-                        .percentileScores(percentileScores)
-                        .build();
+
+            return percentileBuilder
+                    .id(getExistingPercentileId(current.getAssessmentId(), current.getStartDate(), current.getEndDate()))
+                    .percentileScores(calculateMinAndMax(percentileScores, current.getMinScore(), current.getMaxScore()))
+                    .build();
+        }
```
chdietz: percentile is built out ready to update or create

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileParser.java
```diff
@@ -75,89 +78,170 @@ public DefaultPercentileParser(final AssessmentPackageRepository assessmentPacka
...
+                        .maxExclusive(inScores.get(i + 1).getScore())
+                        .build());
+
+            }
+            outScores.add(builder.copy(inScores.get(inScores.size() - 1))
+                    .minInclusive(inScores.get(inScores.size() - 1).getScore())
+                    .maxExclusive(maxScore)
+                    .build());
+            return outScores;
+        }
```
chdietz: refactored and added check for minimum reasonable number of ranks in a percentile table

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/PercentileParser.java
```diff
@@ -15,10 +15,9 @@
      * in the {@link DataElementErrorCollector}.
      *
      * @param payload               byte array to parse
-     * @param elementErrorCollector {@link DataElementErrorCollector} to capture errors
      * @return collection of {@link Percentile}s parsed.  If there are any parse errors
      * an empty list is returned
      */
-    List<Percentile> parse(byte[] payload, DataElementErrorCollector elementErrorCollector);
+    List<Percentile> parse(byte[] payload);
```
chdietz: no need to pass error collector, any errors are thrown

chdietz:
@malaffoon I think this is ready to merge, you are only one commented on it thoughts?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
malaffoon: Does this mean we can get rid of `create` method?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
```diff
@@ -0,0 +1,165 @@
...
+                    .addValue("min_score", percentile.getMinScore())
+                    .addValue("max_score", percentile.getMaxScore())
+                    .addValue("import_id", importId)
+                    .addValue("update_import_id", importId));
+
+            // Collect params to create children of new Percentile with unknown id
+            for (final PercentileScore percentileScore : percentile.getPercentileScores()) {
+                createPercentileScoreParams.add(new MapSqlParameterSource()
+                        .addValue("asmt_id", percentile.getAssessmentId())
+                        .addValue("start_date", percentile.getStartDate())
```
malaffoon: every child has independent start/end date?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public String process(final byte[] payload, final long importId) {
+
+        final List<Percentile> percentiles = parser.parse(payload);
+
+        repository.save(percentiles, importId);
+
+        long created = percentiles.stream().filter(p -> p.getId() == null).count();
```
malaffoon: nit: final
I think this created count should be done before saving? I know we don't use an ORM but many repository implementations modify the objects in place (which would set all the ids making them all look like updates).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/PercentileParser.java
malaffoon: I'm sure i'll see in a minute but ... doesn't this allow us to detect multiple errors?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
```diff
@@ -0,0 +1,165 @@
...
+    public void update(final Collection<Percentile> percentiles, final long importId) {
+
+        final List<MapSqlParameterSource> updatePercentileParams = newArrayList();
+        final List<MapSqlParameterSource> deletePercentileScoreParams = newArrayList();
+        final List<MapSqlParameterSource> insertPercentileScoreParams = newArrayList();
+
+        for (final Percentile percentile : percentiles) {
+            // Collect params for Percentile to update
+            updatePercentileParams.add(new MapSqlParameterSource()
+                    .addValue("id", percentile.getId())
```
malaffoon: It is a little weird that we're doing id-based updates and not allowing asmt_id or start/end date to be changed. I guess the id is synonymous with asmt_id + date-range?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
```diff
@@ -0,0 +1,60 @@
...
+     * Method to create new {@link Percentile}s
+     *
+     * @param percentiles collection of {@link Percentile}s
+     * @param importId    id of the import that is creating the {@link Percentile}s
+     */
+    void create(Collection<Percentile> percentiles, long importId);
+
+    /**
+     * Method to update existing {@link Percentile}s
+     *
```
malaffoon: i think this should note that you can't change asmt_id or date-range values.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/PercentileScore.java
```diff
@@ -38,23 +38,32 @@ public Builder score(final Double score) {
...
+            percentileScore.percentileRank = percentileRank;
             percentileScore.score = score;
-            percentileScore.minScore = minScore;
-            percentileScore.maxScore = maxScore;
+            percentileScore.minInclusive = minInclusive;
+            percentileScore.maxExclusive = maxExclusive;
             return percentileScore;
         }
+
+        public Builder copy(final PercentileScore toCopy) {
```
malaffoon: i don't see the usage in the code but it is a teeny bit nicer for the client code if there is a static copy() method on the object that returns the builder so you can do this:
```
    return orig.copy().maxScore(2500).build();
```
Instead of this:
```
    return PercentileScore.builder().copy(orig).maxScore(2500).build();
```

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultPercentileProcessor.java
chdietz: good idea, fyi already merged but will make a list

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
chdietz: dates are part of the key (asmt,start,end) used to find the parent id on create of children

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/PercentileRepository.java
chdietz: good idea.  
So, there may be a need for a delete, for now that would be an ops/sql process

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/PercentileParser.java
chdietz: you will see that parse takes a payload and,
- it returns a list of parsed `Percentile`s   -or- it throws an exception
- if it's just parse errors the exception is ImportException with collected errors in message.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcPercentileRepository.java
chdietz: I could use the composite natural key of asmt_it, start_date, end_date but I have the id.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/PercentileScore.java
chdietz: good idea, will add to list of tweaks

PR #453, opened 2018-02-01
OLAP API: modify OLAP migrate to run VACUUM REINDEX

agorina:


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -1,6 +1,7 @@
 sql:
   migrate:
     finalizeList:
+      - VACUUM REINDEX fact_student_exam
```
agorina: I cannot come up with a way for QA to test it.
Anybody has ideas?

jtreuting: Is there something in the AWS portal that will show the last time Redshift was reindexed?  maybe could run migrate and then check that in the portal before and after

agorina: Good idea! I do not think there is anything in the AWS portal, but there is way to query the system tables. Thank you.

PR #456, opened 2018-02-06
fix a couple bugs when importing packages

malaffoon:
* The tabulator output often produces trailing empty records rows that are just a bunch of field delimiters ... the processor would fail which seems a little extreme.
* We've talked about allowing comments in our CSV.
* Import payloads submitted to the wrong end-point were causing bad duplicate detection. For example, if i accidentally post a package payload to the exams end-point, it properly fails to load anything. But then if i try to post the same payload to the packages end-point, it claims it is a duplicate payload.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -54,49 +54,49 @@ services:
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
 
-  exam-processor:
-    image: smarterbalanced/rdw-ingest-exam-processor
-    ports:
-      - 8082:8008
-    links:
-      - config-server
-      - rabbitmq
-    environment:
-      - spring.profiles.active=local-docker
-      - CONFIG_SERVICE_ENABLED=true
-      - CONFIG_SERVICE_URL=http://config-server:8888
-    extra_hosts:
-      - "dwhost:$DATAWAREHOUSE_HOST"
-
-  package-processor:
-    image: smarterbalanced/rdw-ingest-package-processor
-    ports:
-      - 8083:8008
-    links:
-      - config-server
-      - rabbitmq
-    environment:
-      - spring.profiles.active=local-docker
-      - CONFIG_SERVICE_ENABLED=true
-      - CONFIG_SERVICE_URL=http://config-server:8888
-    extra_hosts:
-      - "dwhost:$DATAWAREHOUSE_HOST"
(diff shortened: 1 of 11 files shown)


docker/docker-compose.yml
```diff
@@ -54,49 +54,49 @@ services:
     extra_hosts:
       - "dwhost:$DATAWAREHOUSE_HOST"
 
-  exam-processor:
```
malaffoon: doh, i'll revert this file change

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -72,6 +76,9 @@ public DefaultAssessmentParser(final SbacScoreConfiguration scoreConfiguration,
             assessmentParserHelper.newAssessmentRecord(recordIterator.next());
```
malaffoon: There is an edge case that isn't handled here: if the very first record (after the headers) is one of those empty record rows, this will still report an error. Since that isn't what i was trying to fix here, i didn't spend the time to rework the iterator logic to handle it.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorTest.java
```diff
@@ -44,7 +43,7 @@
 
 
     @Before
-    public void createProcessors() throws IOException {
```
malaffoon: fixing warnings, these tests do not throw checked exceptions

import-service/src/main/resources/application.sql.yml
```diff
@@ -16,8 +16,8 @@ sql:
     findOne: >-
       SELECT * FROM import WHERE id=:id
 
-    findOneByDigest: >-
-      SELECT * FROM import WHERE digest=:digest
+    findOneByContentAndDigest: >-
+      SELECT * FROM import WHERE content=:content AND digest=:digest
```
agorina: Should we change an index on this table to include `digest` and `content`?  I am not sure how often this will be executed.

malaffoon: i did an explain on dev and it uses the digest index and says it will be const lookup time

PR #457, opened 2018-02-07
Added VACUUM to olap migrate.

agorina:
I am not convinced we are doing the right thing. 
While it is needed, I am not sure if we need to run it with every migrate.

Thoughts?


malaffoon:
Why wouldn't we want to? It is something that needs to be done when data is added to the system and it isn't terribly slow is it?

agorina:
What is your definition of **terribly** slow? :)

agorina:
The last time @wtritch  ran it:
`sql> vacuum
[2018-02-06 10:09:38] completed in 5m 11s 314ms
`

PR #458, opened 2018-02-07
missed truncate new tables

chdietz:


migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -47,6 +47,8 @@ sql:
       - TRUNCATE TABLE staging_student_group
       - TRUNCATE TABLE staging_student_group_membership
       - TRUNCATE TABLE staging_user_student_group
+      - TRUNCATE TABLE staging_percentile_score
+      - TRUNCATE TABLE staging_percentile
       - TRUNCATE TABLE staging_asmt
```
chdietz: add tables to truncate

agorina: No test changes needed?  Are we missing a test for it? I see on in the OLAP but not the regular reporting. Maybe we should add one.

PR #464, opened 2018-03-09
DWR-1559 Remove unused data elements from Redshift

agorina:
Adjusted based on the schema changes: SmarterApp/RDW_Schema#247


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -719,6 +715,7 @@ sql:
                 DELIMITER ','
                 COMPUPDATE OFF
 
+ # TODO: remove it if we do not need claim level data
```
agorina: Just wanted to point this out. I did not want to remove it since we may need it.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -1,7 +1,6 @@
 sql:
   migrate:
     finalizeList:
-      - VACUUM REINDEX fact_student_exam
```
malaffoon: Why remove this one? From what i've seen the full VACUUM is the more time-consuming one.

agorina: The short answer is because we are removing "interleaved" sort key.

And the longer one: from Redshift docs (https://docs.aws.amazon.com/redshift/latest/dg/r_VACUUM_command.html): 

"REINDEX -Analyzes the distribution of the values in **interleaved sort key** columns, then performs a full VACUUM operation. 

VACUUM REINDEX takes significantly longer than VACUUM FULL because it makes an additional pass to analyze the interleaved sort keys. The sort and merge operation can take longer for interleaved tables because the interleaved sort might need to rearrange more rows than a compound sort...."

malaffoon: Ok, thanks for the explanation.

PR #466, opened 2018-03-14
DWR-1216 OLAP Migrate: modify to support IAB

agorina:
Some stats from AWS DEV:

Warehouse:
`select count(*) from exam;` -- 36,950,293
`select count(*) FROM (SELECT DISTINCT asmt_id, school_year, student_id FROM exam ) tmp;`
-- 13,455,425

When I was testing the migrate, from Redshift, after these changes:
`select count(*) from fact_student_exam;` -- 1,676,116
`select count(*) from fact_student_iab_exam;` -- 11,779,309

And the execution time with the `migrate.olap-batch.size` set to 5 mil.:
olap-batch-5mil

Overall run time 46 min.

This leads me to a question: any objections if I change the default and all the deployment/config script to `migrate.olap-batch.size=5000000`? (@malaffoon)


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -117,7 +122,7 @@ private boolean checkRunJobs() {
 
         // if this is not the first run then only return work if it is a full count
         // (otherwise it keeps scraping the last few records, effectively defeating the micro-batching)
-        if (subsequent && migrateImportValues.getImportCount() < batchSize) {
+        if (microBatchingEnabled && subsequent && migrateImportValues.getImportCount() < batchSize) {
```
agorina: For a case when we run a migrate on a scheduled cron time, this was missing the last chunk.

migrate-olap/src/test/resources/OlapAndStagingCodesTeardown.sql
```diff
@@ -6,9 +6,9 @@ delete from staging_gender;
 delete from staging_grade;
 
 DELETE FROM school_year WHERE year IN (1999, 1998);
-DELETE FROM completeness WHERE id IN (-99,-98);
-DELETE FROM administration_condition WHERE id IN (-99,-98, -97, -96);
-DELETE FROM ethnicity WHERE id IN (-99,-98);
-DELETE FROM gender  WHERE id IN (-99,-98);
-DELETE FROM grade WHERE id IN (-99,-98);
+DELETE FROM completeness WHERE id < 0 ;
```
agorina: I got lazy to track all the ids, and since we do not expect any negative numbers in the real life I took this shortcut.

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
```diff
@@ -49,7 +49,7 @@ public void setUp() throws JobExecutionException {
         jobLauncher = mock(JobLauncher.class);
         job = mock(Job.class);
 
-        reportingJobRunner = new MigrateJobRunner(migrateRepository, importRepository, jobLauncher, job, batchSize);
+        reportingJobRunner = new MigrateJobRunner(migrateRepository, importRepository, jobLauncher, job, batchSize, true);
```
wtritch: This is only true in tests?  ...I'm still confused by this parameter.  Maybe some javadocs would help?

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
```diff
@@ -84,6 +84,21 @@ public void isShouldStopTheLoopWhenNotFullBatchFoundOnSubsequentCall() throws Jo
         verify(jobLauncher, times(1)).run(same(job), any(JobParameters.class));
     }
 
+    @Test
+    public void isShouldNotStopTheLoopWhenNotFullBatchFoundOnSubsequentCallWithoutMicroBatching() throws JobExecutionException {
+        reportingJobRunner = new MigrateJobRunner(migrateRepository, importRepository, jobLauncher, job, batchSize, false);
+
```
agorina: @wtritch , there is the test with 'false' value.

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
agorina: Would this comment clarify it:
`While running this job frequently with the small batch size, we do not need to complete the last chunk if it is less than full size. The next round will pick it up. This is considered 'microBatchingEnabled`.

`When running it less frequently (let's say daily), we need to complete the less chunk regardless of its size. In this case microBatchingEnabled must be set to false`

wtritch: Ahhh, yes.  That makes it clear to me :-)

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
malaffoon: Bit nervous about this change. Imagine that test results are streaming in, every few seconds. This migrate process will continue to scrape exams and migrate them. We don't want that. I think each of the batch strategies need a different stopping mechanism. We can talk off-line about it?

malaffoon:
So 5000000 worked fine on a dc2.large cluster without straining it too much? Basically, i want the default setting to be something that won't break a normal, small cluster. If 5 million doesn't seem to be a problem then i'm fine with changing the default. Of course, if 5 million isn't a problem, how high can it go? Is this something where it the Aurora export and Redshift import / copy from staging can handle basically anything?

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
agorina: Ok, let's talk tomorrow.

agorina:
About 5 mil.: As you could tell I was using whatever instances we have and they handled it just fine. 
The largest extract from Aurora was 'exam' file and it was about 0.5GB. After 6 GB it starts breaking it into multiple files, and we do not support that.  I admit that 6-8 min per batch seems long enough to me. I am not sure I would want to go higher, but if needed we could test.

agorina:
Per discussion with @malaffoon I have rolled back changes to MigrateJobRunner since it had a problem. Merging as is and will created a JIRA to address the job runner issue.

agorina:
Just in case somebody is interested: after a little research, @malaffoon and I have agreed to change the default migrate.olap-batch.size to 10 millions.

PR #467, opened 2018-03-15
refine stopping condition for migrate job runner

malaffoon:
We want the migrate job runner to run as long as it has full batches, and then one more partial batch to catch any stragglers. But then it shouldn't continue processing partial batches because that would defeat the purpose of batching things as it processes exams as they trickle in.

The previous incarnation was a bit more complicated than it had to be. Admittedly, this does change the behavior a bit: every run will do a partial batch on the last iteration where before it would only do a partial batch on the first iteration.

### DIFF
--- a/migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
+++ b/migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
@@ -44,11 +44,16 @@ public class MigrateJobRunner implements MigrateLifecycle {
 
     public void run() {
         try {
-            boolean subsequent = false;
             Optional<MigrateJobParameters> jobParameters;
-            while (checkRunJobs() && (jobParameters = getNextJobParams(subsequent)).isPresent()) {
+            while (checkRunJobs() && (jobParameters = getNextJobParams()).isPresent()) {
                 jobLauncher.run(job, jobParameters.get());
-                subsequent = true;
+
+                // don't continue if the actual size is less than the limit
+                // this allows one run with less than batch-size, to catch the leftovers
+                // (otherwise it keeps scraping the last few records, effectively defeating the batching)
+                if (jobParameters.get().getBatchSize() < batchSize) {
+                    break;
+                }
             }
         } catch (final Exception e) {
             logger.error("migrate reporting job failure [" + e.getMessage() + "]");
@@ -101,23 +106,12 @@ public class MigrateJobRunner implements MigrateLifecycle {
         return isRunning() && isEnabled();
     }
 
-    private Optional<MigrateJobParameters> getNextJobParams(final boolean subsequent) {
+    private Optional<MigrateJobParameters> getNextJobParams() {
         final MigrateImportValues migrateImportValues = importRepository
                 .getMigrateImportValues(migrateRepository.findLastMigratedAt(), batchSize);
 
         // no work to do ...
         if (migrateImportValues.getImportCount() == 0) {
-            return Optional.empty();
-        }
-
-        // no work to do ...
-        if (subsequent && migrateImportValues.getFirstAt().equals(migrateImportValues.getLastAt())) {
-            return Optional.empty();
-        }
-
-        // if this is not the first run then only return work if it is a full count
-        // (otherwise it keeps scraping the last few records, effectively defeating the micro-batching)
-        if (subsequent && migrateImportValues.getImportCount() < batchSize) {
             return Optional.empty();
         }
 
--- a/migrate-common/src/main/resources/application.sql.yml
+++ b/migrate-common/src/main/resources/application.sql.yml
@@ -43,10 +43,10 @@ sql:
 
       existsImportContentInTimeRange:
           SELECT 1 AS existsCodeImportContent
-            FROM import wi
-          WHERE wi.content = :content
-             AND ((wi.status != 0 AND wi.updated >= :first_at AND wi.updated <= :last_at) OR
-               (wi.status = 0 AND wi.created <= :last_at))
+            FROM import
+          WHERE content = :content
+             AND ((status != 0 AND updated >= :first_at AND updated <= :last_at) OR
+               (status = 0 AND created <= :last_at))
           LIMIT 1
 
   # Create records with effective embargo for ALL districts regardless of timestamps. The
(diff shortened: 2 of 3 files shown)


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -44,11 +44,16 @@ public MigrateJobRunner(final MigrateRepository migrateRepository,
...
             Optional<MigrateJobParameters> jobParameters;
-            while (checkRunJobs() && (jobParameters = getNextJobParams(subsequent)).isPresent()) {
+            while (checkRunJobs() && (jobParameters = getNextJobParams()).isPresent()) {
                 jobLauncher.run(job, jobParameters.get());
-                subsequent = true;
+
+                // don't continue if the actual size is less than the limit
+                // this allows one run with less than batch-size, to catch the leftovers
+                // (otherwise it keeps scraping the last few records, effectively defeating the batching)
+                if (jobParameters.get().getBatchSize() < batchSize) {
```
malaffoon: Moving the logic to the end of the loop allows us to run the partial batch only once.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -110,17 +115,6 @@ private boolean checkRunJobs() {
...
         }
 
-        // no work to do ...
-        if (subsequent && migrateImportValues.getFirstAt().equals(migrateImportValues.getLastAt())) {
-            return Optional.empty();
-        }
-
-        // if this is not the first run then only return work if it is a full count
-        // (otherwise it keeps scraping the last few records, effectively defeating the micro-batching)
-        if (subsequent && migrateImportValues.getImportCount() < batchSize) {
```
malaffoon: This check got moved into the loop above.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -110,17 +115,6 @@ private boolean checkRunJobs() {
             return Optional.empty();
         }
 
-        // no work to do ...
-        if (subsequent && migrateImportValues.getFirstAt().equals(migrateImportValues.getLastAt())) {
```
malaffoon: Alla and i tried to figure out a situation where this would be true and couldn't, so it got yanked.

migrate-common/src/main/resources/application.sql.yml
```diff
@@ -43,10 +43,10 @@ sql:
...
       existsImportContentInTimeRange:
           SELECT 1 AS existsCodeImportContent
-            FROM import wi
-          WHERE wi.content = :content
-             AND ((wi.status != 0 AND wi.updated >= :first_at AND wi.updated <= :last_at) OR
-               (wi.status = 0 AND wi.created <= :last_at))
+            FROM import
+          WHERE content = :content
+             AND ((status != 0 AND updated >= :first_at AND updated <= :last_at) OR
+               (status = 0 AND created <= :last_at))
```
malaffoon: unrelated: this query was the only one in here that used a table alias and it isn't needed

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
```diff
@@ -61,27 +61,16 @@ public void isShouldLaunchJobInLoop() throws JobExecutionException {
     }
 
     @Test
-    public void isShouldStopTheLoopWhenFirstEqualsLast() throws JobExecutionException {
+    public void isShouldDoOneMoreWhenNotFullBatchFoundOnSubsequentCall() throws JobExecutionException {
         when(importRepository.getMigrateImportValues(null, batchSize))
                 .thenReturn(MigrateImportValues.builder().firstAt(firstAt).lastAt(lastAt).migrateCodes(true).importCount(batchSize).build())
-                .thenReturn(MigrateImportValues.builder().firstAt(lastAt).lastAt(lastAt).migrateCodes(true).importCount(batchSize).build());
+                .thenReturn(MigrateImportValues.builder().firstAt(lastAt).lastAt(nextLastAt).migrateCodes(true).importCount(batchSize - 1).build())
+                .thenReturn(MigrateImportValues.builder().firstAt(lastAt).lastAt(nextLastAt).migrateCodes(true).importCount(1).build());
```
malaffoon: Had to tweak this test because we now always do one partial batch.

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
```diff
@@ -61,27 +61,16 @@ public void isShouldLaunchJobInLoop() throws JobExecutionException {
...
 
         reportingJobRunner.run();
 
         verify(importRepository, times(2)).getMigrateImportValues(null, batchSize);
-        verify(jobLauncher, times(1)).run(same(job), any(JobParameters.class));
-    }
-
-    @Test
-    public void isShouldStopTheLoopWhenNotFullBatchFoundOnSubsequentCall() throws JobExecutionException {
```
malaffoon: removed this path

migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunnerTest.java
```diff
@@ -105,15 +94,28 @@ public void isShouldHandleExceptionsInJobAndStopTheLoop() throws JobExecutionExc
...
+                .thenReturn(MigrateStatus.ABANDONED);
+
         reportingJobRunner.run();
         assertThat(reportingJobRunner.isEnabled()).isFalse();
         verify(jobLauncher, never()).run(same(job), any(JobParameters.class));
+
+        // and test that it re-enables itself
+        reportingJobRunner.run();
+        assertThat(reportingJobRunner.isEnabled()).isTrue();
     }
```
malaffoon: Changes to these two tests are just improving coverage from 90% to 100%.

agorina:
Looks good, but I would feel comfortable if we could re-migrate AWS QA using this new logic.

PR #471, opened 2018-03-21
Add BOM handling to the ingest apps that read CSV files.

malaffoon:
Thanks to apache commons-io this is a simple fix. Most of the code is unit tests and minor cleanup of dependencies management.


group-processor/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'org.springframework.boot'
```
malaffoon: Mostly unrelated cleanup to make quotes consistent, alphabetize some of the dependencies, and use a common version for the apache commons libs.

jtreuting:
I'm guessing this means we don't need to update the CSV template file provided as this takes care of that?

malaffoon:
@jtreuting that's what i was trying to address, yes.

PR #473, opened 2018-03-27
fix audit tests

malaffoon:
The way the audit tests work, you have to use the data model to populate the tables, but the tests require that all fields are populated (so we don't miss any auditing). So the data model has to be updated in lock-step with the schema changes.

I'll be checking this in as soon as my local tests pass for a second time. But i will look at any comments.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcExamRepository.java
```diff
@@ -189,7 +193,7 @@ private long createExam(final Exam exam, final long importId) {
                 .studentId(row.getInt("student_id"))
                 .responsibleSchoolId(row.getInt("school_id"))
                 .ideaIndicator(row.getBoolean("iep"))
-                .lep(row.getBoolean("lep"))
+                .lep(getNullable(row, row.getBoolean("lep")))
```
malaffoon: not really part of this emergency patch but LEP is going to be optional

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/ExamRepositoryAuditIT.java
```diff
@@ -0,0 +1,637 @@
+package org.opentestsystem.rdw.ingest.processor.repository.impl;
```
malaffoon: I separated the audit ITs from the non-audit ITs. No real reason but i was getting ready to @Ignore the entire suite and this would've made it easier.

PR #476, opened 2018-03-30
log exceptions as errors instead of info

malaffoon:
The errors were being lost in all the other log messages.
I also pulled out the common `newRunnable` code.


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedStudentExamWriter.java
```diff
@@ -44,7 +43,7 @@
     private Student student;
 
     @Override
-    protected void setUp() {
+    protected void before() {
```
malaffoon: nit: i thought `before` was a better pairing with `after`

PR #254, opened 2017-07-04
Feature/dwr 443 checking data constraints

agorina:
What is included from the JIRA:
Revisit all builders in the exam-processor to check for valid conditions, including but not limited to:
- [x] null checks
- [x] data size, e.g. VARCHAR field limits
- [x] grade (this is meant for the asmt to test verification)
- [x] subject (same as above)
- [x] scale score min/max range
- [x] add check for non-null sessionId
- [x] Establish the pattern of validation and make it consistent.
- [x] These changes may apply to package processor as well. Either make those changes as part of this, or create a new issue if scope is too large.

What is left: 
- add more unit tests, primarily for the max length of the string values.
- review Builders for validation and remove if needed

PR #477, opened 2018-04-04
DWR-1652 Add fact table for longitudinal report

agorina:
I will not merge until I have all the RSTs modified, but would like to get the feedback.

This PR is related to https://github.com/SmarterApp/RDW_Schema/pull/254.


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -419,11 +419,8 @@ sql:
               UPDATE staging_exam
                 SET latest = true
               FROM (
-                     SELECT
-                       id,
-                       rank()
-                       OVER (PARTITION BY student_id, asmt_id, school_year ORDER BY completed_at DESC, id)
-                     FROM staging_exam where deleted = false
+                     SELECT id, rank() OVER (PARTITION BY student_id, asmt_id, school_year ORDER BY completed_at DESC, id)
+                      FROM staging_exam where deleted = false
```
agorina: Nothing changed, just reformatted.

migrate-olap/src/main/resources/application.yml
```diff
@@ -1,6 +1,10 @@
 app:
   school-year: 2018
 
+reporting:
+  high-school-start-grade-id: 9
+  high-school-accountability-grade-id: 11
```
agorina: A few questions/comments here:
1. Is `reporting` the right place for these? Or should it be part of `application`? 
2. I do not like using DB ids in here, but since I am using them in `>` or  `<` comparison, I am not sure  if there is a better way?
3. And please suggest shorter names (if you could think of some).

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -586,6 +583,112 @@ sql:
...
+        #
+        # The above rules are not enforced anywhere in the RDW.
+        # To enforce the above rule (in order to make the reporting accurate):
+        # 1. only migrate when enrolled grade is equal to asmt grade (for non high school) or 'accountability' grade (for high school)
+        # 2. migrate the latest Summative per subject and asmt grade (regardless of the school year)
+
+        fact_student_exam_longitudinal:
+          sql:
+            resetStagingLatestExam: >-
+              UPDATE staging_exam SET latest = false
```
malaffoon: Curious why this is necessary, isn't `latest` set to false by default when the table is populated? 

I'll probably understand as i read more of the review ...

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -586,6 +583,112 @@ sql:
...
+
+            updateStagingLatestExam: >-
+              UPDATE staging_exam
+                SET latest = true
+              FROM (
+                    SELECT se.id, rank() OVER (PARTITION BY student_id, a.subject_id, a.grade_id ORDER BY completed_at DESC, se.id)
+                     FROM staging_exam se JOIN asmt a on a.id = se.asmt_id
+                     WHERE se.deleted = false AND se.type_id = 3
+                       AND ( (a.grade_id < ${reporting.high-school-start-grade-id} AND a.grade_id = se.grade_id) OR (a.grade_id >= ${reporting.high-school-start-grade-id} AND se.grade_id = ${reporting.high-school-accountability-grade-id}))
```
malaffoon: Might want to put a warning somewhere that grade ids *must* be ordered the same as grade sequence. In other words, G9 must have an id less than G10, etc.

migrate-olap/src/main/resources/application.yml
malaffoon: It does feel like these are app-level concerns, like school-year. I don't like adding arbitrary top-level property scopes if we can avoid it (especially one that is redundant like this). Although it doesn't shorten the names you could scope them together. I don't like the grade-id thing either but at least we know the grade ids are explicitly set in scripts we control. 
```
app:
  high-school:
    start-grade-id: 9
    accountability-grade-id: 11
```

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -138,11 +138,16 @@ public Step upsertAsmtsStep() {
...
         final SqlListBuilder sqlsBuilder = new SqlListBuilder(stagingToReportingSqlConfiguration.getEntities());
-        sqlsBuilder.addNext("fact_student_exam", "updateStagingLatestExam")
+        sqlsBuilder
+                .addNext("fact_student_exam", "updateStagingLatestExam")
                 .addNext("fact_student_exam", "update")
                 .addNext("fact_student_exam", "insert")
                 .addNext("fact_student_iab_exam", "update")
-                .addNext("fact_student_iab_exam", "insert");
+                .addNext("fact_student_iab_exam", "insert")
+                .addNext("fact_student_exam_longitudinal", "resetStagingLatestExam")
```
agorina: @malaffoon , because of the `updateStagingLatestExam` on line 142, I need a `resetStagingLatestExam` here.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
agorina: I posted an answer in a place where it is used.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
jtreuting: we could add a sot column and use that for the > < checks for what comes before what

malaffoon: I agree with the over engineering comment. My vote, one of these:
1. Leave it as is since we "know" that those other grades aren't being used and that we made the id's correspond to the grades for all grades currently being used.
2. Specify the HS grade set instead of just the min. Then use IN and NOT IN to distinguish between HS and not HS.

The proper long-term solution would be to have an accountabilityGrade associated with every assessment (per tenant presumably) and then the query simplifies to just matching the accountability grade.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -586,6 +583,112 @@ sql:
...
+                  se.completeness_id,
+                  se.administration_condition_id,
+                  se.scale_score,
+                  se.performance_level,
+                  se.completed_at,
+                  se.updated,
+                  se.update_import_id,
+                  se.migrate_id
+                FROM staging_exam se
+                  LEFT JOIN fact_student_exam_longitudinal fe ON fe.student_id = se.student_id and fe.asmt_id = se.asmt_id and fe.school_year = se.school_year
```
agorina: I do not think this is right. Since we do not want to rely on the asmt id, this needs to be reworked. I will look at it when I am back.

agorina: I believe the right way is to add `subject_id` and `asmt_grade_id` to the `fact_student_exam_longitudinal` and use them in this JOIN (and the one below).

 @malaffoon, does it look right to you?

agorina: Thanks, I will continue with this change then.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -138,11 +139,17 @@ public Step upsertAsmtsStep() {
...
+                .addNext("fact_student_exam_longitudinal", "updateStagingLatestExam")
+                .addNext("fact_student_exam_longitudinal", "updateStagingExamWithAsmtData")
```
agorina: As mentioned in the https://github.com/SmarterApp/RDW_Schema/pull/256, I ran into 
`java.sql.SQLException: [Amazon](500310) Invalid operation: Target table must be part of an equijoin predicate;`. As a work around (to avoid JOIN on UPDATE), I added a step to updated asmt data in the staging table.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -586,6 +583,129 @@ sql:
...
+                FROM asmt a JOIN staging_exam se on se.asmt_id = a.id
+                WHERE se.type_id = 3 AND se.latest = true;
+
+            # iabs are loaded into its own fact table, that is why they are excluded here
+            insert: >-
+              INSERT INTO fact_student_exam_longitudinal (
+                id,
+                school_id,
+                student_id,
+                asmt_id,
```
agorina: `asmt_id` seems to be kind of useless here. But I have decided to keep it for the troubleshooting/ traceability purpose.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRST.java
```diff
@@ -71,6 +71,10 @@ public void itShouldUpsertAndDeleteCodesIfBatchHasImportContentWithCodes() throw
             if ("school_year".equals(table)) {
                 tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), table, -1, "year IN (1999)"));
                 tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), table, 1, "year IN (1998)"));
+            } else if ("grade".equals(table)) {
+                tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), "grade", -1, "id IN (109)"));
```
agorina: I have changed test grade values to positive numbers, hence this change.

agorina:
@malaffoon  and @jtreuting, please re-review this PR since I have made significant changes.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -584,6 +581,129 @@ sql:
...
+                  se.completeness_id,
+                  se.administration_condition_id,
+                  se.scale_score,
+                  se.performance_level,
+                  se.completed_at,
+                  se.updated,
+                  se.update_import_id,
+                  se.migrate_id
+                FROM staging_exam se
+                  LEFT JOIN fact_student_exam_longitudinal fe ON fe.student_id = se.student_id AND fe.asmt_grade_id = se.asmt_grade_id AND fe.subject_id = se.subject_id
```
agorina: This is the core change: use student, assessment grade and subject.

migrate-olap/src/test/resources/application-redshift.yml
```diff
@@ -3,6 +3,11 @@
 # override them locally but the CI system assumes the defaults are for CI resources.
 app:
   school-year: 1999
+  # this uses grade ids assuming that they match grade codes
+  high-school:
+    start-grade-id: 109
+    end-grade-id: 112
+    accountability-grade-id: 111
```
malaffoon: Whenever i think of configuring this i get concerned. Is there any way we can make these the grade codes and then look up (and cache) the actual ids? Or something like that? I just don't like that we'll have to tell the IT folks, "do a quick query to make sure the grade ids match the default values."  Yeah, i know the grade ids are likely to be as reliable as the grade codes but it just feels icky as it is.

agorina: Hmmm....I believe what I did was a feedback on the previous PR (unless I misunderstood it)?

To fully address you concern (once again assuming I read it right), we should do what @jtreuting suggested: add an order column to the grade table, and then use codes here and look up the order?

malaffoon: I think there are two different concerns here:
1. How to represent the grades in HS. I think your change to start/end is the attempt to deal with that. Although i still think that listing all the HS grades would work too.
2. How to configure the grades from #1. We could use the ids or the codes for that. That is what my most recent comment is about. And i don't feel super strongly about it, just communicating the concern.

agorina: Another way to address both concerns:
1. List all the high school grade codes.
2. Change the SQL to JOIN with the grade table and use `grade.code IN` or `NOT IN`.  

Just not sure if this is a priority for now. Debating if I should go ahead and implement it now or file a JIRA and do it later.

Would like to know what people think.

malaffoon: i'd vote for using `IN` and `NOT IN`. I know it won't make a practical difference though so whatever works.

agorina:
Merging but will address any follow up comments if needed.

PR #478, opened 2018-04-04
remove vacuum and analyze from finalize step

malaffoon:
I have also added a blurb to the deployment instructions to set up a task to analyze the redshift database periodically.

### DIFF
--- a/migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/FinalizeStepConfigRST.java
+++ b/migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/FinalizeStepConfigRST.java
@@ -64,6 +64,6 @@ public class FinalizeStepConfigRST {
 
         final SqlListExecutionStep argument = captor.getValue();
 
-        assertThat(argument.getSqls()).containsExactly("VACUUM", "ANALYZE");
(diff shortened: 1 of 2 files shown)


migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/FinalizeStepConfigRST.java
```diff
@@ -64,6 +64,6 @@ public void verifyFinalizeStepConfiguration() {
 
         final SqlListExecutionStep argument = captor.getValue();
 
-        assertThat(argument.getSqls()).containsExactly("VACUUM", "ANALYZE");
+        assertThat(argument.getSqls()).isEmpty();
```
malaffoon: I left the `finalize` step in, even though it is empty, so it would be easier to restore. Any strong opinions about that?

agorina: I think it is pretty easy to add this step if needed, I would remove it.

agorina: neat idea

PR #481, opened 2018-04-17
add curl and jq to image

malaffoon:
In our run book and troubleshooting docs we often suggest hitting a service's actuator end-points. This requires either port-forwarding or exec'ing into the pod, e.g.
```
$ kubectl exec -it import-service-xxx /bin/sh
# apk -U add curl jq
# curl http://localhost:8008/env | jq
# exit
```
This is a bit tedious. If curl were in the image, then the instructions become simpler and less error-prone:
```
$ kubectl exec -it import-service-xxx curl http://localhost:8008/env | jq
```

This change adds about 3.5MB to the image (e.g. 113MB -> 117MB), although about half of that is the apk update which i could remove. That is, "apk add -U ..." vs. "apk add ..."

Thoughts?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,7 +137,7 @@ subprojects {
             label([maintainer: 'Fairway Technologies'])
             copyFile("${jar.archivePath.name}", "${jar.archivePath.name}")
             exposePort(8008, 8080)
-            runCommand("apk add --no-cache tini")
+            runCommand("apk -U add --no-cache tini curl jq")
             entryPoint("/sbin/tini", "--")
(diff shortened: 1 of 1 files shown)


jtreuting:
what is this image used for?  each EC2 instance or only a particular one?

malaffoon:
Every "service" has an image. We have 12 services (7 in ingest, 5 in reporting) right now. Each service may have one or more instances. The image is used to create each instance (pod in kubernetes parlance).

malaffoon:
@wtritch did you just volunteer to add to the troubleshooting guide?  ;)

PR #482, opened 2018-04-18
Added school_year_asmt_grade_code to the olap migrade/fact_student_exam_longitudinal.

agorina:
Related to https://github.com/SmarterApp/RDW_Schema/pull/259


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -649,6 +651,7 @@ sql:
                   se.asmt_id,
                   se.subject_id,
                   se.asmt_grade_id,
+                  CONCAT(se.school_year::VARCHAR(4), CONCAT(',', se.asmt_grade_code::VARCHAR(2))),
```
malaffoon: i thought CONCAT could take any number of arguments?

agorina: I thought so too, but it was complaining. 
I may have done something wrong?

jtreuting: looks like not in Redshift: https://docs.aws.amazon.com/redshift/latest/dg/r_CONCAT.html

PR #484, opened 2018-04-20
add Authority awareness to import service security

malaffoon:
I'm trying to keep the PR's small so this just adds authorities to the security model. Still not really populating the permissions that will be needed for group file validation.

There is some functionality that is now common between ingest and reporting. For now, i'm not moving everything into RDW_Common, but i am thinking about parts of it.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionService.java
@@ -1,0 +1,31 @@
+package org.opentestsystem.rdw.ingest.auth;
+
+import org.opentestsystem.rdw.security.service.PermissionService;
+import org.opentestsystem.rdw.security.service.PermissionServiceException;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Service;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Wrap the basic permission service to cache the results since role/permissions don't change often.
+ */
+@Primary
+@Service
+class CachingPermissionService implements PermissionService {
+
+    private final PermissionService permissionService;
+
+    public CachingPermissionService(@Qualifier("basicPermissionService") final PermissionService permissionService) {
+        this.permissionService = permissionService;
+    }
+
+    @Cacheable("permissionsByRole")
+    public Map<String, Collection<String>> getPermissionsByRole() throws PermissionServiceException {
+        return permissionService.getPermissionsByRole();
+    }
+}
+
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
@@ -1,0 +1,113 @@
+package org.opentestsystem.rdw.ingest.auth;
+
+import com.google.common.collect.ImmutableSet;
+import org.opentestsystem.rdw.security.service.PermissionService;
+import org.opentestsystem.rdw.security.service.PermissionServiceException;
+import org.opentestsystem.rdw.utils.Grant;
+import org.opentestsystem.rdw.utils.TenancyChain;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.stereotype.Service;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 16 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionService.java
```diff
@@ -0,0 +1,31 @@
+package org.opentestsystem.rdw.ingest.auth;
```
malaffoon: Also stolen from reporting.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -1,10 +1,9 @@
 package org.opentestsystem.rdw.ingest.auth;
```
malaffoon: This simplifies a bit since it is no longer concerned with authorities, just converting tokens and caching them.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -69,24 +53,24 @@ public OAuth2AccessToken extractAccessToken(final String value, final Map<String
...
-            if ((hasText(client) && tenancyChain.hasRoleForClient(DataLoadAuthority, client))
-             || (hasText(state) && tenancyChain.hasRoleForState(DataLoadAuthority, state))) {
-                authorities.add(new SimpleGrantedAuthority(DataLoadAuthority));
-            } else {
-                logger.info("User {} doesn't have data load authority: {}", username, tenancyChain.toString());
-            }
+            authorities.addAll(authorityService.getAuthorities(tenancyChain));
+
+            // TODO - convert grants to Permissions (resolve organization ids, etc.)
+            final List<Permission> permissions = newArrayList();
```
malaffoon: Next step.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfiguration.java
```diff
@@ -0,0 +1,43 @@
+package org.opentestsystem.rdw.ingest.auth;
```
malaffoon: There are two completely separate configurations: the default and stub one for testing (thanks @wtritch for that suggestion).

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+class SecurityConfiguration {
+
+    @Bean
+    @ConfigurationProperties(prefix = "security.oauth2")
+    public ForgeRockTokenServices forgeRockTokenServices(final SbacTokenConverter tokenConverter) {
+        return new ForgeRockTokenServices(tokenConverter);
+    }
+
+    @Bean
+    public SbacTokenConverter sbacTokenConverter(final AuthorityService authorityService) {
```
malaffoon: Hmm, i don't think this needs to be in here. The class that changes is the AuthorityService not the token converter.

import-service/src/main/resources/application.yml
```diff
@@ -6,6 +6,11 @@ management:
   security:
     enabled: false
 
+app:
+  client: SBAC
+  state:
+    code: CA
```
malaffoon: trying to use the same property scope as other apps

import-service/src/main/resources/application.yml
```diff
@@ -24,11 +29,12 @@ archive:
   root: file:///tmp/
 
 security:
-  client: SBAC
-  state: CA
-  # cause StubTokenServices to be used
+  # "local" causes stub services to be used
   oauth2:
     token-info-url: local
+  permission-service:
```
esotrope: i dont think this has a `-` in it unless we fixed it

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * A role mapper that prepends "ROLE_" and creates a {@link SimpleGrantedAuthority}
+     */
+    public static Function<String, SimpleGrantedAuthority> simpleRoleAuthorityMapper =
+            role -> new SimpleGrantedAuthority(roleToAuthorityName(role));
+
+    /**
+     * A permission mapper that prepends "PERM_" and creates a {@link SimpleGrantedAuthority}
+     */
+    public static Function<String, SimpleGrantedAuthority> simplePermissionAuthorityMapper =
```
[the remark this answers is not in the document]
malaffoon: Just crafted here. I was toying with the idea of making the mapper flexible for the caller. You're right though, i should either make them public and make the method that uses them public (i.e. do it up right), or make all of that private.

import-service/src/main/resources/application.yml
malaffoon: Since i am doing the bean creation in my configuration class, i added the hyphen to make it more readable. I can take it out for consistency, but this won't be directly shared with the reporting configuration (it is scoped under `security` here).

PR #486, opened 2018-04-25
Added `sequence` to grade table and cut points to Redshift asmt.

agorina:
Adjusted SQL and test data.


agorina:
Merging, but will address comments if needed.

PR #490, opened 2018-05-02
tweak controllers to allow file upload to be processed as either a file or bytes

malaffoon:
This is WIP, just checking in because it doesn't hurt anything and i'm trying to avoid a big checkin.

@wtritch if you get a chance can you think about this one, or maybe we can just chat about it tomorrow? The idea is that most import content is handled by pushing the file contents onto the rabbit queue. But because we expect student groups to be large we want to archive it off and post a small message: it will be processed directly from the archive. This lets a controller choose which technique to use.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
@@ -48,7 +48,7 @@ class AccommodationController extends ScopedImportController {
     @ResponseStatus(HttpStatus.ACCEPTED)
     public RdwImportResource uploadImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
                                           final MultipartHttpServletRequest request) {
-        return super.uploadImport(sbacUser, request);
+        return super.uploadImport(sbacUser, request, false);
     }
 
     @GetMapping(value = "/imports")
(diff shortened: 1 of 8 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/web/AccommodationController.java
```diff
@@ -48,7 +48,7 @@ public RdwImportResource postImport(@AuthenticationPrincipal final SbacUserDetai
     @ResponseStatus(HttpStatus.ACCEPTED)
     public RdwImportResource uploadImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
                                           final MultipartHttpServletRequest request) {
-        return super.uploadImport(sbacUser, request);
+        return super.uploadImport(sbacUser, request, false);
```
malaffoon: Although i don't like meaningless booleans in methods, this is a helper method that is "nearby" and, after playing with a meaningful enum, i reverted to this.

import-service/src/test/java/org/opentestsystem/rdw/ingest/web/GroupControllerIT.java
```diff
@@ -51,7 +51,7 @@ public void itShouldUseServiceToImportGroupUpload() throws Exception {
         final RdwImport rdwImport = RdwImport.builder().id(123L).build();
         final MockMultipartFile file = new MockMultipartFile("file", "test.csv", TextCsv, body);
 
-        when(importService.importContent(any(), eq(body), eq(content), eq(TextCsv), any(), eq(Reprocess))).thenReturn(rdwImport);
+        when(importService.importContent(any(), eq(file), eq(content), eq(TextCsv), any(), eq(Reprocess))).thenReturn(rdwImport);
```
malaffoon: this was the point of the changes: instead of submitting the file body it submits the file itself

wtritch:
I almost wonder if we want to change *all* imports to do this...

malaffoon:
@wtritch i've been thinking about it; as i had the message handling code to the group processor i'll look into making a general behavior.

PR #493, opened 2018-05-09
Refactor the caching permission service test to avoid standing up the…

wtritch:
… entire application context

The problem was that it was using `@SpringBootTest` without limiting the scope, so the entire application context was being stood up, including the repositories.
We were also missing the `@ActiveProfiles("test")` annotation, which means it was attempting to connect the repositories to the `warehouse` schema rather than `warehouse_test`

I just re-worked the test to limit the application context to the classes needed for testing.

### DIFF
--- a/import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionServiceIT.java
+++ b/import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionServiceIT.java
@@ -1,5 +1,8 @@
 package org.opentestsystem.rdw.ingest.auth;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.opentestsystem.rdw.ingest.common.test.CachingTest;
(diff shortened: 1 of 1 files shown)


import-service/src/test/java/org/opentestsystem/rdw/ingest/auth/CachingPermissionServiceIT.java
```diff
@@ -1,32 +1,54 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
 
 @RunWith(SpringRunner.class)
-@SpringBootTest
+@SpringBootTest(classes = {
+        CachingPermissionService.class
+})
+@EnableCaching
```
wtritch: I know `@CachingTest` includes a transitory `@EnableCaching` annotation, but for whatever reason that wasn't actually registering interceptors without duplicating `@EnableCaching` here.

esotrope: 🤔

agorina: I think it enables it when it finds the following in the application.yml, and before since it was loading everything it was finding it:

```
spring:
   cache:
```

PR #495, opened 2018-05-10
Add migrate for fact_exam_claim_score.

agorina:


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -513,6 +513,74 @@ sql:
...
+                  se.updated,
+                  se.update_import_id,
+                  se.migrate_id
+                FROM staging_exam_claim_score scs
+                  JOIN staging_exam se ON se.id = scs.exam_id
+                  LEFT JOIN fact_exam_claim_score fe ON fe.student_id = se.student_id and fe.asmt_id = se.asmt_id and fe.school_year = se.school_year
+                WHERE se.type_id IN (1, 3) AND se.latest = true AND fe.exam_id IS NULL;
+
+            update: >-
+              SELECT 1;
```
agorina: This is not being executed, but it is needed since the configuration that loads it - expects it.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -513,6 +513,74 @@ sql:
...
+                  se.administration_condition_id,
+                  scs.category,
+                  se.completed_at,
+                  se.updated,
+                  se.update_import_id,
+                  se.migrate_id
+                FROM staging_exam_claim_score scs
+                  JOIN staging_exam se ON se.id = scs.exam_id
+                  LEFT JOIN fact_exam_claim_score fe ON fe.student_id = se.student_id and fe.asmt_id = se.asmt_id and fe.school_year = se.school_year
+                WHERE se.type_id IN (1, 3) AND se.latest = true AND fe.exam_id IS NULL;
```
agorina: I am debating if the check for the `type_id IN (1, 3)` is needed since if there is not data in the `staging_exam_claim_score` nothing will be migrated. Would like to know what people think about it.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -513,6 +513,74 @@ sql:
...
+
+            update: >-
+              SELECT 1;
+
+            deleteAsPartOfParentUpdate: >-
```
agorina: This deletes any 'master' updates.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -75,6 +75,7 @@ public Step deleteEntitiesStep() {
         final SqlListBuilder sqlsBuilder = new SqlListBuilder(stagingToReportingSqlConfiguration.getEntities());
         sqlsBuilder
                 .addNext("fact_student_exam", "delete")
+                .addNext("fact_exam_claim_score", "delete")
```
malaffoon: nit: seems like the delete of a "child" table should happen before the "master". Yeah, i know it isn't a real master/child relationship but logically it is.

agorina: Good point, I will flip the order.

agorina:
Merging, but will address comments if needed.

PR #496, opened 2018-05-11
Initial wiring and validation of Subject definition import payloads

wtritch:
Since Subject ingest is likely to be a large-ish task I'm breaking it into more than one PR.

This PR contains:
* the initial wiring to send the Subject XML from the import service to the package processor
* a registered subject handler within the package processor
* validation of the subject payload


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
```diff
@@ -66,7 +66,7 @@ public void configure(final ResourceServerSecurityConfigurer resources) {
 
             @Override
             public void configure(final HttpSecurity http) throws Exception {
-                final String contentRoot = "/{content:(exams|packages|accommodations|organizations|norms)}";
+                final String contentRoot = "/{content:(exams|packages|accommodations|organizations|norms|subjects)}";
```
wtritch: I'm using the same authorizations for the subjects endpoints as for the other import endpoints (other than groups.)  Does that sound reasonable?  I believe it requires ASMTDATALOAD permissions

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/SubjectController.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+//    @PostMapping("/imports")
+//    @ResponseStatus(HttpStatus.ACCEPTED)
+//    public RdwImportResource postImport(@AuthenticationPrincipal final SbacUserDetails sbacUser,
+//                                        @RequestHeader(value = CONTENT_TYPE, required = false) final String contentType,
+//                                        @RequestBody final byte[] body,
+//                                        @RequestParam final Map<String, String> requestParams) {
+//        return super.postImport(sbacUser, contentType, body, requestParams);
+//    }
```
wtritch: Right now a user POSTs an XML file, it gets written to the ArchiveService, and it's digest/location is sent to the package processor over Rabbit.
If I want to accept a byte[] payload and process it the same way it'll probably take some work in the shared ImportService.

So my question is: Is there a compelling reason to support this method other than to be consistent with our other import-service endpoints?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/WebMvcContext.java
```diff
@@ -1,14 +1,21 @@
...
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
 import java.util.List;
 
 @Configuration
 public class WebMvcContext extends WebMvcConfigurerAdapter {
 
+    @Override
+    public void configureContentNegotiation(final ContentNegotiationConfigurer configurer) {
+        configurer.defaultContentType(MediaType.APPLICATION_JSON);
```
wtritch: Now that we're loading the jackson xml library, it likes to occasionally return values from the controller as xml... this tells the Spring web framework that if the browser doesn't specify how they want the response data, use application/json.

package-processor/build.gradle
```diff
@@ -21,6 +23,7 @@ dependencies {
     compile 'commons-io:commons-io'
     compile 'mysql:mysql-connector-java'
     compile 'org.apache.commons:commons-csv'
+    compile 'org.opentestsystem.rdw.common:rdw-common-archive'
```
wtritch: The package processor now need to fetch the subject xml from the archive service

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ImportPayloadSink.java
```diff
@@ -23,4 +24,7 @@
 
     @Input(NORMS_INPUT)
     SubscribableChannel normsInput();
+
+    @Input(SUBJECT_INPUT)
+    SubscribableChannel subjectInput();
```
wtritch: This binds the package processor to the "SUBJECT" channel

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessor.java
```diff
@@ -60,6 +64,11 @@ public void processNorms(final Message<?> message) {
         handleIncomingMessage(message, normsProcessor);
     }
 
+    @ServiceActivator(inputChannel = ImportPayloadSink.SUBJECT_INPUT)
+    public void processSubject(final Message<?> message) {
+        handleIncomingMessage(message, subjectProcessor);
+    }
```
wtritch: This sends SUBJECT messages to the processing service

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorApplication.java
```diff
@@ -3,12 +3,19 @@
...
 
 /**
  * App for package processor
  */
-@SpringBootApplication
+@SpringBootApplication(exclude = {
+        ElastiCacheAutoConfiguration.class,
+        MailSenderAutoConfiguration.class,
+        ContextStackAutoConfiguration.class
+})
```
wtritch: I had some fun issues running the package processor locally.  It kept trying to connect to AWS services.  I took this from the import-service configuration.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationContext.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * This class is responsible for maintaining the current state of a
+ * Subject validation.  It tracks the current "location" as well as stores
+ * and collates any validation failure messages.
+ */
+public class SubjectValidationContext {
```
wtritch: The SubjectValidationContext gets passed around during the validation process.  It represents the current validation state.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+        final String location = new LocationStrategy.ImportContentLocationStrategy(SUBJECT).location(digest);
+        try (final InputStream subjectData = archiveService.openResource(location)) {
+            final Subject subject = serializationService.parseSubject(subjectData);
+
+            final SubjectValidationContext validationContext = validate(subject);
+            if (!validationContext.isValid()) {
+                throw new ImportException(BAD_FORMAT, validationContext.toJson());
+            }
+
+            //DO STUFF HERE
```
wtritch: This is the big blinky TODO where we actually do stuff with the validated Subject model.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This validator is responsible for validating {@link ItemClaim} models as well
+ * as their child {@link Target} models.
+ */
+@Component
+public class ClaimValidator implements SubjectValidator {
```
wtritch: I decided to break validation into a collection of focused validators to avoid having a single ginormous "validate" method.  (This one validates claims!)

package-processor/src/main/resources/application.yml
```diff
@@ -9,6 +9,23 @@ server:
...
+      static: us-west-2
+    # disable CloudFormation stuff
+    stack:
+      auto: false
+    credentials:
+      instance-profile: false
+
+# use LocalArchiveService by default
+archive:
+  root: file:///tmp/
```
wtritch: Now that the package-processor is interacting with S3, it needs some aws creds.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
```diff
@@ -0,0 +1,75 @@
...
+                    && claim.getTargets() != null
+                    && !claim.getTargets().isEmpty()) {
+                context.failure("A non-organizational claim cannot contain targets");
+            }
+
+            if (claim.isOrganizational()) {
+                validateTargets(context, claim.getTargets());
+            }
+
+            context.pop();
```
agorina: I may have missed it, but I do not see a validation for the max allowed number of storable claims.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/PerformanceLevelValidator.java
```diff
@@ -0,0 +1,62 @@
...
+            context.push("AssessmentType", assessmentType.getCode());
+            validatePerformanceLevels(context, assessmentType.getPerformanceLevels());
+
+            if (assessmentType.getSubScoring() != null) {
+                context.push("SubScoring");
+                validatePerformanceLevels(context, assessmentType.getSubScoring().getPerformanceLevels());
+                context.pop();
+            }
+            context.pop();
+        }
```
agorina: Ditto about max number of performance levels.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SecurityConfigurer.java
jtreuting: yeah that seems reasonable,  the role `ASMTDATALOAD` is not super accurate anymore but that is something that happened before this one.  it fits with the other types of imports listed there i'd say.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/SubjectController.java
malaffoon: We can discuss but this is sort of left-over from the very beginnings when we'd post TRT payloads with raw data instead of a file upload. Apart from developer convenience i don't really see a need to support this anymore for any of our payloads. And the developer convenience isn't that hard to work around. If i were doing it from scratch i'd support only the file upload by default and then add special code in the exam controller/service to accept byte[] payload.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessorApplication.java
malaffoon: That's what you get for making processors less isolated  :P

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationContext.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return The current failure messages and locations
+     */
+    public List<SubjectValidationFailure> getFailures() {
+        return FailureOrdering.sortedCopy(failures);
+    }
+
+    /**
+     * @return A JSON representation of the current failure messages and locations, limited to 100 failure messages
+     */
+    public String toJson() {
```
malaffoon: I guess it doesn't matter a lot but ...

Having to add the ObjectMapper and Json processing awareness to this context just to format the messages seems to cross concerns. You could put this formatting stuff in a utility method outside the context, keeping the context cleaner. Of course, if that utility method lives right beside this class i guess it isn't really improving things a lot.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationFailure.java
```diff
@@ -0,0 +1,34 @@
+package org.opentestsystem.rdw.ingest.processor.model;
+
+/**
+ * This class represents a validation failure.  It contains
+ * both the location of the failure within the Subject model
+ * as well as a detailed error message.
+ */
+public class SubjectValidationFailure {
```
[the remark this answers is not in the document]
malaffoon: Will and i talked a bit about this offline. I agree that we could've used DataElementErrorCollector here instead, and i poked Will a bit about it. But the ParserHelper isn't really appropriate since this validation works on POJOs instead of serialized content.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationFailure.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private String location;
+    private String message;
```
malaffoon: `final`?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
malaffoon: ~~public~~

malaffoon:
Although you made changes so you didn't *have* to have XML to test stuff, i'm a little surprised there isn't a test with raw XML as input.

package-processor/src/main/resources/application.yml
malaffoon: Please add a note to https://github.com/SmarterApp/RDW/blob/develop/deploy/Upgrade.v1_2.AWS.md about that.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationFailure.java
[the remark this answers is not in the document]
agorina: I thought we would need to support some level of 'update' to the subjects, and that may require validation against the serialized content.  

As I said before I just feel that it is a little confusing to have two different models for the same concern within the same project.

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/SubjectController.java
malaffoon: If we do decide that file-based import is the norm, i wonder if we should be adding `filename` to the import table? Right now, the filename is stuffed into the `batch` column in some special circumstances which makes sense to me but one could imagine that a client would want to identify a batch of files they upload so we'd need two columns. 

Also, if we decide to get rid of the byte[] end-point, we need to update the API docs (it will simplify them a bit which would make them a bit more readable).

wtritch: So it's only the exam controller that actively uses the byte[] end-point in production?

package-processor/src/main/resources/application.yml
wtritch: Filed a PR here https://github.com/SmarterApp/RDW/pull/31

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationFailure.java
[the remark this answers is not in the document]
wtritch: I'm not sure what you mean about 'update' and serialized content?
Hm... I suppose I could extend the DataElementErrorCollector to be more hierarchy-aware.  Right now it's a flat list of element -> error, but that won't work with the subject payload where the same element (PerformanceLevel) can exist under different parents.  I could also likely re-using the DataElementError rather than introduce a SubjectValidationFailure.  As Mark said, the ParserHelper isn't very useful to me, since I'm validating a model rather than XML.  Let me play with it and see how it falls out.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
wtritch: doh, added

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/PerformanceLevelValidator.java
wtritch: added

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
wtritch: doh, fixed all these

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+        if (scorableClaimCount > 6) {
```
agorina: Should this be configurable? 

I added this into the application.yml in reporting:
```
reporting:
  subjects:
    max-claims: 6
```

wtritch: ...not if we're going to be modifying our schema to only have 6 columns, right?

import-service/src/main/java/org/opentestsystem/rdw/ingest/web/SubjectController.java
malaffoon: I'm pretty sure nobody uses the byte[] end-point in production.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/SubjectValidationFailure.java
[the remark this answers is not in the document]
agorina: Thank you!

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/validator/ClaimValidator.java
wtritch: Fixed, bound to a new RDW_Common AssessmentBoundaries value

agorina: Thank you. And I did the same on reporting: https://github.com/SmarterApp/RDW_Reporting/pull/1292

PR #497, opened 2018-05-14
misc cleanup of group processing

malaffoon:


group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/GroupProcessor.java
```diff
@@ -41,10 +41,8 @@ public void process(final Message<?> message) {
             final String digest = new String((byte[]) message.getPayload(), UTF_8);
             groupMessageProcessor.process(digest, importId);
 
-            // TODO - need to offset the updated timestamp on the import record? Or sleep (ugh, no!)
```
malaffoon: This is now taken care of (see below)

group-processor/src/main/resources/application.sql.yml
```diff
@@ -203,3 +202,8 @@ sql:
     DELETE FROM upload_student_group
     WHERE import_id = :import_id
 
+  # As noted above, the updated values are randomly shifted to avoid overloading migrate.
+  # So we need to pause a bit before moving on and updating the import record. Since the
+  # maximum random shift is 1100us, sleeping for 10ms will be more than enough.
+  wait: >-
+    SELECT SLEEP(0.010);
```
malaffoon: Okay, here's the ugly. Feel free to mock and laugh, you won't be the first. But, honestly, i couldn't find a better solution. This keeps the "fix" near the "cause" (both in these SQL statements). It doesn't propagate the ugliness, i.e. no change to ImportRepository, etc. 

Yes, the correct solution is to change migrate to be more like the OLAP version of migrate. Lots of work though, maybe later.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
```diff
@@ -66,21 +75,39 @@ public DefaultAuthorityService(final PermissionService permissionService,
...
                 final PermissionScope.Builder builder = builders.computeIfAbsent(permission, p -> PermissionScope.builder());
 
                 switch(grant.getLevel()) {
                     case CLIENT:
+                    case STATE_GROUP:
                     case STATE:
                         builder.statewide(true);
                         break;
-                    // TODO - deal with all organization levels, not just client/state
-                    // TODO   need an organization repository
```
malaffoon: TODOne. Copied in part from reporting, thanks @esotrope.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/ExpiringAuthenticationCache.java
```diff
@@ -0,0 +1,75 @@
+package org.opentestsystem.rdw.ingest.auth;
```
malaffoon: This is unrelated to recent changes (although it is change-adjacent). Basically, i pulled this out so i could use it in both the production and stub token services, providing better test coverage, etc.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacUser.java
```diff
@@ -58,6 +58,10 @@ public PermissionScope getPermissionScope(final String id) {
         return PermissionScope.EMPTY;
     }
 
+    public List<Permission> getPermissions() {
+        return permissions;
+    }
```
malaffoon: Added this method so that the permissions would marshal when hitting the /imports/user end-point (a diagnostic helper).

task-service/src/main/java/org/opentestsystem/rdw/ingest/tasking/repository/OrganizationDataRepository.java
```diff
@@ -3,7 +3,7 @@
 /**
  * Source of organization data.
  */
-public interface OrganizationRepository {
+public interface OrganizationDataRepository {
```
malaffoon: I renamed this to avoid conflicting with the common OrganizationRepository. This interface represents the data that can be pulled from ART (as opposed to the interface for getting School/District model objects from the database). Many of the modified files are due to just this name change.

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -32,13 +32,15 @@
     private static final List<String> ValidDelimiter = newArrayList(",", "\\t", "|");
 
     @Value("${sql.load-csv}")
-    public String rawSql;
+    String rawSql;
 
     @Value("${sql.cleanup}")
-    public String cleanupSql;
+    private String cleanupSql;
```
malaffoon: not really sure why these were public?

group-processor/src/main/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingLoadService.java
```diff
@@ -98,11 +100,9 @@ else if (uri.startsWith(localPrefix)){
...
     public void cleanup(final long importId) {
-        final SqlListBuilder sqlListBuilder = new SqlListBuilder(sqlConfiguration.getEntities());
-        final List<String> sql = sqlListBuilder
-                .addNext(cleanupSql)
-                .build();
+        repository.execute(new SqlListBuilder(sqlConfiguration.getEntities()).addNext(cleanupSql).build(),
+                ImmutableMap.of("import_id", importId));
 
-        repository.execute(sql, ImmutableMap.of("import_id", importId));
+        repository.execute(new SqlListBuilder(sqlConfiguration.getEntities()).addNext(waitSql).build());
```
malaffoon: Annoying but execute(String, Map) calls template.update which doesn't work with SELECT statements. While execute(String) calls template.execute which does work with SELECTs.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
```diff
@@ -66,21 +75,39 @@ public DefaultAuthorityService(final PermissionService permissionService,
     @Override
     public Set<Permission> getPermissions(final TenancyChain tenancyChain) {
         final Map<String, PermissionScope.Builder> builders = newHashMap();
+        final List<Grant> grants = getGrants(tenancyChain);
 
+        // get reference data needed to map roles to permissions
         final Map<String, Collection<String>> permissionsByRole = getPermissionsByRole();
-        for (final Grant grant : getGrants(tenancyChain)) {
-            // TODO - should we stick role into set as well? don't think it is needed/expected
+        final Map<Grant.Level, Map<String, Long>> orgMap = getOrganizationsById(grants);
```
malaffoon: NOTE: i am not caching the organizations. My concern is that schools/districts do change often enough that caching will be problematic. And the token is cached for up to 15 minutes, so this won't be called too often.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
esotrope: not sure what i did to get the TY here but your're welcome 😉

group-processor/src/main/resources/application.sql.yml
agorina: Who was the first one laughing? 😉

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
esotrope: I'll take credit if it works and blame Rob and Chris if it doesn't ; J

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/LocationStrategy.java
```diff
@@ -1,64 +0,0 @@
-package org.opentestsystem.rdw.ingest.common.util;
```
malaffoon: moved to RDW_Common

PR #499, opened 2018-05-15
DO NOT MERGE: WIP/POC for ingesting subject definitions

wtritch:
I wanted to get some feedback on my approach for ingesting a subject payload.  Please comment on any issues you see!


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
```diff
@@ -1,14 +1,50 @@
...
-     * Finds an id of the subject by its code
+     * Finds an id of the organizational claim by its code
      *
-     * @param code the code of the subject
+     * @param subjectId The subject id
+     * @param code      the code of the claim
      * @return the id, null if code not found
      */
-    Integer findIdByCode(String code);
+    Integer findIdByCode(int subjectId, String code);
```
wtritch: When we had a known list of claims for ELA and Math we could *know* that claims for ELA and claims for Math never used the same code: ("1-LT" vs "1" or "2-W" vs "2") but now that we're allowing user-defined subjects, we need to scope finding a claim by code to include the subject id.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/SubjectAssessmentTypeRepository.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    void delete(String subjectCode, String assessmentTypeCode);
+
+    /**
+     * Test if the given subject assessment type has reported exams.
+     *
+     * @param type  The subject assessment type
+     * @return True if exams of the given subject and assessment type exist
+     */
+    boolean hasExams(WarehouseSubjectAssessmentType type);
```
wtritch: This is used to determine if there will be a ripple-effect when modifying a subject's assessmentType's number of performance levels.
I have some concerns about potentially locking the `exam` table for the duration of the subject ingest transaction...  Alternatively, we could take a half-step back and reject the change IFF there are any assessments loaded that reference the subject/assessmentType in question?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectProcessor.java
```diff
@@ -52,7 +66,11 @@ public String process(final byte[] payload, final long importId) throws ImportEx
                 throw new ImportException(BAD_FORMAT, validationContext.toJson());
             }
 
-            //DO STUFF HERE
+            subjectService.upsert(subject, importId);
+            subjectAssessmentTypeService.upsert(subject);
+            claimService.upsert(subject);
+            //TODO targetService, scorableClaimService, standardService, translationService
```
wtritch: This is where we pass the Subject definition to the individual services with bite-sized responsibilities.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
```diff
@@ -1,14 +1,50 @@
...
+    void delete(String subjectCode, String claimCode);
+
+    /**
+     * Test if a claim has associated assessment items.
+     *
+     * @param subjectCode   A subject code
+     * @param claimCode     A claim code
+     * @return True if the claim is associated with assessment items
+     */
+    boolean hasItems(String subjectCode, String claimCode);
```
wtritch: Similar to SubjectAssessmentTypeRepository#hasExams, this is used to determine if we can remove a claim. (We cannot remove a claim that has assessment items referencing it.)
I have fewer reservations about potentially locking the assessment or item tables during the subject ingest transaction than I do with potentially locking the exam table.  Thoughts?

package-processor/src/main/resources/application.sql.yml
```diff
@@ -108,8 +103,43 @@ sql:
...
+        WHERE s.code = :subject_code
+
+    delete: >-
+      DELETE c FROM claim c
+        JOIN subject s ON s.id = c.subject_id
+      WHERE s.code = :subject_code
+        AND c.code = :claim_code
+
+    hasItems: >-
+      SELECT EXISTS(
```
wtritch: I know `EXISTS` isn't super-performant in MySQL... any suggestions @agorina ?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseSubjectAssessmentType.java
```diff
@@ -0,0 +1,147 @@
...
+
+import org.opentestsystem.rdw.common.model.subject.SubjectAssessmentType;
+
+import java.util.Objects;
+
+/**
+ * This model represents the definition of an AssessmentType's variables
+ * within the context of a Subject.
+ */
+public class WarehouseSubjectAssessmentType {
```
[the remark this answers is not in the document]
agorina: Based on your comment: `SubjectAssessmentTypeDefinition`?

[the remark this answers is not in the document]
malaffoon: I'm fine with `Warehouse` prefix.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
```diff
@@ -9,25 +9,84 @@
...
+
     @Autowired
     public JdbcClaimRepository(final NamedParameterJdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
     }
 
     @Override
     @Cacheable(value = "claim")
-    public Integer findIdByCode(final String code) {
+    public Integer findIdByCode(final int subjectId, final String code) {
```
agorina: Do we keep it 'Cacheable'?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcSubjectRepository.java
```diff
@@ -7,27 +7,64 @@
...
+
     @Autowired
-    JdbcSubjectRepository(final NamedParameterJdbcTemplate jdbcTemplate) {
+    public JdbcSubjectRepository(final NamedParameterJdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
     }
 
     @Override
     @Cacheable(value = "subject")
     public Integer findIdByCode(final String code) {
```
agorina: Ditto about 'cacheable'.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
malaffoon: I think locking the exam table (in the warehouse) has the same impact as locking the item table.

Yeah, locking the assessment table will have little or no effect since it is seldom updated/migrated.

package-processor/src/main/resources/application.sql.yml
agorina: It is _selectively_ performant. :)
And I do not think I have a better suggestion.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcSubjectRepository.java
```diff
@@ -7,27 +7,64 @@
...
+    @Override
+    public Integer create(final String code, final long importId) {
+        final KeyHolder keyHolder = new GeneratedKeyHolder();
+        jdbcTemplate.update(sqlCreateSubject,
+                new MapSqlParameterSource()
+                        .addValue("code", code)
+                        .addValue("import_id", importId),
+                keyHolder);
+
+        return keyHolder.getKey().intValue();
```
malaffoon: maybe a TODO to flush the cache used above?

package-processor/src/main/resources/application.sql.yml
```diff
@@ -229,3 +259,70 @@ sql:
 
       delete: >-
         DELETE FROM percentile_score WHERE percentile_id=:percentile_id
+
+  subject:
+    findIdByCode: >-
+      SELECT id FROM subject WHERE code = :code
+
+    insert: >-
+      INSERT INTO subject (code, import_id, update_import_id) VALUES (:code, :import_id, :import_id);
```
agorina: Nit: I have learned today that it you want to do 'batchUpdate', it does not like semicolon at the end.
I am assuming you are inserting one at a time here, but I would still remove it.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/SubjectClaimService.java
```diff
@@ -0,0 +1,21 @@
+package org.opentestsystem.rdw.ingest.processor.service;
+
+import org.opentestsystem.rdw.common.model.subject.Subject;
+
+/**
+ * This service is responsible for processing upserts to
+ * organizational subject claims.
+ */
+public interface SubjectClaimService {
```
malaffoon: curious why this is a service, not a repo, but i suppose i'll find out if i keep reading ...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectProcessor.java
malaffoon: Ok, i see now.

package-processor/src/main/resources/application.sql.yml
malaffoon: I think this will be fine ... it is just converting a NULL/1 into a boolean.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
wtritch: It *is* still cacheable (the tests validate that.) As to whether or not it *should* be... I don't have enough of a grasp of the TRT parsing assessment package processing to say whether it makes a large difference... but since we don't ingest assessment packages very often, I'd be ok removing the `@cacheable` Thoughts?

package-processor/src/main/resources/application.sql.yml
wtritch: doh, how'd that sneak in there?  Removed

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcSubjectRepository.java
wtritch: hm... good point.  I'll either need to flush the cache (for the associated key) or we can remove the caching altogether if we don't think it provides a lot of value...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/SubjectClaimService.java
wtritch: Because I'm stuffing business logic in here =)  (well, maybe not this one, but definitely in its siblings)

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectAssessmentTypeService.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Delete the given subject assessment type definition.
+     * NOTE that this only allows deleting the definition IFF there are no existing exams
+     * for the target subject and assessment type.
+     *
+     * @param type The type definition to delete
+     */
+    private void delete(final WarehouseSubjectAssessmentType type) {
+        if (repository.hasExams(type)) {
```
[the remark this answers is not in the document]
wtritch: shoot, good point.  In that case, we don't even need to test for exams.  We can stop at testing for assessment existence.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectClaimService.java
```diff
@@ -0,0 +1,49 @@
...
+
+        for (final String submitted : submittedClaims) {
+            if (existingClaims.remove(submitted)) continue;
+
+            repository.create(subjectCode, submitted);
+        }
+
+        for (final String existing : existingClaims) {
+            if (repository.hasItems(subjectCode, existing)) {
+                throw new IllegalArgumentException("Cannot claim: " + existing + " due to existing assessment items");
```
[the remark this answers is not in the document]
wtritch: doh, fixed

wtritch:
@agorina at the end I'll add an "end-to-end" IT with a sample XML payload but right now I'm focusing on unit tests to hit my branches

agorina:
@wtritch, makes sense, thank you for feedback!

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectAssessmentTypeService.java
```diff
@@ -0,0 +1,107 @@
...
+                        existing.getAssessmentTypeCode());
+            }
+            if (!Objects.equals(existing.getSubScorePerformanceLevelCount(), updated.getSubScorePerformanceLevelCount())) {
+                throw new IllegalArgumentException("Unable to update subject assessment type sub score performance level counts due to existing exams for subject: " +
+                        existing.getSubjectCode() +
+                        " assessment type: " +
+                        existing.getAssessmentTypeCode());
+            }
+        }
+        repository.update(updated);
```
agorina: The check above *does not guaranty* that this update happens *when there is no exams*. There is a timing issue, and 'transactional' does not mean 'synchronized' (at least with the transaction lock level we are using).  Another transaction could insert an exam in between of the time gap.

So, my question is: what will happen if there are exams? Will it just fail, or may it cause more catastrophic issues?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcSubjectRepository.java
agorina: Right, that was my question about 'Cacheable' on the repositories.
I believe Chris did a research at some point with the conclusion that it could be removed, but I am not exactly sure.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTargetService.java
```diff
@@ -0,0 +1,67 @@
...
+        for (final ItemClaim claim : subject.getClaims()) {
+            if (!claim.isOrganizational()) continue;
+
+            final List<Target> submitted = claim.getTargets();
+            if (submitted == null) continue;
+
+            for (final Target target : submitted) {
+                if (existing.remove(getKey(claim, target)) != null) continue;
+
+                repository.create(subject.getCode(), claim.getCode(), target.getCode());
```
agorina: I  would suggest batching targets here instead of doing one at a time.

And I do not want to repeat the same comment in different places, but I think it is better to batch multiple operations instead of looping one at a time.

wtritch: Good call, since we could be potentially dealing with a large number of targets I switched the interface to work against collections.  I'll do the same for standards, since they could also have large numbers.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectAssessmentTypeService.java
wtritch: I changed the above to check for the existence of assessments rather than exams.  And assessments *must* exist before we start receiving results for them.  And if a user is simultaneously (within the scope of the transaction) submitting a subject definition while submitting assessment definitions nothing should break.  I believe we'll only reference the counts during the pivoting/migration to reporting and olap.

agorina: This is good. Thank you!

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/TargetRepository.java
```diff
@@ -1,15 +1,52 @@
...
+     */
+    void delete(Collection<Integer> targetIds);
+
+    /**
+     * Test any of the given targets are referenced by any assessment items.
+     *
+     * @param targetIds The target ids to test
+     * @return The first target id that is reference by an assessment item
+     */
+    Integer hasItems(Collection<Integer> targetIds);
```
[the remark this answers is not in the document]
agorina: You are just being too nice :)

wtritch:
Closing in favor of a clean PR

PR #501, opened 2018-05-17
improve message for bad date

malaffoon:
The parser helper has a helper function to parse dates. It requires the format to be ISO-8601 whatever. It uses the LocalDate parse exception message which is a bit cryptic, something like `"Text \u00278/1/17\u0027 could not be parsed at index 0"`. Since this is the third or fourth time i've stared at the message and wondered what `\u0027` is and confused `index 0` for the line offset (it's really the field text offset) i thought i'd clean it up. It now emits something like `"invalid date [8/1/17], use format YYYY-MM-DD"`.

Since this is a teeny change, i'll commit and address any concerns raised.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
@@ -5,6 +5,7 @@ import org.springframework.util.StringUtils;
 import javax.validation.constraints.NotNull;
 import java.time.LocalDate;
 import java.time.Year;
+import java.time.format.DateTimeParseException;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -135,7 +141,7 @@ public String validate(final String name, final String rawValue, final int maxLe
             StringUtils.hasText(value) ? Integer.parseInt(value.trim()) : null;
 
     public static final Function<String, LocalDate> toLocalDateOrNull = value ->
-            StringUtils.hasText(value) ? LocalDate.parse(value.trim()) : null;
+            StringUtils.hasText(value) ? toLocalDate.apply(value.trim()) : null;
```
malaffoon: Avoid code duplication by reusing the function.

PR #502, opened 2018-05-19
Adjusted to match Redshift schema refactoring.

agorina:
Related to https://github.com/SmarterApp/RDW_Schema/pull/275

### DIFF
--- a/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
+++ b/migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
@@ -421,8 +421,8 @@ sql:
                     school_group.id in (SELECT ss.school_group_id from staging_school ss WHERE ss.deleted = 1)
                     AND NOT EXISTS(SELECT rs.id from school rs WHERE rs.school_group_id = school_group.id);
 
-      # ------------ fact_student_exam -----------------------------------------------------------------------
-        fact_student_exam:
+      # ------------ exam -----------------------------------------------------------------------
+        exam:
           sql:
             # this updates all types of the exams so we do not need to repeat it for iab
             updateStagingLatestExam: >-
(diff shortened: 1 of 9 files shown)


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -545,47 +535,37 @@ sql:
...
                   se.completed_at,
                   se.updated,
                   se.update_import_id,
                   se.migrate_id
                 FROM staging_exam_claim_score scs
                   JOIN staging_exam se ON se.id = scs.exam_id
-                  LEFT JOIN fact_exam_claim_score fe ON fe.student_id = se.student_id and fe.asmt_id = se.asmt_id and fe.school_year = se.school_year
-                WHERE se.type_id IN (1, 3) AND se.latest = true AND fe.exam_id IS NULL;
+                  LEFT JOIN exam_claim_score e ON e.student_id = se.student_id and e.asmt_id = se.asmt_id and e.school_year = se.school_year
+                WHERE se.type_id IN (1, 3) AND se.latest = true AND e.exam_id IS NULL;
```
agorina: It bothers me a little that the way we migrate does not enforce that the exam id in here exists in the exam table.  But since the migrate rules are exactly the same, they should match....

Since Redshift does not enforce FK, I do not know how we could enforce it. Open to suggestions.

jtreuting: thinking about the worst case of records get into the exam_claim_score and not into the exam table, they still won't be included in the results as if they weren't migrated at all.  so even if we could enforce it there isn't much of a difference in the outcome right?

agorina: Yeah, the worst case the two tables get out of synch. 

But on the other hand it is possible with all the tables in Redshift.

PR #276, opened 2017-07-19
DWR-557 Migrate did not migrate a student when gender is null

agorina:
When a student is create as part of the student group CSV upload, the gender may be null.

PR #503, opened 2018-05-21
Feature/ingest subject definition

wtritch:
This PR introduces subject ingest into a combination of existing, new, and modified repositories/db tables.
See https://github.com/SmarterApp/RDW_Schema/pull/276 for associated schema changes


exam-processor/src/main/resources/application.sql.yml
```diff
@@ -71,7 +71,12 @@ sql:
 
     subjectClaimScore:
       findBySubjectAndAsmtType: >-
-        SELECT id, code FROM subject_claim_score WHERE subject_id = :subject_id and asmt_type_id =:asmt_type_id
+        SELECT scs.id, scs.code
+        FROM subject_claim_score scs
+          JOIN subject_asmt_type sat ON sat.subject_id = scs.subject_id
+        WHERE scs.subject_id = :subject_id
+          AND sat.asmt_type_id = :asmt_type_id
+          AND sat.sub_score_performance_level_count > 0
```
agorina: I know what is happening here, but I think it is confusing. If the 'subject_claim_score' is *not* bound to an assessment type, then why are we searching by it? 

It would be more understandable if we break it into 2 steps:
1. does this asmt type supports claims (AKA `sub_score`)?
2. if the answer to the above is yes - then find them.

package-processor/build.gradle
```diff
@@ -30,4 +32,6 @@ dependencies {
 
     compile project(':rdw-ingest-common')
     testCompile project(path: ':rdw-ingest-common', configuration: 'tests')
+
+    runtime files(Jvm.current().toolsJar)
```
malaffoon: nit: can you add a comment for this, i've already forgotten what it is for

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseItemDifficulty.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * This class represents item difficulty cutpoints within the context
+ * of a subject and grade.
+ */
+public class WarehouseItemDifficulty {
+
+    private Integer id;
+    private String gradeCode;
+    private Float moderateLowEnd;
+    private Float difficultLowEnd;
```
malaffoon: I think we should use `Double` in our model.

malaffoon: Can these be null?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseSubjectAssessmentType.java
```diff
@@ -0,0 +1,147 @@
+package org.opentestsystem.rdw.ingest.processor.model;
+
+import org.opentestsystem.rdw.common.model.subject.SubjectAssessmentType;
+
+import java.util.Objects;
+
+/**
+ * This model represents the definition of an AssessmentType's variables
```
malaffoon: nit: s/variables/attributes

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
```diff
@@ -1,14 +1,50 @@
...
 
+import java.util.Set;
+
 /**
  * Repository for dealing with the claims
  */
 public interface ClaimRepository {
     /**
-     * Finds an id of the subject by its code
+     * Finds an id of the organizational claim by its code
```
malaffoon: nit: s/organizational/subject ??  not sure on this cause i don't know what is meant be "organizational" here

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
```diff
@@ -1,14 +1,50 @@
...
+    void delete(String subjectCode, String claimCode);
+
+    /**
+     * Test if a claim has associated assessment items.
+     *
+     * @param subjectCode   A subject code
+     * @param claimCode     A claim code
+     * @return True if the claim is associated with assessment items
+     */
+    boolean hasItems(String subjectCode, String claimCode);
```
malaffoon: nit: i think we talked about this offline but i don't like "hasItems" ... problem is i can't come up with a better name. Maybe `isReferenced` or `isUsed`?  Or is it important that it is items that reference it?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
```diff
@@ -13,4 +17,44 @@
      * @return the id if it exists, null if it doesn't
      */
     Integer findIdByLevelAndSubject(int level, int subjectId);
```
malaffoon: Are the `int subjectId` methods going to go away at some point, preferring `String subjectCode`?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
```diff
@@ -13,4 +17,44 @@
...
+
+    /**
+     * Find all depth of knowledge definitions for the given subject.
+     *
+     * @param subjectCode A subject code
+     * @return  The depth of knowledge definitions for the given subject
+     */
+    Collection<WarehouseDepthOfKnowledge> findBySubject(final String subjectCode);
```
malaffoon: nit: ~~final~~

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StandardRepository.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    void delete(Collection<Integer> standardIds);
+
+    /**
+     * Test if any of the standards are referenced by any assessment items.
+     *
+     * @param standardIds   The standard ids to test
+     * @return The first standard id that is reference by an assessment item
+     */
+    Integer hasItem(Collection<Integer> standardIds);
```
malaffoon: `hasItems`?  But see previous note about what we really want to call these methods.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/SubjectTranslationRepository.java
```diff
@@ -0,0 +1,43 @@
+package org.opentestsystem.rdw.ingest.processor.repository;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * This repository is responsible for persisting Subject
+ * display message key/value pairs.
```
malaffoon: nit: i think we can let the line length go up to 100?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/TargetRepository.java
```diff
@@ -1,15 +1,52 @@
 package org.opentestsystem.rdw.ingest.processor.repository;
 
+import org.opentestsystem.rdw.ingest.processor.model.WarehouseTarget;
+
+import java.util.Collection;
+
 /**
  * Repository for dealing with the target translation table
```
malaffoon: Is this comment correct, it is the translation table?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
```diff
@@ -3,31 +3,88 @@
...
 
     @Override
-    @Cacheable(value = "claim")
-    public Integer findIdByCode(final String code) {
+    public Integer findIdByCode(final int subjectId, final String code) {
         try {
-            return jdbcTemplate.queryForObject(sqlFindIdByCode, new MapSqlParameterSource("code", code), Integer.class);
+            return jdbcTemplate.queryForObject(sqlFindIdByCode,
+                    new MapSqlParameterSource("code", code)
+                            .addValue("subject_id", subjectId),
```
malaffoon: nit: i would consider adding helpers like the ImmutableMap.of where they have them for of(K,V), of(K,V,K,V), etc. (for this class that's all you'd need and i'd hesitate going much beyond 3 or 4 pairs).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
```diff
@@ -13,4 +17,44 @@
...
+    void create(String subjectCode, Collection<WarehouseDepthOfKnowledge> depths);
+
+    /**
+     * Given a collection of depth of knowledge ids, return the first that
+     * is referenced by an existing assessment item, or null if none are referenced.
+     *
+     * @param depthIds A collection of depth ids
+     * @return A referenced depth id or null if none are referenced
+     */
+    Integer hasItems(Collection<Integer> depthIds);
```
malaffoon: Curious why this return value? I suppose i could dig through the code but it seems ... odd.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
esotrope: i actually would vote for using spring's `SqlParamerterSource` interface everywhere we are using springs `JdbcTemplate` since they both come from the same package. also ImmutableMap.of doesnt allow null values which is annoying for our search object mapping

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectClaimService.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final ClaimRepository repository;
+
+    @Autowired
+    public DefaultSubjectClaimService(final ClaimRepository repository) {
+        this.repository = repository;
+    }
+
+    @Override
+    public void upsert(final Subject subject) {
```
malaffoon: It seems obvious but just checking: we're not worrying about Transactional, on purpose, correct?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectDepthOfKnowledgeService.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        repository.create(subjectCode, newDepths);
+        repository.update(updatedDepths);
+
+        final Map<Integer, WarehouseDepthOfKnowledge> deletedDepthsById = existingDepths.values().stream()
+                .collect(Collectors.toMap(
+                        WarehouseDepthOfKnowledge::getId,
+                        dok -> dok
+                ));
+        final Integer referencedId = repository.hasItems(deletedDepthsById.keySet());
```
malaffoon: Seems like you'd want to check this earlier: if there will be problems resolving anything we should know that before trying to do anything.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectItemDifficultyService.java
```diff
@@ -0,0 +1,85 @@
...
+        final Map<Integer, WarehouseItemDifficulty> difficultiesById = difficulties.stream()
+                .collect(Collectors.toMap(
+                        WarehouseItemDifficulty::getId,
+                        difficulty -> difficulty
+                ));
+        final Integer referencedId = repository.hasItems(difficultiesById.keySet());
+        if (referencedId == null) return;
+
+        final WarehouseItemDifficulty referencedDifficulty = difficultiesById.get(referencedId);
+        throw new IllegalArgumentException("Unable to " + action + " existing item difficulty cutpoint id: " + referencedId +
```
malaffoon: i guess this answers an earlier question i have but ... i question its usefulness: i think it should either report *all* codes that are still being used (meh), or just complain that at least one of the difficulties are in use.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
malaffoon: I think you misinterpreted my comment, i was suggesting something like this:
```
    public static SqlParameterSource params(String key, Object value) {
        return new MapSqlParameterSource(key, value);
    }

    public static SqlParameterSource params(String k1, Object v1, String k2, Object v2) {
        return new MapSqlParmeterSource(k1, v1).addValue(k2, v2);
    }
    ...
```

exam-processor/src/main/resources/application.sql.yml
wtritch: Reverted this change and left subject_claim_score bound to asmt_type_id

package-processor/build.gradle
wtritch: added

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseItemDifficulty.java
wtritch: They *should not* be null, but 0 is a valid value.  I'd rather validate against presence than hope that a 0 wasn't just a default de-serialized value...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseSubjectAssessmentType.java
wtritch: done

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
wtritch: Claims can be scorable, organizational, or both.  Exams can contain a score for each scorable claim. Organizational claims hold targets in the claim/target hierarchy.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/ClaimRepository.java
wtritch: Changed to `isReferenced`

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
wtritch: I hope so, but I'd rather not re-factor our package import as part of this jira =)

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
wtritch: doh, fixed

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/StandardRepository.java
wtritch: Changed to isReferenced

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/SubjectTranslationRepository.java
wtritch: Fine.  But only because you asked with a minimum of snark.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/TargetRepository.java
wtritch: changed to Repository for managing target persistence

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcClaimRepository.java
wtritch: I'd rather not to re-write all of the param source lines from:
```
.addValue("key", value),
.addValue("key", value),
```
to
```
"key", value,
"key", value,
```
although it would be slightly less boiler-plate...

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/DepthOfKnowledgeRepository.java
wtritch: It is odd, I believe I called this out before.  It's because I want to be able to tell the user *which* of of the potentially thousand targets they're trying to remove is currently referenced.  I just kept with the same signature for smaller-sized collections like depth-of-knowledge.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectClaimService.java
wtritch: correct, the transactional is higher up than this call, because I don't want a half-inserted or half-updated object.  I wrap the entire subject import message processing action in a transaction

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectItemDifficultyService.java
wtritch: For difficulties, sure.  For 1000 targets... my gut says that's a pretty horrific user experience.

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/UpsertAsmtsIT.java
```diff
@@ -63,6 +64,7 @@ public void itShouldInsert() {
         assertThat(jobExecution.getExitStatus()).isEqualTo(ExitStatus.COMPLETED);
         verifyTableCountsAfterTest(tableTestCounts);
 
+        final Map<String, Object> asmt = reportingJdbcTemplate.queryForMap("SELECT * from asmt where id = -99");
```
agorina: Is this being used?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseScorableClaim.java
```diff
@@ -0,0 +1,79 @@
...
+            claim.code = code;
+            claim.displayOrder = displayOrder;
+            claim.asmtTypeCode = asmtTypeCode;
+            return claim;
+        }
+
+        public Builder copy(final WarehouseScorableClaim claim) {
+            this.id = claim.id;
+            this.code = claim.code;
+            this.displayOrder = claim.displayOrder;
```
agorina: Seems to be missing `asmtTypeCode`.

wtritch: shoot, nice catch

migrate-reporting/src/test/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/UpsertAsmtsIT.java
wtritch: doh, no.  That was some debugging while the test was failing, removed.

PR #509, opened 2018-06-02
Migrate Target related data to reporting data mart.

agorina:


migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -454,7 +454,7 @@ sql:
                   JOIN claim wc ON wc.id = wt.claim_id
                   JOIN subject ws ON ws.id = wc.subject_id
                 WHERE wa.type_id = 3 AND ( ws.code <> 'Math' OR wc.code='1' ) AND
-                      wa.updated > ? AND wa.updated <= ? AND wa.updated <> wa.created
+                      wa.updated > ? AND wa.updated <= ? AND wa.updated <> wa.created AND wa.deleted = 0
```
agorina: Unrelated, and it really does not change anything for the OLAP migrate. But it is cleaner this way.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/StagingToReportingStepsConfig.java
```diff
@@ -68,6 +69,7 @@ public Step deleteEntitiesStep() {
                 "student_group_membership",
                 "exam",
                 "exam_item",
+                "exam_target_score",
```
agorina: These are the two tables added to the migrate process.

migrate-reporting/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -474,6 +474,41 @@ sql:
...
+            INSERT INTO exam_target_score (id, exam_id, target_id, student_relative_residual_score, standard_met_relative_residual_score)
+              SELECT
+                sets.id,
+                sets.exam_id,
+                sets.target_id,
+                sets.student_relative_residual_score,
+                sets.standard_met_relative_residual_score
+              FROM staging_exam_target_score sets
+                LEFT JOIN exam_target_score rets ON rets.id = sets.id
+              WHERE rets.id IS NULL;
```
malaffoon: This is migrating all target scores, shouldn't it be filtering to just those for summative and (ELA or Math-1)?

agorina: Since `warehouse.to.staging.sql.yml` pre-filtered only those you mentioned, this is not needed here.

PR #511, opened 2018-06-04
Update to Camden.SR7

malaffoon:
@wtritch any reason we shouldn't do this? I was going to do it in reporting as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,8 +89,8 @@ subprojects {
             dependency 'org.mockito:mockito-core:2.16.0'
         }
         imports {
-            // spring cloud Camden.SR5 includes spring cloud stream Brooklyn.SR2
-            mavenBom 'org.springframework.cloud:spring-cloud-dependencies:Camden.SR5'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -76,9 +76,9 @@ subprojects {
             dependency "org.opentestsystem.rdw.common:rdw-common-utils:${rdwCommonVersion}"
 
             // commonly used libraries
-            dependency 'com.google.code.gson:gson:2.8.0'
-            dependency 'com.google.guava:guava:21.0'
-            dependency 'commons-io:commons-io:2.4'
+            dependency 'com.google.code.gson:gson:2.8.5'
+            dependency 'com.google.guava:guava:23.0'
+            dependency 'commons-io:commons-io:2.6'
```
malaffoon: Bumped a couple of others.

wtritch:
I like keeping up to date.  We should hopefully find out quickly if there's a non-backwards-compatible change =)

PR #512, opened 2018-06-04
mttests

malaffoon:
After fiddling with the failing MT tests, we came up with this change. Because of the nature of the change (we just changed the test) it is not urgent to merge this.

I also tweaked the driver app to clean up output and stuff and things. Didn't really need it but i needed to do something whilst i was pontificating.

### DIFF
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
@@ -11,9 +11,7 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Import;
 
-import java.util.List;
-
-import static com.google.common.collect.Lists.newArrayList;
+import java.util.Map;
 
 /**
  * A console application that runs all available {@link MultiThreadedTest}s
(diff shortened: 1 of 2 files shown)


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedApplicationTest.java
```diff
@@ -34,34 +32,46 @@
...
             System.exit(1);
         }
 
-        for (int i = 0; i < 20; i++) {
-            for (final String test : beans) {
-                logger.info("Running test for [" + test + "] with thread count =" + TestThreadCount);
-                final boolean passed = ctx.getBean(test, MultiThreadedTest.class).run(TestThreadCount);
-                logger.info(test + " " + (passed ? "passed" : "failed"));
+        logger.info("Running {} iterations of {} tests, each with thread count {}", iterations, beans.size(), threadCount);
+        int failed = 0;
```
malaffoon: instead of collecting all the results, just keep track of failure count

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedStudentGroupAddStudent.java
```diff
@@ -45,10 +42,16 @@ protected void before() {
...
+
     }
 
     @Override
     protected void work() {
+        // add student to new groups
+        final List<Integer> groupIds = newArrayList(
+                repository.upsert(newGroup("test" + Math.random(), schoolId), 1).getId(),
+                repository.upsert(newGroup("test2" + Math.random(), schoolId), 1).getId()
+        );
```
malaffoon: Changed this test to be a bit more realistic: it adds the same student to different groups in each thread.

PR #513, opened 2018-06-06
Refactor our retry configuration to avoid double-instantiating retry …

wtritch:
…AOP advice.


agorina:
I thought we resolved it before? 
Well, I trust you.

malaffoon:
@agorina it might've been the upgrade to Spring that changed the behavior. Basically, we were getting lucky before since it was instantiating multiple RetryConfigurations. This forces the one true RetryConfiguration to be instantiated and ordered *before* instantiating any of the repositories that use it.

PR #514, opened 2018-06-08
Fix to state embargo migrate in OLAP.

agorina:


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -271,7 +271,7 @@ public Step updateEmbargoStep() {
         sqlsBuilder.addNext("state_embargo", "insert");
 
         return stepBuilderFactory.get(updateEmbargoStepName)
-                .tasklet(new SqlListExecutionStep(
+                .tasklet(new SqlListEmbargoExecutionStep(
```
agorina: This it the fix. This step deletes and re-inserts embargo but it was not doing it conditionally based on the embargo data being loaded into staging table. So a subsequent migrate deletes an entry.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/DeleteEntitiesStepRST.java
```diff
@@ -68,8 +66,6 @@ public void itShouldDeleteButNotInsertEntities() {
         tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), "student_ethnicity", -1, "ethnicity_id IN (-99)"));
         tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), "student_ethnicity", 0, "ethnicity_id IN (-98)"));
 
-        getStepExecutionContext().put("batch", mock(Migrate.class));
```
agorina: I have deleted it because I noticed that it was using the wrong name when putting a mock into context, and then I realized that it is not needed at all.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/UpdateEmbargoRST.java
```diff
@@ -78,4 +87,32 @@ public void itShouldUpdateStateEmbargo() {
...
+            @Sql(executionPhase = BEFORE_TEST_METHOD, statements = {
+                    "INSERT INTO staging_state_embargo (aggregate, migrate_id) VALUES (0, -10);",
+            }, scripts = {"classpath:OlapAndStagingEmbargoSetup.sql"}, config = @SqlConfig(dataSource = "olapDatasource")),
+
+            @Sql(executionPhase = AFTER_TEST_METHOD, statements = {
+                    "DELETE FROM staging_state_embargo;",
+            }, scripts = {"classpath:OlapAndStagingEmbargoTeardown.sql"}, config = @SqlConfig(dataSource = "olapDatasource"))
+    })
+    @Test
+    public void itShouldDoNothingIfEmbargoFlagIsNotSet() {
```
agorina: This test proves the fix.

PR #516, opened 2018-06-08
async job runner

malaffoon:
I'm not 100% sure that this is the best practice approach to this problem. I did play with another approach where i made the TaskExecutor for the JobLauncher a SimpleAsyncTaskExecutor. My concern with that approach is that the JobLauncher will return right away and any scheduling will not take into account when the job actually completes. Since the sole purpose of the migrate apps is to run this one job, it seemed wrong somehow to use an async executor way down in the innards. 

Anyway, this seems to work and is simpler.

### DIFF
--- a/migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateLifecycle.java
+++ b/migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateLifecycle.java
@@ -3,9 +3,14 @@ package org.opentestsystem.rdw.migrate.common;
 import org.springframework.context.Lifecycle;
 
 /**
- * An interface for a component that may be disabled and/or paused.
+ * An interface for a runnable migrate component that may be disabled and/or paused.
  */
 public interface MigrateLifecycle extends Lifecycle {
+
+    /**
+     * take action
+     */
+    void run();
 
     /**
      * @param flag true to enable component, false to disable
--- a/migrate-olap/build.gradle
+++ b/migrate-olap/build.gradle
@@ -66,6 +66,28 @@ cleanAllTest migrateAllTest
  gradle rst)
 */
 // If you are using different instances the full set of variables can be found in application-redshift.yml.
+//
+// If you want to run the application from within your IDE, to debug or whatever, you'll need to set all those
+// environment variables (and a couple more) OR you can pass them in as program arguments. I guess you could also
+// just modify the embedded application.yml. But here are the program arguments, without the more sensitive secrets:
+/*
+--archive.root=s3://rdw-ci-archive
+--archive.cloud.aws.credentials.accessKey=access-key
+--archive.cloud.aws.credentials.secretKey=secret-key
+--migrate.aws.redshift.role=arn:aws:iam::269146879732:role/rdw-redshift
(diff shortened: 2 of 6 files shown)


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateLifecycle.java
```diff
@@ -3,10 +3,15 @@
...
 /**
- * An interface for a component that may be disabled and/or paused.
+ * An interface for a runnable migrate component that may be disabled and/or paused.
  */
 public interface MigrateLifecycle extends Lifecycle {
 
+    /**
+     * take action
+     */
+    void run();
```
malaffoon: To be able to use AoP (i.e. the @Async and @Scheduled annotations) the methods being invoked all have to be in an interface so i pulled this up from the concrete implementation.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -35,9 +39,10 @@ public MigrateReportingJobRunner(final MigrateRepository migrateRepository,
         super(migrateRepository, importRepository, jobLauncher, job, batchSize);
     }
 
+    @Async
```
malaffoon: Strictly speaking this isn't necessary since the scheduling is delayed. But it made me think that, while it is executing a migrate, is it responding to health checks, etc. So i added it.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/RunOnStartup.java
```diff
@@ -0,0 +1,26 @@
...
+ * {@literal @}PostConstruct annotation. Instead, this bean forces it to run on startup.
+ *
+ * @see MigrateOlapReportingJobRunner
+ */
+@Component
+public class RunOnStartup {
+    private final MigrateLifecycle jobRunner;
+
+    public RunOnStartup(final MigrateLifecycle jobRunner) {
+        this.jobRunner = jobRunner;
```
malaffoon: This is what fails if the thing passed in isn't an interface. A very cryptic error about proxying CGLIB and stuff and things.

migrate-olap/build.gradle
```diff
@@ -66,6 +66,28 @@ cleanAllTest migrateAllTest
...
+--spring.migrate_datasource.password=password
+--spring.olap_datasource.url-server=rdw-qa.cibkulpjrgtr.us-west-2.redshift.amazonaws.com:5439
+--spring.olap_datasource.url-db=ci
+--spring.olap_datasource.username=bob
+--spring.olap_datasource.password=password
+--spring.warehouse_datasource.url-server=rdw-aurora-ci.cugsexobhx8t.us-west-2.rds.amazonaws.com:3306
+--spring.warehouse_datasource.url-schema=bob_warehouse_test
+--spring.warehouse_datasource.username=sbac
+--spring.warehouse_datasource.password=password
+ */
```
malaffoon: just preserving notes so i can copy and paste next time

PR #517, opened 2018-06-12
set timeouts for connect, socket and query

malaffoon:
Thoughts when doing this:
* connect and validation timeouts should be short, i chose 10s. Could be lower, what do you think?
* query timeout depends on app, either 30s or 120s
* socket timeout needs to be longer than query timeout (or else query won't be aborted), i added 10s to query timeout

### DIFF
--- a/exam-processor/src/main/resources/application.yml
+++ b/exam-processor/src/main/resources/application.yml
@@ -29,16 +29,20 @@ spring:
       &characterEncoding=${spring.datasource.character-encoding:utf8}\
       &rewriteBatchedStatements=${spring.datasource.rewrite-batched-statements:true}\
       &noAccessToProcedureBodies=${spring.datasource.no-access-to-procedure-bodies:true}\
+      &connectTimeout=${spring.datasource.connect-timeout:10000}\
+      &socketTimeout=${spring.datasource.socket-timeout:40000}\
       "
     username: root
     password:
     testWhileIdle: true
     validationQuery: SELECT 1
+    validationQueryTimeout: 10000
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
     maxActive: 12
     initialSize: 6
     maxWait: 10000
+    jdbcInterceptors: "QueryTimeoutInterceptor(queryTimeout=30)"
 
   jackson:
     default-property-inclusion: non_null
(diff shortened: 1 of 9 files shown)


exam-processor/src/main/resources/application.yml
```diff
@@ -29,16 +29,20 @@ spring:
...
     password:
     testWhileIdle: true
     validationQuery: SELECT 1
+    validationQueryTimeout: 10000
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
     maxActive: 12
     initialSize: 6
     maxWait: 10000
+    jdbcInterceptors: "QueryTimeoutInterceptor(queryTimeout=30)"
```
malaffoon: This is just how you do query timeouts with the apache tomcat connection pool.

group-processor/src/main/resources/application.yml
```diff
@@ -42,15 +42,19 @@ spring:
...
     username: root
     password:
     testWhileIdle: true
     validationQuery: SELECT 1
+    validationQueryTimeout: 10000
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
     maxActive: 10
     initialSize: 4
+    jdbcInterceptors: "QueryTimeoutInterceptor(queryTimeout=120)"
```
malaffoon: Made this one 120s for really large group files.

exam-processor/src/main/resources/application.yml
agorina: Nice!

migrate-olap/src/main/resources/application.yml
```diff
@@ -64,11 +64,14 @@ spring:
       ?useSSL=${spring.migrate_datasource.use-ssl:false}\
       &useLegacyDatetimeCode=${spring.migrate_datasource.use-legacy-datetime-code:false}\
       &characterEncoding=${spring.migrate_datasource.character-encoding:utf8}\
+      &connectTimeout=${spring.migrate_datasource.connect-timeout:10000}\
+      &socketTimeout=${spring.migrate_datasource.socket-timeout:0}\
```
malaffoon: Set migrate-olap to be unlimited, but have placeholders so it is easier to override if desired.

PR #519, opened 2018-06-15
Modified SQL for group upload.

agorina:


build.gradle
```diff
@@ -24,7 +24,7 @@ apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.2.0-91"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.2.0-366"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.2.0-369"
```
agorina: I accidentally merged the index change into develop.

group-processor/src/main/resources/application.sql.yml
```diff
@@ -91,32 +91,42 @@ sql:
...
             SET sg.subject_id = sub.subject_id,
-            sg.update_import_id = :import_id
+              sg.update_import_id = :import_id
+
+          update-modified-group-by-student: >-
+            UPDATE student_group sg
+              JOIN
+              (SELECT loading.group_id FROM
+                (SELECT upload.group_id AS group_id, GROUP_CONCAT(DISTINCT sgm.student_id ORDER BY sgm.student_id) AS students
+                  FROM upload_student_group upload
```
malaffoon: Is this right, shouldn't this be `FROM student_group esg ...`?

group-processor/src/main/resources/application.sql.yml
```diff
@@ -91,32 +91,42 @@ sql:
...
+                    LEFT JOIN student_group_membership sgm ON sgm.student_group_id = upload.group_id
+                  WHERE upload.import_id = :import_id
+                  GROUP BY upload.group_id) existing JOIN
+                (SELECT group_id, GROUP_CONCAT(DISTINCT student_id ORDER BY student_id) AS students
+                  FROM upload_student_group
+                  WHERE group_id IS NOT NULL AND import_id = :import_id
+                  GROUP BY group_id) loading ON loading.group_id = existing.group_id AND (NOT loading.students <=> existing.students)
+              ) sub ON sub.group_id = sg.id
+            SET sg.update_import_id = :import_id
+            WHERE sg.update_import_id <> :import_id
```
malaffoon: This might not be necessary with the above fix. The checks against `group_id IS NOT NULL` is what should be stopping the double-updating of a single group.

group-processor/src/main/resources/application.sql.yml
agorina: My understanding is that this find all students for the groups that are in the `upload_student_group`. And the reason there was `student_group` before is to find the subject change. Does it sound right?

malaffoon: Yup, yup i retract my comment.

PR #520, opened 2018-06-15
resubmit-groups

malaffoon:
Found this bug while testing the large student group file upload: the default behavior when resubmitting an import is to read the payload from the archive (S3) and then put it in the message body. But groups don't do that, they just put the digest (archive location) in the message.


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/AuthorityService.java
```diff
@@ -12,7 +12,7 @@
  * This service knows all about roles, permissions and spring security authorities.
  * Some mixing of concerns here, and the name is a bit misleading - deal with it.
  */
-interface AuthorityService {
+public interface AuthorityService {
```
malaffoon: Had to make a few things more visible for group import service to use.

PR #521, opened 2018-06-15
Fix a bug discovered by the load test.

agorina:


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -833,6 +833,7 @@ sql:
                 asmt_id                     = se.asmt_id,
                 school_year_asmt_grade_code = CONCAT(se.school_year::VARCHAR(4), CONCAT(',', se.asmt_grade_code::VARCHAR(2))),
                 grade_id                    = se.grade_id,
+                school_year                 = se.school_year,
```
agorina: I know, ideally I should have added a test for this, but..

I am going to reload `awsdev` with a small migrate batch size to run the functional test.

agorina: Just to explain, @malaffoon helped me walk it through (even though I know I wrote it :) 

We are not migrating by the exam id here (and this is the important key to remember). So a student took an exam in 2016, then we found an exam in 2018. We updated that student with a new exam, but the school year was outdated. And this caused all kinds of problem when Will ran reports.

PR #522, opened 2018-06-16
More changes to support configurable data elements for Ingest.

agorina:
I suggest to review each commit separately, should be easier this way.

I hope that this is enough to fulfill the requirement. @jtreuting please let me know what you think.


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ParserHelper.java
```diff
@@ -24,6 +24,10 @@ public ParserHelper(@NotNull final DataElementErrorCollector elementErrorCollect
         this.elementErrorCollector = elementErrorCollector;
     }
 
+    protected DataElementErrorCollector getElementErrorCollector(){
```
agorina: Need it in the parent class.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ConfigurableDataElement.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Represents configurable data elements processed during an exam (aka TRT) ingest that
+ * could be configured as required via {@link DataElementsConfiguration}.
+ * <p>
+ * If an element is not listed here, it is validated based on the TRT rules.
+ * <p>
+ * If a data element is not required it could be missing (or empty), but if present must have a valid value.
+ */
+public enum ConfigurableDataElement {
```
agorina: This is the main change. I do not know why github thinks it is a new file, I renamed it and changed to include _only_ data elements that could be configured.

Since we do not process all elements from TRT and in some cases (such as an item), we configure the entire element, it did not look right to include all other elements.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -112,28 +107,16 @@ public Student parseStudent(final Examinee examinee, final int schoolYear, final
                 .firstName(parserHelper.validate(FirstName, getBestAttributeValue(examinee, FirstName), 60))
                 .lastOrSurname(parserHelper.validate(LastOrSurname, getBestAttributeValue(examinee, LastOrSurname), 60))
                 .middleName(parserHelper.validate(MiddleName, getBestAttributeValue(examinee, MiddleName), 60))
-                .ssid(parserHelper.validate(StudentIdentifier, getBestAttributeValue(examinee, StudentIdentifier), 65))
+                .ssid(parserHelper.validate("StudentIdentifier", getBestAttributeValue(examinee, "StudentIdentifier"), 65, true))
```
agorina: This is not 'configurable' and always required.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -112,28 +107,16 @@ public Student parseStudent(final Examinee examinee, final int schoolYear, final
...
-            errorCollector.add(new DataElementError(StudentGroupName.name(), "none", "must provide a student group"));
-        }
-
-        studentBuilder
-                .ethnicityIds(ethnicity)
-                .groups(groups);
+                .lepExitAt(parserHelper.validate(LEPExitDate, getBestAttributeValue(examinee, LEPExitDate), toLocalDate))
+                // each individual ethnicity value is validated by the ethnicity service,
+                // but a validation of at least one ethnicity being required is done here
+                .ethnicityIds(parserHelper.isNotEmpty(Ethnicity, ethnicityService.getEthnicity(examinee, errorCollector)))
```
agorina: Added a helper since I needed in on other places.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -112,28 +107,16 @@ public Student parseStudent(final Examinee examinee, final int schoolYear, final
...
+                .groups(parseGroups(examinee, schoolYear, schoolId, parserHelper));
```
agorina: Groups are always optional. Could not be configured as required.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -152,7 +135,7 @@ public StudentExamAttributes parseStudentExamAttributes(final Examinee examinee,
                 .engProfLvl(parserHelper.validate(EnglishLanguageProficiencyLevel, getBestAttributeValue(examinee, EnglishLanguageProficiencyLevel), 20))
                 .t3ProgramType(parserHelper.validate(TitleIIILanguageInstructionProgramType, getBestAttributeValue(examinee, TitleIIILanguageInstructionProgramType), 30))
                 .primDisabilityType(parserHelper.validate(PrimaryDisabilityType, getBestAttributeValue(examinee, PrimaryDisabilityType), 3))
-                .gradeId(parserHelper.validate(GradeLevelWhenAssessed, getBestAttributeValue(examinee, GradeLevelWhenAssessed), checkNotBlank.andThen(gradeRepository::findIdByCode)))
+                .gradeId(parserHelper.validate("GradeLevelWhenAssessed", getBestAttributeValue(examinee, "GradeLevelWhenAssessed"), checkNotBlank.andThen(gradeRepository::findIdByCode)))
```
agorina: Always required.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -68,20 +73,22 @@
...
         this.claimTargetRepository = claimTargetRepository;
         this.examineeProcessor = examineeProcessor;
+        this.optionalDataElements = requiredDataElementsConfiguration.getOptionalDataElements();
     }
 
     @Override
     public Exam parseExam(final TDSReport report, final int schoolId, final Assessment assessment) throws ImportException {
         final DataElementErrorCollector errorCollector = new DataElementErrorCollector();
-        final ParserHelper parserHelper = new ParserHelper(errorCollector);
+        final ExamParserHelper parserHelper = new ExamParserHelper(errorCollector, optionalDataElements);
```
agorina: I will add more changes that will be using it.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -152,10 +159,10 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
      * Helper to parse score data. Because the score data consists of multiple elements
      * this uses builders to assemble the separate parts.
      *
-     * @param opportunity opportunity containing score elements
-     * @param claimCodeToId known claim codes for this assessment
+     * @param opportunity    opportunity containing score elements
```
agorina: Just reformatted.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultTDSReportProcessorIT.java
```diff
@@ -98,6 +99,75 @@ public void itShouldProcessMessageWithXmlContent() {
         assertThat(countRowsInTable(jdbcTemplate, "student_group_membership")).isEqualTo(beforeGroupMembershipCount + 2);
     }
 
+    @Test
+    public void itShouldReturnErrorsForDefaultRequiredStudentExamDataElement() {
+        int beforeIabExamCount = countRowsInTableWhere(jdbcTemplate, "exam", "type_id = 2");
+
+        for (final String sample : new String[]{
+                "TDSReport.iab.noElement.xml"
```
agorina: I added test to make sure we can process and and return the right errors.

exam-processor/src/test/resources/TDSReport.iab.emptyOpportunityAttributes.xml
```diff
@@ -0,0 +1,74 @@
...
+    <ExamineeRelationship name="StateName" value="ARMED FORCES PACIFIC" context="INITIAL" contextDate="2016-10-12T21:51:46.003"/>
+    <ExamineeRelationship name="StateAbbreviation" value="AP" context="INITIAL" contextDate="2016-10-12T21:51:46.003"/>
+    <ExamineeRelationship name="ResponsibleDistrictIdentifier" value="UCLALS" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+    <ExamineeRelationship name="OrganizationName" value="UCLA" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+    <ExamineeRelationship name="ResponsibleInstitutionIdentifier" value="APTEST1" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+    <ExamineeRelationship name="NameOfInstitution" value="APTESTSCHOOL" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+    <ExamineeRelationship name="StateName" value="ARMED FORCES PACIFIC" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+    <ExamineeRelationship name="StateAbbreviation" value="AP" context="FINAL" contextDate="2016-10-12T22:00:49.069"/>
+  </Examinee>
+  <Opportunity dateCompleted="">
```
agorina: `dateCompleted` is the only attribute that I could not remove (still gives an XSD validation error). But since it could be empty I think it is good enough.

exam-processor/src/test/resources/application-optional.yml
```diff
@@ -4,4 +4,4 @@ spring:
 
 validation:
   requiredDataElements:
-    - SchoolId
+    -
```
agorina: This seems to be a way to load an empty list.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ConfigurableDataElement.java
```diff
@@ -37,7 +37,11 @@
...
     // this that is an 'umbrella' data element that represents all possible Race/Ethnicity elements further configured in the db
     Ethnicity("Ethnicity"),
-    // Below is an 'umbrella' data element that represents validation for an exam having at least one item.
+    // All exam's data is optional; the following could be configured as required
+    SessionId("SessionId"),
+    Completeness("Completeness"),
+    AdministrationCondition("AdministrationCondition"),
+    // Below is an 'umbrella' element that represents validation for an exam having at least one item.
     // An exam's `item` is an optional element, but if present all its required attributes must be present.
     ExamItems("ExamItems");
```
agorina: I hope that this is enough to complete this requirement.  There are a lot of data elements we process, just in case. I do not see a reason in making them configurable since they are not really used by the system.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -99,16 +103,16 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
...
                 .studentExamAttributes(examineeProcessor.parseStudentExamAttributes(report.getExaminee(), schoolId, errorCollector))
                 .schoolYear(parserHelper.validate("Test AcademicYear", test.getAcademicYear(), toYear))
                 .assessmentId(assessment.getId())
                 .asmtVersion(parserHelper.validate("Test AssessmentVersion", test.getAssessmentVersion(), 30, false))
                 .opportunity(opportunity.getOpportunity())
-                .oppId(parserHelper.validate("Opportunity oppId", opportunity.getOppId(), 60, false))
-                .sessionId(parserHelper.validate("sessionId", opportunity.getSessionId(), 128, true))
-                .completenessId(parserHelper.validate("completeness", opportunity.getBestCompleteness(), checkNotBlank.andThen(completenessRepository::findIdByCode).andThen(new CheckNotNull<>())))
-                .administrationConditionId(parserHelper.validate("AdministrationCondition", opportunity.getAdministrationCondition(), checkNotBlank.andThen(adminConditionRepository::findIdByCode).andThen(new CheckNotNull<>())))
+                .oppId(parserHelper.validate("OpportunityId", opportunity.getOppId(), 60, true))
```
agorina: I was surprised to see that we all `null` here. I thought that `OpportunityId` is required and not configurable. So I made this change.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultTDSReportProcessorIT.java
```diff
@@ -98,6 +99,76 @@ public void itShouldProcessMessageWithXmlContent() {
...
+        }
+        assertThat(countRowsInTable(jdbcTemplate, "exam")).isEqualTo(beforeIabExamCount);
+    }
+
+    @Test
+    public void itShouldReturnErrorsForDefaultRequiredExamDataElement() {
+        int beforeItemCount = countRowsInTable(jdbcTemplate, "exam_item");
+        int beforeIabExamCount = countRowsInTableWhere(jdbcTemplate, "exam", "type_id = 2");
+
+        for (final String sample : new String[]{
```
jtreuting: is there a reason for the for loop in these tests?  future planning for adding more sample TRTs?

agorina: 'Copy and paste' error. Thank you for noticing it.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ConfigurableDataElement.java
```diff
@@ -0,0 +1,57 @@
...
+    Ethnicity("Ethnicity"),
+    // All exam's data is optional; the following could be configured as required
+    SessionId("SessionId"),
+    Completeness("Completeness"),
+    AdministrationCondition("AdministrationCondition"),
+    // Below is an 'umbrella' element that represents validation for an exam having at least one item.
+    // An exam's `item` is an optional element, but if present all its required attributes must be present.
+    ExamItems("ExamItems");
+
+    private final String value;
```
malaffoon: I know we've used `value` as the name of the extra data associated with an enum but it can be confusing (e.g. ConfigurableDataElement.values() returns the enums, not the `value`s). Perhaps `element` or `name`? Another possibility is to get rid of the value completely since they are all set to exactly the same thing as the enum `name()`.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -92,16 +103,16 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
...
```
malaffoon: why is the name `OpportunityId` instead of `oppId`, seems inconsistent?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -188,10 +199,10 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
                         .computeIfAbsent(scoreOf, k -> ExamClaim.builder().claimId(claimCodeToId.get(k)));
                 if (ScaleScoreMeasureLabel.equals(label)) {
                     builder.scaleScore(parserHelper.validate(scoreOf + " " + label, score.getValue(), this::toScaleScore))
-                           .scaleScoreStdErr(parserHelper.validate(scoreOf + " " + label + " StandardError", score.getStandardError(), toPositiveDoubleOrNull));
+                            .scaleScoreStdErr(parserHelper.validate(scoreOf + " " + label + " StandardError", score.getStandardError(), toPositiveDoubleOrNull));
```
malaffoon: nit: seems more readable with the dots lined up

exam-processor/src/main/java/utils/ExamParserHelper.java
```diff
@@ -39,7 +42,17 @@ public ExamParserHelper(final DataElementErrorCollector elementErrorCollector,
...
      */
-    public String validate(final DataElement dataElement, final String rawValue, final int maxLength) {
+    public String validate(final ConfigurableDataElement dataElement, final String rawValue, final int maxLength) {
         return super.validate(dataElement.name(), rawValue, maxLength, !optionalDataElements.contains(dataElement));
     }
+
+    /**
+     * Validates a given collection to be non null/empty depending if an element is required or not
+     */
+    public <T> Collection<T> isNotEmpty(final ConfigurableDataElement dataElement, final Collection<T> collection) {
```
malaffoon: I think it would be more consistent to have a public validate method that takes a collection and then have a helper validate function that checks not empty, so you'd end up with something more like: `parserHelper.validate(element, collection, checkNotEmpty);`

exam-processor/src/test/resources/TDSReport.iab.emptyOpportunityAttributes.xml
malaffoon: Yes, that's why i left it required. They (SB) established that it was required but could be blank. Odd, but that's the way they did it.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ConfigurableDataElement.java
agorina: Removed the value.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
agorina: Because @esotrope had trained me that abbreviations are bad! :wink:

But in this case I think it makes sense to change it, will do.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
agorina: fixed.

exam-processor/src/main/java/utils/ExamParserHelper.java
agorina: It seems this way, and I have tried to change it...but ran into some complications. So leaving it as is. If you want we can chat about it tomorrow.

PR #526, opened 2018-06-19
Task/add performance level attributes

wtritch:
Add "shortName" and "suffix" attribute handling to the PerformanceLevel subject element.

This PR defaults a performance level's optional "shortName" text to the "name" text and defaults the optional "suffix" text to an empty string.

### DIFF
--- a/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
+++ b/package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
@@ -93,7 +93,10 @@ class DefaultSubjectTranslationService implements SubjectTranslationService {
         for (final PerformanceLevel performanceLevel : performanceLevels.getPerformanceLevels()) {
             context.push(String.valueOf(performanceLevel.getLevel()));
             context.setMessage("name", performanceLevel.getName());
+            context.setMessage("short-name", isBlank(performanceLevel.getShortName()) ? performanceLevel.getName() : performanceLevel.getShortName());
+            context.setMessage("suffix", performanceLevel.getSuffix() == null ? "" : performanceLevel.getSuffix());
             context.setMessage("color", performanceLevel.getColor());
+
             context.pop();
         }
         context.pop();
@@ -223,7 +226,7 @@ class DefaultSubjectTranslationService implements SubjectTranslationService {
         }
 
         void setMessage(final String lastPath, final String message) {
-            if (isBlank(message)) return;
+            if (null == message) return;
 
             final String messageKey = getMessageKey(lastPath);
             final String existingMessage = existingMessages.remove(messageKey);
--- a/package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationServiceTest.java
+++ b/package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationServiceTest.java
@@ -115,18 +115,30 @@ public class DefaultSubjectTranslationServiceTest {
         verify(repository).create(eq(SubjectCode), messageCaptor.capture());
         assertThat(messageCaptor.getValue()).contains(
                 entry("subject.mysub.asmt-type.ica.level.1.name", "Level 1"),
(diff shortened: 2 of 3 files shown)


package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationServiceTest.java
```diff
@@ -242,6 +254,8 @@ public void itShouldSetStandardMessages() {
             levels.add(PerformanceLevel.builder()
                     .level(i)
                     .name("Level " + i)
+                    .shortName(i == 1 ? null : "Short Level " + i)
+                    .suffix(i == 1 ? null : "Suffix")
```
jtreuting: not sure if it matters but this setup doesn't give you any where the short name is provided and the suffix isn't.

wtritch: Expect the unexpected, we have no idea how this is going to be used, even for ELPAC =)

wtritch: Hm... I suppose, but they're independent variables.

jtreuting: that's why i'm not sure if it matters, if you were planning on coverage of all combinations or not.  but real world situation could be to have the short name and no suffix.  So sometimes use the full version and sometimes use the short one, but no need for the additional suffix.

wtritch: Yeah, but the back-end and this test doesn't do anything with the values, and I haven't covered every permutation of every property on every class either =)  Just because the UI tends to use the properties together doesn't mean much here =)

wtritch: Sorry, I must be a little brain burnt.  To answer your simple first question: No, it doesn’t matter :-)

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
```diff
@@ -93,7 +93,10 @@ private void appendPerformanceLevels(final MessageContext context, final Perform
         for (final PerformanceLevel performanceLevel : performanceLevels.getPerformanceLevels()) {
             context.push(String.valueOf(performanceLevel.getLevel()));
             context.setMessage("name", performanceLevel.getName());
+            context.setMessage("short-name", isBlank(performanceLevel.getShortName()) ? performanceLevel.getName() : performanceLevel.getShortName());
+            context.setMessage("suffix", performanceLevel.getSuffix() == null ? "" : performanceLevel.getSuffix());
```
esotrope: should these all not be trimming? i would expect to see `isBlank(safeTrim(level.getSuffix())) ? "" : level.getSuffix().trim()` or some such

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
```diff
@@ -223,7 +226,7 @@ MessageContext pop() {
         }
 
         void setMessage(final String lastPath, final String message) {
-            if (isBlank(message)) return;
+            if (null == message) return;
```
esotrope: so we're Yoda programming now?

wtritch: Well, I don't want to store `null` message text, but I *do* want to store EMPTY message text =/

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
wtritch: hm... good question.  Yeah, I should probably be trimming the text, but I'll do it down in the setMessage method.

PR #535, opened 2018-06-27
A bug fix for the optional `completeness` and `administration condition`.

agorina:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -159,6 +158,14 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
         return accommodationIds;
     }
 
+    private Integer toValidAdministrationCondition(final String value) {
```
agorina: According to the TRT spec, if not present and is not configured as optional, it has a default value.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -159,6 +158,14 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
...
+        return (Integer) new CheckNotNull<>().apply(adminConditionRepository.findIdByCode((value == null || value.isEmpty()) ? "SD" : value));
+    }
+
+    private Integer toValidCompleteness(final String value) {
```
agorina: Ditto.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessorTest.java
```diff
@@ -405,8 +405,8 @@ public void itShouldReturnErrorsForConfiguredRequiredElements() {
                     "{\"elementName\":\"ExamItems\",\"value\":\"none\",\"error\":\"must provide at least one value for ExamItems\"}," +
                     "{\"elementName\":\"OppId\",\"error\":\"value may not be blank\"}," +
                     "{\"elementName\":\"SessionId\",\"error\":\"value may not be blank\"}," +
-                    "{\"elementName\":\"Completeness\",\"value\":\"Complete\",\"error\":\"invalid value\"}" +
```
agorina: Yeah, I should have noticed that this error message is not right. 

But we have Nohemi who is better than me!

PR #536, opened 2018-06-28
Support optional data element migration to Redshift.

agorina:


agorina:
Merging, will address comments if needed.

PR #538, opened 2018-07-02
Remove unused data elements.

agorina:


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/step/StagingToReportingStepsConfig.java
```diff
@@ -99,12 +99,9 @@ public Step upsertSubjectsStep() {
                 .addNext("target", "deleteAsPartOfParentUpdate") //target depends on claim
                 .addNext("claim", "deleteAsPartOfParentUpdate")
                 .addNext("claim", "insert")
-                .addNext("claim", "update")
```
agorina: Since we are migrating immutable data only, there is no update.

PR #540, opened 2018-07-03
Ingest an optional assessment type long name if defined by a subject

wtritch:


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectTranslationService.java
```diff
@@ -75,6 +75,7 @@ private void appendAssessmentTypes(final MessageContext context, final List<Subj
         for (final SubjectAssessmentType assessmentType : assessmentTypes) {
             context.push("asmt-type").push(assessmentType.getCode());
             context.setMessage("name", assessmentType.getName());
+            context.setMessage("long-name", isBlank(assessmentType.getLongName()) ? assessmentType.getName() : assessmentType.getLongName());
```
wtritch: Default to the name if a longName isn't defined.

PR #546, opened 2018-09-09
Apply (optional) transformations on TRT files.

malaffoon:
SmarterBalanced has an issue with a test result vendor providing "bad" data, specifically the bankKey has an incorrect prefix: instead of "200" they are getting "10200". To help with this (and other bad data issues that have cropped up in the past), this change allows an XSLT to be applied to the TRT as it is being processed. These transformations can be configured by user.

Basic testing worked so i'm posting this. I still have to experiment with a live system using a configuration server to make sure the XSLT content is properly handled.

There is a change to RDW_Common that supports this: https://github.com/SmarterApp/RDW_Common/pull/87.

----
NOTE: it would be sufficient to simply apply a correction to the bankKey as it is being processed. This is done in exactly one place in the processing code so would be okay. I thought it was worth a couple extra hours to do this but feel free to chastise me if you feel differently.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -1,35 +1,57 @@
...
 public class ExamProcessor {
     private static final Logger logger = LoggerFactory.getLogger(ExamProcessor.class);
 
     private final TDSReportProcessor reportProcessor;
     private final ImportRepository importRepository;
+    private final ResourceLoader resourceLoader;
+    private final Transformations transformations;
+
+    // cache of transformations by user
+    private final Map<String, String> userTransformations = new HashMap<>();
```
malaffoon: I don't want to go to the configuration server every time a TRT comes in, so i cache by user. Yes, this means that any change to the transformation configuration requires bouncing the exam processor.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -40,25 +62,21 @@ public void process(final Message<?> message) {
 
         logger.debug("received {} import {}", accessor.getContent(), importId);
 
+        final String transformation = getTransformationForUser(accessor.getUserLogin());
```
malaffoon: This loads the XSLT file/resource into a String. Yup, the whole thing in a String.

agorina:
Could you please clarify why you made this configurable *per user*? Wouldn't it be enough per an installation?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -68,13 +86,26 @@ private void process(final TDSReport tdsReport, final long importId) {
...
-        } catch (final RuntimeException e) {
-            logger.info("failed to convert xml to tds report with error: " + e.getMessage());
-            importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
-        }
-        return null;
+    private String getTransformationForUser(final String username) {
+        if (username == null || username.isEmpty()) return null;
+
+        return userTransformations.computeIfAbsent(username, user -> {
+            final String transformation = transformations.get(user);
```
malaffoon: `transformation` here is the resource name, e.g. `"binary-http://localhost:8888/*/*/master/xslt/alice.xsl"`.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/config/Transformations.java
```diff
@@ -0,0 +1,26 @@
...
+ * Configuration for transformations
+ */
+@Component
+@ConfigurationProperties(prefix = "transformations")
+public class Transformations {
+    private final Map<String, String> exam = newHashMap();
+
+    public Map<String, String> getExam() {
+        return exam;
+    }
```
malaffoon: Yeah, the english doesn't roll off the tongue but it makes the configuration look okay:
```yaml
transformations:
  exam:
    dwtest@example.com: "binary-${spring.cloud.config.uri}/*/*/master/xslt/alice.xsl"
```

exam-processor/src/test/resources/application-test.yml
```diff
@@ -1,3 +1,8 @@
 spring:
   datasource:
     url-schema: warehouse_test
+
+transformations:
+  exam:
+    default: test
+    dwtest@example.com: "binary-${spring.cloud.config.uri}/*/*/master/xslt/dwtest.xsl"
```
malaffoon: i'll remove this, it isn't being used

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -0,0 +1,19 @@
+<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
malaffoon: This is the actual transformation file that i expect to deliver to SmarterBalanced.

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -0,0 +1,19 @@
...
+
+  <!-- identity rule copies everything by default -->
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- this rule removes leading 10 from item bank key -->
+  <xsl:template match="Item[starts-with(@bankKey, '10')]/@bankKey">
```
jtreuting: would this match a bankkey of 100 and change that to 0 or 1000 to 00?  Seems like we might need a full match right.

malaffoon: You are correct. Since the bankkey has been 200 or 180 for everything i've seen i didn't worry about it too much. There is `matches` capability in XSLT 2.0 so i can improve it.

exam-processor/src/test/resources/application-test.yml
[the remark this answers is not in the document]
malaffoon: Correct (i think). The system will look for the user, then the default. It is perfectly happy not finding a match. I think there are three common scenarios:
1. No transformations at all, none of these properties set. System behaves as today.
2. A system-wide transformation, so only `transformations.exam.default` will be set. It is applied to *all* TRTs.
3. A problematic vendor so only that user's value set, and the transformation is applied to only that user's TRTs.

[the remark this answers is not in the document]
agorina: Thank you.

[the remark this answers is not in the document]
malaffoon: Well shoot ... testing the full stack, the username isn't being included with the message. So i may have revert this and just have a single transformation for exams.  fyi, @agorina , @jtreuting

PR #547, opened 2018-09-12
Handle the new MC/MS/ESRB response format using an XSLT transformatio…

wtritch:
…n to bring it back to expected values.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,7 @@ subprojects {
             dependency 'org.apache.commons:commons-csv:1.5'
             dependency 'org.apache.commons:commons-lang3:3.7'
             dependency 'org.slf4j:slf4j-api:1.7.25'
+            dependency 'net.sf.saxon:Saxon-HE:9.7.0-21'
 
             // commonly used test libraries
             dependency 'junit:junit:4.12'
--- a/exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
+++ b/exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
@@ -1,5 +1,5 @@
 <TDSReport>
-    <Test grade="4" mode="online" contract="SBAC" bankKey="200" testId="SBAC-IAB-FIXED-G4M-G-MATH-4" subject="MATH"
+    <Test grade="4" mode="online" contract="SBAC" bankKey="10200" testId="SBAC-IAB-FIXED-G4M-G-MATH-4" subject="MATH"
           name="(SBAC)SBAC-IAB-FIXED-G4M-G-MATH-4-Winter-2016-2017" handScoreProject="0" assessmentType="interim" academicYear="2016" assessmentVersion="9786"/>
     <Examinee isDemo="0" key="4054">
         <ExamineeAttribute name="Birthdate" value="1900-01-01" context="INITIAL" contextDate="2016-10-12T21:51:46.003"/>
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -81,6 +81,7 @@ subprojects {
             dependency 'org.apache.commons:commons-csv:1.5'
             dependency 'org.apache.commons:commons-lang3:3.7'
             dependency 'org.slf4j:slf4j-api:1.7.25'
+            dependency 'net.sf.saxon:Saxon-HE:9.7.0-21'
```
wtritch: This registers the latest Saxon 9.7 version (9.8 HE supports some XSLT 3.0 but 9.7 was called out as more stable)

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -15,5 +15,12 @@
     </xsl:attribute>
   </xsl:template>
 
+  <!-- this rule converts IAT multiple-choice, multiple-select, and ESRB responses to the expected format ("A,C,D") -->
+  <xsl:template match="Response[contains(text(),'choiceInteraction_')]/text()">
+    <xsl:variable name="choices" select="replace(., '\s*&lt;(?!/?value)[^&gt;]+&gt;\s*', '', ';j')"/>
+    <xsl:variable name="choiceCharacters" select="replace($choices, '&lt;[^&lt;]+-choice-(\w)&lt;[^&gt;]+&gt;', '$1,')"/>
+    <xsl:value-of select="substring($choiceCharacters, 1, string-length($choiceCharacters)-1)"/>
+  </xsl:template>
```
wtritch: Note that this rule will not be applied unless added to the configuration-provided XSL in a runtime environment.

build.gradle
malaffoon: Should these changes be in common? That's where the actual transformation utility is. It would also allow you to test the changes?

wtritch: That was my argument.  I don't necessarily want to add Saxon to every project we have, and common's features are fully accessible with XSLT 1.0...  But I could be convinced if you felt strongly.

exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
```diff
@@ -107,43 +107,41 @@
...
+        <Item adminDate="2018-08-27T15:38:51.64-05:00" bankKey="10200" clientId="" dropped="0" format="IAT"
+              isSelected="1" key="37115" mimeType="text/xml" numberVisits="1" operational="1" pageNumber="4"
+              pageTime="22400" pageVisits="1" position="5" responseDuration="22400" score="0" scoreStatus="SCORED"
+              segmentId="(SBAC)SBAC-GEN-INTR-ICA-ELA-COMBINED-6-SEG1-Winter-2018-2019">
+            <Response date="2018-08-27T15:39:05.14-05:00" key="3d3b10b9-df2d-4410-8751-88080d832347" type="">&lt;itemResponse&gt;&lt;response
+                id="choiceInteraction_1.RESPONSE"&gt;&lt;value&gt;choiceInteraction_1-choice-B&lt;/value&gt;&lt;/response&gt;&lt;response
+                id="choiceInteraction_2.RESPONSE"&gt;&lt;value&gt;choiceInteraction_2-choice-A&lt;/value&gt;&lt;/response&gt;&lt;/itemResponse&gt;
+            </Response>
+            <ScoreInfo maxScore="1" scoreDimension="overall" scorePoint="0" scoreStatus="Scored">
+                <ScoreRationale/>
```
malaffoon: I can do it if you want but i think that we should add a test in ExamProcessorIT that captures the result of this being converted into a TDSReport and validate the content of the item responses. I just think we need a test somewhere that actually executes the sample.

wtritch: Can do

build.gradle
malaffoon: Nope, fine with me.

exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
wtritch: done

PR #548, opened 2018-09-18
Task/new response format conversions

wtritch:
This adds support for migrating the new data formats for:
- EBSR
- MI
- TI
- HTQ
- EQ
- GI
- SA
- ER/WER

to the expected format required by IRIS.


build.gradle
```diff
@@ -87,6 +87,8 @@ subprojects {
             dependency 'junit:junit:4.12'
             dependency 'org.assertj:assertj-core:3.6.2'
             dependency 'org.mockito:mockito-core:2.16.0'
+            dependency 'org.xmlunit:xmlunit-core:2.6.2'
+            dependency 'org.xmlunit:xmlunit-matchers:2.6.2'
```
wtritch: I added in an xmlunit dependency for testing... thoughts?

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorTest.java
```diff
@@ -66,17 +77,93 @@ public void itShouldProcessMessageWithXmlContent() {
...
+                .map(Node::getTextContent)
+                .collect(Collectors.toList());
+        assertThat(regionObjectValues).containsExactly("Step1", "Step2", "Step3", "Step4");
+        Assert.assertThat(giXML, hasXPath("/AnswerSet/Question/QuestionPart/SnapPoint/text()", equalTo("70@91,361;193,361;299,361;407,361;299,345;299,377")));
+
+        //SA | ER | WER
+        assertThat(items.get(8).getResponse().getContent())
+                .contains("<p>This is <strong>free-entry</strong> text that <i>students</i> can:")
+                .contains("<ul><li>enter</li><li>style</li><li>organize</li></ul>")
+                .contains("<p>however they like. El Niño");
```
wtritch: I have some concerns about exactly how encoded the free-entry text is.  I believe the TRTs are UFT8-encoded, so I'm hoping characters like `ñ` aren't encoded as HTML entities like `&ntilde;`

exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
```diff
@@ -144,43 +144,91 @@
...
-              mimeType="text/plain" segmentId="(SBAC)SBAC-IAB-FIXED-G4M-G-MATH-4-Winter-2016-2017">
-            <Response date="2016-10-12T22:00:24.765" type="value"> &lt;itemResponse&gt;&lt;response id="RESPONSE"&gt;&lt;value&gt;1 a&lt;/value&gt;&lt;value&gt;2 b&lt;/value&gt;&lt;value&gt;3
-                b&lt;/value&gt;&lt;/response&gt;&lt;/itemResponse&gt; </Response>
-            <ScoreInfo scorePoint="0" maxScore="1" scoreDimension="overall" scoreStatus="Scored">
-                <ScoreRationale>
-                    <Message>
-                        successfully scored
-                    </Message>
-                </ScoreRationale>
+        <!-- Sample Table Interaction (TI) (from xls, no sample in TRTs) -->
```
wtritch: We didn't have TRT samples for some of these interactions, so I hand-jammed them from Brandt's Excel doc.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorTest.java
wtritch: Also note that I'm only converting the interaction types that were explicitly called out in Brandt's Excel doc, where he stripped out some interaction types that aren't currently used by SmarterBalanced.

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -1,6 +1,30 @@
...
+    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:trt="urn:trt">
   <xsl:output method="xml" omit-xml-declaration="yes" indent="yes"/>
 
+  <xsl:variable name="escapedChars">
+    <entry key="&amp;lt;">&lt;</entry>
+    <entry key="&amp;gt;">&gt;</entry>
+    <entry key="&amp;amp;">&amp;</entry>
+    <entry key="&amp;quot;">&quot;</entry>
+  </xsl:variable>
```
wtritch: These are the only characters that are being *unescaped* currently.  I'm hoping that with UTF-8 TRTs additional characters won't be HTML-encoded, but we could really use a sample TRT with non-ASCII characters.  Since all of the samples we have don't bother to encode `"` as `&quot;` I'm hoping AIR is using a minimal encoding.

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -15,12 +39,124 @@
...
+      <responseTable>
+        <tr><th id="col0" /><th id="col1" /><th id="col2" /><th id="col3" /><th id="col4" /><th id="col5" /><th id="col6" /><th id="col7" /><th id="col8" /><th id="col9" /><th id="col10" /></tr>
+        <tr><td /><td>B</td><td /><td /><td>A</td><td /><td /><td>C</td><td /><td /><td>D</td></tr>
+      </responseTable>
+    </responseSpec>
+  -->
+  <xsl:template match="Response[contains(text(),'tableInteraction_')]/text()">
+    <xsl:variable name="escapedResponse" select="replace(., '.+&lt;value&gt;(.+)&lt;/value&gt;.+', '$1', 's')"/>
+    <xsl:value-of select="trt:unescape($escapedResponse)"/>
+  </xsl:template>
```
wtritch: Note that TI, EQ, GI, and writing interactions all share the same behavior.  While I could group them all under a single
```
Response[
  contains(text(), 'tableInteraction_') or
  contains(text(), 'equationInteraction_') or
  ...
]/text()
```
I wonder if the reduction in clarity is worth the code re-use.  Thoughts?

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -15,12 +39,124 @@
...
+
+  <!--
+    This rule converts Short Answer / Writing Extended Response Interaction (SA|ER|WER) responses to the expected WYSIWYG format:
+    <p>This is <strong>free-entry</strong> text that <i>students</i> can:
+    <ul><li>enter</li><li>style</li><li>organize</li></ul>
+    <p>however they like.
+  -->
+  <xsl:template match="Response[contains(text(),'textEntryInteraction_')]/text()">
```
wtritch: With support for free-text entry, if a student for whatever reason enters the value `gridInteraction_` in their entry, we will have problems with our XLS template matchers.  Thoughts?

build.gradle
malaffoon: 👍

exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
```diff
@@ -107,7 +107,7 @@
         <Score measureOf="Overall" measureLabel="ScaleScore" value="2595.97876140521" standardError="50.8535397655728"/>
```
malaffoon: nit: might want to rename the resource to something like `TDSReport.iab.xslt-test.xml`

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -1,6 +1,30 @@
-<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
malaffoon: nit: same nit, might change the name of the resource to be more generic

exam-processor/src/test/resources/bankkey.xsl
malaffoon: I know this is going to be confusing in 6 months. Perhaps call these `"escapedXmlChars"`? Cause it is fine to have escaped characters that aren't part of the XML-skeleton, right? i.e. if `&#39;` is in the text, it can be left that way? Alternatively, more words or examples ... i guess we have the unit tests.

exam-processor/src/test/resources/bankkey.xsl
```diff
@@ -15,12 +39,124 @@
...
+    This rule converts EBSR multiple-choice and multiple-select responses to the expected format:
+    <itemResponse>
+      <response id="EBSR1">
+        <value>A</value>
+      </response>
+      <response id="EBSR2">
+        <value>C</value>
+      </response>
+    </itemResponse>
+  -->
```
malaffoon: 👍   nice comment!

exam-processor/src/test/resources/bankkey.xsl
malaffoon: I like it the way it is ... separate.

exam-processor/src/test/resources/bankkey.xsl
malaffoon: I don't think that is a problem we have to worry about right now.

Perhaps we can discuss this afternoon as part of the "final solution".

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorTest.java
```diff
@@ -66,17 +77,93 @@ public void itShouldProcessMessageWithXmlContent() {
...
+    public void itShouldPerformTransformationsOnTRT() throws Exception {
         final ArgumentCaptor<TDSReport> reportCaptor = ArgumentCaptor.forClass(TDSReport.class);
         processor.process(message);
 
         verify(tdsReportProcessor).process(reportCaptor.capture(), anyLong());
         final TDSReport report = reportCaptor.getValue();
         final List<Opportunity.Item> items = report.getOpportunity().getItem();
         assertThat(items.get(0).getBankKey()).isEqualTo(200);
+
+        //MC
```
malaffoon: I think this is fine so no need to change but ... another way to do this would be to have pairs of XML files: source and expected transformation. Then just read, transform and compare the file resources.

exam-processor/src/test/resources/TDSReport.iab.bankkey.xml
wtritch: good point, will do

wtritch: done:
bankkey.xls -> 2018_AIR_format_transformation.xls
TDSReport.iab.bankkey.xml -> TDSReport.iab.2018_AIR_format.xml

exam-processor/src/test/resources/bankkey.xsl
wtritch: done

exam-processor/src/test/resources/bankkey.xsl
wtritch: done:
```
  <!--
    These are characters that must be converted when "unescaping" embedded
    XML response data. (e.g. "&amp;lt;" will result in "<")
    Characters not present in the map below are left un-touched (e.g. "&amp;#39;" will result in "&#39;" rather than "'")
  -->
```

exam-processor/src/test/resources/bankkey.xsl
wtritch: Danke =)

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorTest.java
wtritch: I wanted them all in one because I wasn't sure whether the XLST would be performed recursively.  I also wasn't positive there wouldn't be selection conflicts when I started =)

malaffoon: Well, it could still be one big file with all the transformations. You could have many pairs of files.

PR #549, opened 2018-12-14
update dependencies

malaffoon:


build.gradle
```diff
@@ -10,21 +10,21 @@ buildscript {
...
-    id 'com.jfrog.artifactory' version '4.4.18'
+    id 'com.jfrog.artifactory' version '4.8.1'
 }
 
 apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
-String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.2.2-RELEASE"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.2.0-390"
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.3.0-SNAPSHOT"  // TODO
```
malaffoon: I will update these once i've committed the schema and common changes and we have updated artifacts.

build.gradle
```diff
@@ -10,21 +10,21 @@ buildscript {
     }
     dependencies {
         classpath("com.bmuschko:gradle-docker-plugin:3.6.2")
-        classpath("io.spring.gradle:dependency-management-plugin:1.0.0.RELEASE")
-        classpath("org.springframework.boot:spring-boot-gradle-plugin:1.5.2.RELEASE")
+        classpath("io.spring.gradle:dependency-management-plugin:1.0.6.RELEASE")
```
malaffoon: These changes will need to be made in RDW_Reporting as well.

common/src/test/java/org/opentestsystem/rdw/ingest/common/model/RdwImportQueryTest.java
```diff
@@ -103,7 +103,7 @@ public void itShouldParseInstant() {
                 .before("-P2DT8M")
                 .before("-pt10m")
                 .build();
-        assertThat(query.getBefore()).isLessThan(Instant.now());
+        assertThat(query.getBefore()).isBefore(Instant.now());
```
malaffoon: Interesting. Although `isBefore` existed in the older version, `isLessThan` basically did the same thing. In the new version, it is more particular.

group-processor/src/test/java/org/opentestsystem/rdw/ingest/group/service/impl/DefaultProcessingGroupImportServiceIT.java
```diff
@@ -106,6 +106,15 @@ public void itShouldUpdateModifiedGroupSubjects() {
                 ImmutableMap.of(), Integer.class)).isEqualTo(1);
     }
 
+    @Test
+    public void itShouldNotUpdateUnaffectedGroups() {
```
malaffoon: Not really  necessary, but i added this when i was playing and decided to leave it in.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacUser.java
```diff
@@ -66,9 +66,6 @@ public Builder copy() {
         return new Builder().copy(this);
     }
 
-    public static Builder builder() {
```
malaffoon: This i don't like, but i don't see a workaround. In the latest Spring security they added a builder for `User`. However, the way they did it, it can't be extended, so there is no way we can have a static `builder()` method in here. Probably that indicates we are fighting against the wrong approach but i'll have revisit it later.

PR #550, opened 2018-12-15
RP-21, make consistent with RDW_Reporting

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacUser.java
```diff
@@ -66,6 +66,16 @@ public Builder copy() {
...
     }
 
+    /**
+     * The base User has a static builder() method that returns the base Builder.
+     * Unfortunately, the base Builder cannot be extended so we can't override
+     * the builder() method. Thus the silly name.
+     *
+     * @return new builder
+     */
+    public static Builder builderExt() {
```
malaffoon: I like this better because you can still type `SbacUser.b` and it will autocomplete for you.

PR #551, opened 2018-12-18
RP-25, add migrate end-points

malaffoon:
This adds Spring actuator-style end-points to the migrate services to trigger a migrate, see the migrate job status, or re-enable a disabled migrate service.


migrate-olap/src/main/resources/application.yml
```diff
@@ -9,6 +9,11 @@ server:
     max-connections: 2
     max-threads: 2
 
+# enable lifecycle endpoints for pausing and resuming migration
```
malaffoon: Turns out the Spring upgrade i did changed the default behavior so explicitly enable pause/resume endpoints.

PR #553, opened 2018-12-24
RP-35, add language table to warehouse

malaffoon:
First chunk of changes, these go with schema changes: https://github.com/SmarterApp/RDW_Schema/pull/312


migrate-reporting/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -1094,9 +1094,9 @@ sql:
                 we.migrant_status,
```
malaffoon: To reduce changes happening all at once, this hacks things so the reporting database(s) don't have to change yet. Once the changes are made there, this will probably revert to just getting the language_id.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -154,7 +153,7 @@ private void parse(final CSVRecord record) {
                     .schoolYear(parserHelper.validate(logMsgPrefix + AcademicYear, toSchoolYear(record.get(AcademicYear)), toLong.andThen(toYear)))
                     .label(parserHelper.validate(logMsgPrefix + AssessmentLabel, record.get(AssessmentLabel), 255, true))
                     .version(parserHelper.validate(logMsgPrefix + AssessmentVersion, record.get(AssessmentVersion), 30, false))
-                    .gradeId(parserHelper.validate(logMsgPrefix + AssessmentGrade, record.get(AssessmentGrade), toGrade.andThen(gradeRepository::findIdByCode).andThen(new CheckNotNull<>())));
+                    .gradeId(parserHelper.validate(logMsgPrefix + AssessmentGrade, record.get(AssessmentGrade), toGrade.andThen(gradeRepository::findIdByCode)));
```
malaffoon: This is due to moving and sharing the GradeRepository: the findIdByCode already throws an exception so the checkNotNull is not necessary.

PR #554, opened 2018-12-26
RP-35, add language to reporting, olap

malaffoon:
There is a schema change to go with this: https://github.com/SmarterApp/RDW_Schema/pull/313


build.gradle
```diff
@@ -24,7 +24,8 @@ apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.3.0-105"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.3.0-396"
+//String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.3.0-396"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.3.0-SNAPSHOT"
```
malaffoon: will revert and update once schema is merged and built

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -599,6 +622,7 @@ sql:
                   CASE WHEN se.economic_disadvantage = true THEN 1 WHEN se.economic_disadvantage = false THEN 0 ELSE 2 END,
                   CASE WHEN se.migrant_status = true THEN 1 WHEN se.migrant_status = false THEN 0 ELSE 2 END,
                   COALESCE(se.elas_id, 1),
+                  COALESCE(se.language_id, 0),
```
malaffoon: I'm not 100% sure of this: We don't want null values in redshift so we default null values to something reasonable. For language, i see two choices: 0 ('eng') or 99 ('mis'). Based on the data i've seen in production, it feels like 0 ('eng') is the better choice.

migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -795,6 +822,7 @@ sql:
                 economic_disadvantage       = CASE WHEN se.economic_disadvantage = true THEN 1 WHEN se.economic_disadvantage = false THEN 0 ELSE 2 END,
                 migrant_status              = CASE WHEN se.migrant_status = true THEN 1 WHEN se.migrant_status = false THEN 0 ELSE 2 END,
                 elas_id                     = COALESCE(se.elas_id, 1),
+                language_id                 = COALESCE(se.language_id, 1),
```
malaffoon: doh, that's wrong

PR #555, opened 2018-12-31
RP-35, fix bug: language wasn't being migrated properly

malaffoon:


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -53,7 +53,7 @@
     static final String updateEmbargoStepName = "updateEmbargoStep";
     static final String finalizeStepName = "finalize";
 
-    public static final List<String> codes = newArrayList("administration_condition", "completeness", "elas", "ethnicity", "gender", "grade", "school_year");
+    public static final List<String> codes = newArrayList("administration_condition", "completeness", "elas", "ethnicity", "gender", "grade", "language", "school_year");
```
malaffoon: This is the fix: `language` needed to be in this list.

migrate-olap/src/test/java/org/opentestsystem/rdw/ingest/migrate/olap/step/UpsertCodesStepRST.java
```diff
@@ -29,12 +29,16 @@ public void itShouldDoNothingIfBatchDoesNotHaveCodeImportContent() {
         // collect counts of rows in each reporting code table before the step call
         final List<TableTestCountHelper> tableTestCounts = newArrayList();
         for (final String table : codes) {
-            if ("school_year".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), table, 0, "year IN (1999, 1998)"));
-            } else if ("grade".equals(table)) {
-                tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), table, 0, "id IN (108, 109)"));
-            } else {
-                tableTestCounts.add(new TableTestCountHelper(olapJdbcTemplate.getJdbcOperations(), table, 0, "id IN (-98, -99)"));
+            switch(table) {
```
malaffoon: This is only tangentially related: I find the if / else if / else construct harder to read and maintain than a switch statement. It is simple enough here (and i could probably have left it alone) but the one further down is more complex.

migrate-olap/src/test/resources/OlapCodesPreload.sql
```diff
@@ -1,7 +1,7 @@
 INSERT INTO school_year(year) VALUES (1999);
 INSERT INTO grade(id, code, sequence) VALUES (109, 'AA', 9), (-107, '27', 7), (-111, '31', 11);
 INSERT INTO elas(id, code) VALUES (-99, 'before test');
-INSERT INTO language(id, code) VALUES (-99, 'abc');
+INSERT INTO language(id, code) VALUES (-99, 'bef');
```
malaffoon: This change just makes the test more generic, all the "before" values start with `bef`

PR #556, opened 2019-01-03
fix warning about maxIdle > maxActive

malaffoon:


exam-processor/src/main/resources/application.yml
```diff
@@ -40,8 +40,10 @@ spring:
     validationQueryTimeout: 10000
     driverClassName: com.mysql.jdbc.Driver
     initialize: false
-    maxActive: 12
     initialSize: 6
+    maxActive: 12
+    minIdle: ${spring.datasource.initialSize}
+    maxIdle: ${spring.datasource.maxActive}
```
malaffoon: The WARN emitted in the log has been noted a couple times by the customer, so i'm fixing it. This just sets the min/max based on the other settings; it matches the default behavior and it matches the runtime adjustment made when the WARN is emitted.

PR #564, opened 2019-02-01
tweak so assessments with no item specifications can be loaded

malaffoon:


common/src/main/resources/application-ingest-common.yml
```diff
@@ -1,10 +1,10 @@
 validation:
   scaleScore:
-    min: 1500
+    min: 1000
```
malaffoon: ELPAC has lower minimum scores than SmarterBalanced tests. I have to wonder if these parameters need to be provided with subject definition instead of being system-wide.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -234,7 +235,12 @@ private Double toScore(final String scoreColHeader, final CSVRecord record) {
         }
 
         private void addItem(final CSVRecord record) {
```
malaffoon: This is the main change.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultItemParser.java
```diff
@@ -71,6 +72,8 @@ public DefaultItemParser(final ClaimRepository claimRepository,
 
     @Override
     public Item parse(final CSVRecord record, final DataElementErrorCollector errorCollector, final String asmtNaturalId, final int subjectId) {
```
malaffoon: .. along with this.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultSubjectProcessorIT.java
```diff
@@ -55,32 +49,19 @@
     @Autowired
     public ArchiveService archiveService;
 
-    @Before
```
malaffoon: unrelated: i don't like having a setup method preparing separate things for separate tests, so i moved each setup to each test.

PR #569, opened 2019-03-15
fix migrate enable to actually enable the job runner 

malaffoon:
Missed this corner case: after using the actuator end-point to enable migrate, triggering a migrate using the actuator end-point doesn't work.


migrate-common/src/test/java/org/opentestsystem/rdw/migrate/common/MigrateEndpointTest.java
```diff
@@ -41,7 +41,7 @@ public void itShouldGetGoodMigrateInfo() {
     }
 
     @Test
-    public void itShouldBadGoodMigrateInfo() {
+    public void itShouldGetBadMigrateInfo() {
```
malaffoon: unrelated: fix test name

PR #571, opened 2019-03-22
RP-182: create framework for running Groovy scripts from Java code

gregcharles:
Added module for Groovy code -- a framework for running Groovy scripts.


script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
```diff
@@ -0,0 +1,25 @@
+package org.opentestsystem.rdw.ingest.script
+
+/**
+ * Base class for Groovy pipeline scripts.
+ */
+abstract class DSLScriptBase extends Script {
+
+    DSLScriptBase bindProperties(Map<String, Object> properties) {
```
esotrope: i feel like we could get away with just putting "DSL" in the javadocs for explanation

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
```diff
@@ -0,0 +1,46 @@
...
+        script.bindProperties(properties);
+
+        assertThat(script.getProperty("key"), is("value"));
+        assertThat(script.getProperty("key2"), is("value2"));
+    }
+
+    @Test
+    public void itShouldBindInputStream() {
+        script.bindInput(inputStream);
+        assertThat(script.getProperty("input"), is(inputStream));
```
esotrope: are we eventually binding output too?

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
esotrope: what part of the evaluated script will be a DSL btw? do we plan on introducing custom expressions or just expose utility classes inside the groovy scripts?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void testCompilerError() {
+        try {
+            compiler = new ScriptCompiler("pruntln 'hello'; return 'OK");
+            fail("Expecting compile error");
+        } catch (CompilationFailedException cfe) {
+            assertThat(cfe.getMessage(), containsString("pruntln"));
+        }
```
malaffoon: Instead of this, consider using ExpectedException. I think we have at least one example usage somewhere, something like:
```
    @Rule
    public ExpectedException thrown = ExpectedException.none();

    @Test
    public void testCompilerError() {
        thrown.expect(CompilationFailedException.class);  // should be more specific if possible
        thrown.expectMessage(containsString("pruntln"));
        new ScriptCompiler("pruntln 'hello'; return 'OK'");
    }
```

script-common/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile 'org.codehaus.groovy:groovy-all'
+    compile 'org.jdom:jdom2:2.0.5'
+    compile 'jaxen:jaxen:1.1.4'
+
+    testCompile group: 'junit', name: 'junit', version: '4.12'
```
malaffoon: pretty sure this is inherited from parent

script-common/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+ext['groovy.version'] = '2.5.6'
+
+// This pulls in the dependency management which is great.
+// But it also defines silly boot tasks which we don't want here.
+apply plugin: 'org.springframework.boot'
+bootRepackage.enabled = false
+bootRun.enabled = false
+
+repositories {
+    mavenCentral()
```
malaffoon: this is inherited, did it not work without it?

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
```diff
@@ -0,0 +1,25 @@
+package org.opentestsystem.rdw.ingest.script
+
+/**
+ * Base class for Groovy pipeline scripts.
+ */
```
malaffoon: doh, it's hard to context switch to Groovy; i have to stop myself from making dumb comments

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.*;
```
malaffoon: please tweak your settings so this doesn't happen - set Java, Imports, "Class count..." and "Names count..." to 99.

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
```diff
@@ -0,0 +1,46 @@
+package org.opentestsystem.rdw.ingest.script;
```
malaffoon: Hmm, i'm going to have to pull this down and play with it ... should `src/test/groovy` have java-based tests in it? (that's a non-rhetorical question btw)

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
esotrope: interesting. I always used `@Test(expect = EX.class)`

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
esotrope: mine is 9999 ; p

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
malaffoon: If i don't care about the message i also prefer `expect=...`. This is just if you want to assert something about the exception.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
```diff
@@ -0,0 +1,25 @@
+package org.opentestsystem.rdw.ingest.script
```
malaffoon: I admit i haven't thought this all the way through but ...

I was expecting that there would be no groovy code in the actual codebase. That all the groovy scripts would be saved as part of the deployed environment. Perhaps we can discuss and you can help me grok this.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
gregcharles: DSLScriptBase seems to be something of a standard name for a Groovy script base class. There's not a whole lot of examples, but the couple I found used DSLScriptBase, so I went with it. There will be more meta-programing and DSL fun stuff in the next PR, which is for XML extensions. I probably should have combined those JIRAs.

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
gregcharles: No, there's no need. Scripts can support return values. They don't support input parameters unless you write them as functions though, and I thought it would be easier not to have to do that. We'll bind everything up in the setup code, and it will appear as inherent variables for the script writers.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
gregcharles: Thanks, I'd never used this rule, but I've updated tests now. If there's another example in RDW Ingest, I can't find it.

script-common/build.gradle
gregcharles: I just copied this from the build.gradle for the "common" module. You're right that bootRun.enabled seems unnecessary though, so I removed it.

script-common/build.gradle
gregcharles: I thought I'd copied this junit depedency from another build.gradle, but I must have copied it from the root one. I just removed it.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
gregcharles: Yeah, wait till I push some meta-programing over. You have to start thinking about Java vs. compiled Groovy vs. external script Groovy!

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
gregcharles: Ha! I did this a year ago, but apparently you need to do it separately for Groovy.

script-common/src/test/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBaseTest.java
gregcharles: Yeah, I wanted the ScriptCompilerTest specifically to be Java, because we'll be calling it from Java code. It works fine. However, when I also made the DSLScriptBaseTest a java file in the src/test/java, both Intellij and CLI can't find its tests at all ... not after a clean build, not after an invalidate and restart, not after any tweaking and cajoling I could think of.  In src/test/groovy, it works fine. However, I can rename that file to DSLScriptBaseTest.groovy, and that also works, so I'll just do that.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
gregcharles: The next push will include some examples of the DSL I'm starting to form. It's likely possible to still achieve this with a Java base class, but it's cleaner in Groovy. It would be possible to externalize this base class as we do the scripts themselves though. I think it's best to leave this the way it is for now, and think about refactoring it later.

PR #573, opened 2019-03-24
modifications for configurable subjects for migrate-olap

malaffoon:
There are no changes to aggregate reports for alt scores, so there aren't many required changes here. However, i chose to push the subject metadata and i tried to make the staging tables reflect the warehouse schema per our general staging philosophy.

A lot of the changes are tests and their data set up. The real changes are the SQL changes and the migrate configuration.

I've run all the tests multiple times but i admit there is a chance that, when deployed to awsqa, i may have to debug some more.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-109"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.4.0-413"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.4.0-SNAPSHOT"
 String dockerPrefix = "smarterbalanced"
 
 allprojects {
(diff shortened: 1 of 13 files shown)


build.gradle
```diff
@@ -22,7 +22,7 @@ apply plugin: 'io.spring.dependency-management'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
 String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-109"
-String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.4.0-413"
+String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.4.0-SNAPSHOT"
```
malaffoon: will update once schema build completes

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/step/OlapReportingMigrateStepsConfig.java
```diff
@@ -57,7 +57,7 @@
             "elas", "ethnicity", "gender", "grade", "language", "military_connected", "school_year");
```
malaffoon: Real changes.

PR #574, opened 2019-03-28
RP-183: Add XML helper methods to script framework

gregcharles:
This is the first phase toward creating a basic DSL for the script pipeline, and I've included a README.md to demonstrate some samples of how the scripts for XML transformation could be written. I'd appreciate any feedback on the approach and general and more specific design elements. E.g., does it make sense to dynamically add XML methods to the script class or would it be better to have separate base classes based on the type of  input being processed? Also, are there other helper methods that we could add?

### DIFF
--- a/script-common/README.md
+++ b/script-common/README.md
@@ -2,13 +2,13 @@
 
 ### Helper functions for XML
 XML support can be added to a Script class by calling:
- ```script.addXmlExtensions()```. We assume that an XML document
+ `script.addXmlExtensions()`. We assume that an XML document
  will be converted into another XML document, and all elements, including
- commentsnot, not specifically altered, should appear in the converted
+ comments not specifically altered, should appear in the converted
  document exactly as they appear in the original. 
 
 This will automatically parse the input into a JDOM 2 Document, 
- which is made available to the scripts in the inherent variable ```document```.
+ which is made available to the scripts in the inherent variable `document`.
  However, it will also enable some helper functions that act implicitly
  on the document, and may ease writing XML transformation scripts and
  make it unnecessary for the script writer to ever reference the
@@ -24,13 +24,13 @@ transform '//Movie' by { movie ->
 }
 ```
 
-Here ```//Movie``` is an XPath expression that returns all Movie elements
+Here `//Movie` is an XPath expression that returns all Movie elements
 in the document. The code within the braces will be applied to each
-matched elements, and follows general Groovy syntax rules. The ```movie ->```
+matched elements, and follows general Groovy syntax rules. The `movie ->`
 syntax just
 assigns a name for the element that the code within the braces can use.
 It can be any string that makes sense for the script, or it can also be omitted. In this case,
-the element is automatically assigned the name ```it```. 
+the element is automatically assigned the name `it`. 
 
 For example, this script works exactly the same as the first one:
 
@@ -42,11 +42,11 @@ transform '//Movie' by {
 }
 ```
 
-The syntax ```movie.title``` (or ```it.title```) refers to the title
-attribute of the movie element. It is shorthand for ```movie.getAttributeValue()```,
-and ```movie.title = 'some text'``` is shorthand for ```movie.setAttribute('title', 'some text')```.
-An exception to this rule is ```movie.text```, which is shorthand for
-```movie.getText()```, and returns the text content of the movie element
+The syntax `movie.title` (or `it.title`) refers to the title
+attribute of the movie element. It is shorthand for `movie.getAttributeValue('title')`,
+and `movie.title = 'some text'` is shorthand for `movie.setAttribute('title', 'some text')`.
+An exception to this rule is `movie.text`, which is shorthand for
+`movie.getText()`, and returns the text content of the movie element
 instead of an attribute value.
 
 For the transformation rule, any legal Groovy (or Java) syntax is allowed.
@@ -92,9 +92,9 @@ writing Groovy rules. Therefore, a helper function for XSL has been provided:
 applyXsl xsl
 ```
 
-The ```xsl``` variable here is defined in the configuration for the script,
+The `xsl` variable here is defined in the configuration for the script,
 and can be a string containing XSL rules, a file represented as a
-```java.io.File``` object, or a ```java.io.InputStream``` object.
+`java.io.File` object, or a `java.io.InputStream` object.
 
 It is also possible to combine XSL transformations with additional
 Groovy rules:
--- a/script-common/README.md
+++ b/script-common/README.md
@@ -1,0 +1,131 @@
+## RDW Ingest - Script Pipeline
+
+### Helper functions for XML
+XML support can be added to a Script class by calling:
+ ```script.addXmlExtensions()```. We assume that an XML document
+ will be converted into another XML document, and all elements, including
+ commentsnot, not specifically altered, should appear in the converted
+ document exactly as they appear in the original. 
+
+This will automatically parse the input into a JDOM 2 Document, 
+ which is made available to the scripts in the inherent variable ```document```.
+ However, it will also enable some helper functions that act implicitly
+ on the document, and may ease writing XML transformation scripts and
+ make it unnecessary for the script writer to ever reference the
+ document object explicitly. 
+ 
+ Some examples follow:
+#### Transform
+ ```groovy
+transform '//Movie' by { movie -> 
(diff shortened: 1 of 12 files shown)


script-common/src/test/resources/xsl/exam.xsl
```diff
@@ -0,0 +1,174 @@
+<!DOCTYPE xsl:stylesheet >
+<xsl:stylesheet
```
esotrope: are the xsl used to test feature parity? or will it be an option to use them in the pipeline?

script-common/README.md
```diff
@@ -0,0 +1,131 @@
+## RDW Ingest - Script Pipeline
+
+### Helper functions for XML
+XML support can be added to a Script class by calling:
+ ```script.addXmlExtensions()```. We assume that an XML document
+ will be converted into another XML document, and all elements, including
+ commentsnot, not specifically altered, should appear in the converted
```
malaffoon: nit: typo

script-common/README.md
```diff
@@ -0,0 +1,131 @@
+## RDW Ingest - Script Pipeline
+
+### Helper functions for XML
+XML support can be added to a Script class by calling:
+ ```script.addXmlExtensions()```. We assume that an XML document
```
malaffoon: nit: when inlining mono formatting you can surround with single back-ticks. I'd give an example, but this comment is also markdown and i'm too lazy to figure out how to suppress the formatting in the example.

script-common/README.md
```diff
@@ -0,0 +1,131 @@
...
+ ```groovy
+transform '//Movie' by {  
+    if (it.title == 'Jaws') {
+        it.title = it.title.toUpperCase()
+    }
+}
+```
+
+The syntax ```movie.title``` (or ```it.title```) refers to the title
+attribute of the movie element. It is shorthand for ```movie.getAttributeValue()```,
```
malaffoon: The `movie.getAttributeValue()` example is confusing. Should it be `movie.getAttributeValue('title')`?

script-common/README.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+It is also possible to reverse this logic to delete all elements
+(matching the XPath expression) that don't match the rule:
+ 
+```groovy
+delete '//Movie' unless { movie -> 
+    movie.actor == 'Richard Basehart'
+}
+```
```
malaffoon: Is there nice syntax for applying a transform while using when/unless?

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
```diff
@@ -19,7 +34,116 @@ abstract class DSLScriptBase extends Script {
...
+     * @return reference to the script class (for chained calls).
+     */
+    DSLScriptBase enableXmlExtensions() {
+        Document document = new SAXBuilder().build(getProperty("input"))
+
+        this.setProperty("document", document)
+
+        // Unescapes up embedded XML in TRT values so it can treated as regular XML.
+        String.metaClass.unescapeXml {
+            delegate.replaceAll('&lt;','<').replaceAll('&gt;','>')
```
malaffoon: Is there not a helper in the XML libraries for this? I ask because there are additional sequences that need to be unescaped, aren't there?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptXmlTest.java
```diff
@@ -0,0 +1,258 @@
...
+    public void itShouldFilterOutElements() {
+        // Test ability to filter out elements with a boolean rule. All elements matching the XPath expression
+        // and also matching this rule will be removed.
+        String scriptDefinition =
+                "delete '//Item' when { item ->" + LF +
+                "    item.key == '37115' || item.key == '92135'" + LF +
+                        "}" + LF +
+                "outputXml";
+
+        Assert.assertThat(originalXml, hasXPath("//Item[@key='37115']", not(isEmptyOrNullString())));
```
malaffoon: strongly recommend statically importing `assertThat`

script-common/build.gradle
```diff
@@ -15,5 +15,10 @@ dependencies {
     compile 'org.codehaus.groovy:groovy-all'
     compile 'org.jdom:jdom2:2.0.5'
     compile 'jaxen:jaxen:1.1.4'
+    compile 'net.sf.saxon:Saxon-HE'
+
+    testCompile 'org.xmlunit:xmlunit-core'
+    testCompile 'org.xmlunit:xmlunit-matchers'
+    testCompile 'org.springframework.boot:spring-boot-starter-test'
```
malaffoon: curious why you'd need spring-boot for all this

script-common/src/test/resources/xsl/exam.xsl
gregcharles: The xsl can be called from the script in place of the coded rules, or to complement them. I do have one test case that performs the complete transformation by applying this xsl file via a Groovy script, and another one that uses a reduced set of xsl transformations and makes up the rest with script commands.

script-common/README.md
gregcharles: Oh, right. "comment snot" should have a space in it. I'll fix that!

script-common/README.md
gregcharles: OK, good to know! I'm changing these.

script-common/README.md
gregcharles: It's just wrong. I'll fix that. 

You're right that Groovy usually maps movie.title to movie.getTitle(). However, I have set it up to override that behavior, and instead call getAttributeValue('propertyname'). I sort of wonder if I shouldn't also get sub-elements with the same syntax. If an attribute name collided with a child element name though, it would get messy.

script-common/README.md
gregcharles: You'd want to transform something inside a delete command? Do you mean to transform the ones you don't delete?  You can actually do any of the transformations here that you can do in a transform command. The key is that you have to return true or false, which determines whether the element will be kept or discarded. (The final statement in Groovy automatically becomes the return value, but you can also have explicit returns.)

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptXmlTest.java
gregcharles: I'll see what I can do. I copied these tests, with some tweaks, from the ExamProcessorTest, and they include two assertThat methods from different classes. Obviously, only one can be statically imported. I'll try to figure out something clever to make that less ugly though.

script-common/build.gradle
gregcharles: Yes, you're good. It looks like I just need it for a Hamcrest matcher (in org.hamcrest.text)
that it includes, but that my other dependencies don't. I'll either just bring in that library, or just implement my own matcher.

script-common/README.md
malaffoon: Sorry, my comment wasn't clear. I'm thinking of conditional transform, perhaps something like:
```
transform '//Movie' { it.title = it.title.toUppercase() } 
when { it.title == 'Jaws' }
```

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptXmlTest.java
gregcharles: I stand corrected. You can statically import methods with the same name from two different classes, I guess as long as the signatures are different. You learn something every day!

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
gregcharles: Yes, I had a lot of trouble with this one. There are, in fact, builtin XML unescape/escape methods and they are called automatically when parsing and formatting the XML. What seems to be happening is that TI elements in the TRT have escaped XML in their text content, which is unescaped by the parser. However the XML has a <value> tag that itself contains an escaped HTML table definition in it. That means the html is doubly escaped, and so after parsing is still singly escaped. That would actually be fine, except the transform rule then tosses out the rest of the XML and just keeps the HTML part.  

tldr; It's sort of a cluster. I changed the name of this method to unescapeHtmlTags(), but I'll keep thinking about it.

PR #576, opened 2019-04-08
RD-185: implement script pipeline

gregcharles:
Implements the pipeline framework. Also converted project from Groovy to straight Java. All Groovy source will now be loaded from external sources and compiled at runtime.


script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineBuildExcepition.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.Map;
+
+/**
+ * Exception class for reporting all script compile errors in a pipeline.
+ */
+public class PipelineBuildExcepition extends Exception {
+    private final Map<ScriptDefinition,Exception> buildErrors;
+
+    public PipelineBuildExcepition(final String message, final Map<ScriptDefinition, Exception> buildErrors) {
```
esotrope: what's an excepition? x P

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ConfigurationCompiler.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * TODO: this is stub. The idea is this class will take the string-based properties provided by the ScriptSource
+ * and enhance some of them into object properties, e.g., data sources or web service clients.
+ */
+public class ConfigurationCompiler {
+    public Map<String,Object> compile(Map<String,String> properties) {
+        Map<String,Object> configuration = new HashMap<>();
```
malaffoon: nit: variable and param can both be `final`

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineFactory.java
```diff
@@ -0,0 +1,64 @@
...
+public class PipelineFactory {
+    // Interface for loading the script base class code, the pipeline definition, and the script code.
+    private final ScriptSource scriptSource;
+
+    // Compiler of Groovy scripts into Java byte code.
+    private final ScriptCompiler scriptCompiler;
+
+    // Compiler of configuration properties, which may involve loading database connections and
+    // other manipulations
+    private final ConfigurationCompiler configCompiler;
```
malaffoon: Can we put a full javadoc comment on the constructor that has these comments in it?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineBuildExcepition.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.ingest.script;
+
+import java.util.Map;
+
+/**
+ * Exception class for reporting all script compile errors in a pipeline.
+ */
+public class PipelineBuildExcepition extends Exception {
+    private final Map<ScriptDefinition,Exception> buildErrors;
```
malaffoon: I'll have to see where this is being used. How are the collected exceptions being handled when logging, storing results, etc. I think this probably makes sense so the downstream handler has all the data it needs. I'm a bit concerned that the exceptions by themselves don't always have enough detail. But, again, i'll wait to see how it works out before recommending a change.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineScript.java
```diff
@@ -0,0 +1,28 @@
...
+        if (properties != null) {
+            for (Map.Entry<String,Object> entry : properties.entrySet()) {
+                this.setProperty(entry.getKey(), entry.getValue());
+            }
+        }
+
+        return this;
+    }
+
+    public PipelineScript bindInput(Reader input) {
```
malaffoon: We're using a `Reader` because ... ? Just worried about the script having to deal with mark/reset and other Reader fiddly bits.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptDefinition.java
```diff
@@ -0,0 +1,114 @@
...
+                .bindInput(input);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("'%s' (%s type), v. %s", name, type, version);
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
malaffoon: do we know these are necessary?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptDefinition.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Get an instance of the script to be run against the given input
+     * @param input an open Reader with the input file for the script to work on
+     * @return an instance of a PipelineScript
+     *
+     * @throws Exception
+     */
+    public PipelineScript getInstance(Reader input) throws Exception {
```
malaffoon: Hmm, do we want to separate ScriptDefinition from PipelineScriptFactory? At the very least we could introduce an interface that ScriptDefinition implements.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptPipelineConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+    @Bean
+    public PipelineFactory pipelineFactory(ScriptSource scriptSource, ScriptCompiler scriptCompiler) {
+        return new PipelineFactory(scriptSource, scriptCompiler, new ConfigurationCompiler());
+    }
+
+    @Bean
+    public ScriptSource scriptSource() {
+        return new ScriptSourceImpl();
+    }
+}
```
malaffoon: nit: all the params and variables can be `final`

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptSourceImpl.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Collections;
+import java.util.Optional;
+
+/**
+ * TODO: This is a stub. It must be replaced with a real implementation that loads scripts from an s3 bucket,
+ * a private GitHub repo, or other source.
+ */
+public class ScriptSourceImpl implements ScriptSource {
```
malaffoon: nit: then i would call it `StubScriptSource`  (i'm not a huge fan of `XxxImpl` naming)

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineDefinition.java
```diff
@@ -0,0 +1,47 @@
...
+ * Definition of a Pipeline including its name, version, node definitions, and roll back information..
+ */
+public class PipelineDefinition {
+    private final String name;
+    private final String version;
+    private final Optional<String> rollbackVersion;
+    private final List<String> options;
+    private final List<ScriptDefinition> scriptDefinitions;
+
+    public PipelineDefinition(final String name,
```
malaffoon: We have a general policy of using a builder once there are more than 3 constructor params. Recommend doing that here during a cleanup step.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptSourceImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    private String loadFromResourceFile(String name) {
+        try {
+            Reader scriptReader = new InputStreamReader(this.getClass().getResourceAsStream(name));
+            return readFully(scriptReader);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to load " + name, e);
+        }
+    }
+
+    private String readFully(Reader reader) throws IOException {
```
malaffoon: Hmm, isn't this available in a common library, guava or apache commons?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
```diff
@@ -1,72 +1,75 @@
 package org.opentestsystem.rdw.ingest.script;
 
 import groovy.lang.Script;
-import org.apache.tools.ant.filters.StringInputStream;
+import org.apache.tools.ant.util.FileUtils;
```
malaffoon: Hmm, this is an odd package to be importing from. Is there a more common one we can use?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptCompilerTest.java
```diff
@@ -1,72 +1,75 @@
...
         exception.expect(RuntimeException.class);
-        compiler = new ScriptCompiler("return 1 / 0");
 
-        compiler.getInstance(null).run();
+        Class<? extends Script> scriptClass = compiler.compile("return 1 / 0");
+        scriptClass.getConstructor().newInstance().run();
+    }
+
+    private String loadFromResourceFile(String name) throws Exception {
+        Reader scriptReader = new InputStreamReader(this.getClass().getResourceAsStream(name));
```
malaffoon: commons io has IOUtils.toString() to do this, avoids creating a reader.

For either, i think the Charset should be explicitly set to "UTF-8" or use Charset.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptDefinitionTest.java
```diff
@@ -0,0 +1,70 @@
...
+        Map<String,Object> configuration = Collections.singletonMap("test", "value");
+        scriptDefinition = new ScriptDefinition("test", "1.0", NodeType.USER);
+        scriptDefinition.setScriptClass(MockScript.class);
+        scriptDefinition.setConfiguration(configuration);
+        input = new StringReader("Hello");
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        input.close();
```
malaffoon: nit: i guess it's good practice but this doesn't really do anything with a StringReader

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptDefinitionTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void itShouldBindExtraPropertiesToScriptProperties() throws Exception {
+        PipelineScript script = scriptDefinition.getInstance(null,
+                Collections.singletonMap("answer", 42));
+
+        assertThat(script.getProperty("test"), is("value"));
+        assertThat(script.getProperty("answer"), is(42));
+    }
+}
+
+class MockScript extends PipelineScript {
```
malaffoon: This will work, just suggesting the alternate approach of a mock object with `when().thenAnswer()` behavior

script-common/src/test/resources/sample_xml/TDSReport.iab.2018_AIR_format.xml
```diff
@@ -77,32 +77,32 @@
...
-        <Accommodation code="TDS_PM0" type="Permissive Mode" value="Permissive Mode Disabled" segment="0"/>
-        <Accommodation code="TDS_PS_L0" type="Print Size" value="No default zoom applied" segment="0"/>
-        <Accommodation code="TDS_RSL_ListView" type="Review Screen Layout" value="List" segment="0"/>
-        <Accommodation code="TDS_SCNotepad" type="Student Comments" value="On" segment="0"/>
-        <Accommodation code="TDS_ST1" type="Strikethrough" value="True" segment="0"/>
-        <Accommodation code="TDS_SVC1" type="System Volume Control" value="Show widget" segment="0"/>
-        <Accommodation code="TDS_TPI_ResponsesFix" type="Test Progress Indicator" value="Show indicator as a fraction and adjust to test length" segment="0"/>
-        <Accommodation code="TDS_TS_Universal" type="Test Shell" value="Universal Shell" segment="0"/>
-        <Accommodation code="TDS_WL_Glossary" type="Word List" value="English Glossary" segment="0"/>
+        <Accommodation scriptCode="ENU" type="Language" value="English" segment="0"/>
```
malaffoon: was this intentional??

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptSourceImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public ScriptDefinition loadScriptCodeAndConfig(final ScriptDefinition scriptDefinition) {
+        return scriptDefinition;
+    }
+
+    private String loadFromResourceFile(String name) {
+        try {
+            Reader scriptReader = new InputStreamReader(this.getClass().getResourceAsStream(name));
```
malaffoon: The charset should be explicitly specified.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptDefinition.java
malaffoon: After thinking about it some more, i really think we should separate the script factory from the script definition.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/PipelineBuildExcepition.java
malaffoon: hint: that extra `i` could be used elsewhere

gregcharles: Ah, got it. :) I need to start using a bigger font.

esotrope: I like how my comment was interpreted as a real question and made you think critically about the exception handling though - that was totally unintended x D

PR #582, opened 2019-04-16
review pipeline framework

malaffoon:
@gregcharles, since Nohemi is making me test RP-182, i dove into it a bit. These changes are part of that exercise. Mostly, i tried to understand stuff, fix warnings and get coverage up. I'll comment inline, let me know what you think.


script-common/build.gradle
```diff
@@ -10,11 +10,12 @@ repositories {
...
 
 dependencies {
+    compile 'commons-io:commons-io'
+    compile 'jaxen:jaxen:1.1.4'
+    compile 'net.sf.saxon:Saxon-HE'
     compile 'org.springframework.boot:spring-boot-starter-jdbc'
     compile 'org.codehaus.groovy:groovy-all:2.5.6'
     compile 'org.jdom:jdom2:2.0.5'
-    compile 'jaxen:jaxen:1.1.4'
-    compile 'net.sf.saxon:Saxon-HE'
```
malaffoon: Just alphabetizing

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -21,19 +21,19 @@ public PipelineDefinition getPipelineDefinition() {
         return pipelineDefinition;
     }
 
-    public String runPipeline(String inputString)  {
+    public String runPipeline(final String inputString)  {
```
malaffoon: a lot of these `final` things ... my IJ is set to warn when something can be final

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/PipelineDefinitionTest.java
```diff
@@ -0,0 +1,62 @@
...
+                .name("test")
+                .version("23")
+                .rollbackVersion("16")
+                .options(ImmutableList.of("xml"));
+
+        assertThat(builder.toString()).contains("test").contains("23").contains("16").contains("xml");
+    }
+
+    @Test
+    public void equalsIsBroken() {
```
malaffoon: @gregcharles, this. There are child classes of VersionedResource that don't overload equals/hashCode. I didn't fix the problem 'cause i wasn't sure what you had in mind, just created a test demonstrating the problem.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptDefinitionTest.java
```diff
@@ -6,11 +6,10 @@
 
 import java.io.Reader;
 import java.io.StringReader;
-import java.util.Collections;
 import java.util.Map;
 
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
+import static java.util.Collections.singletonMap;
+import static org.assertj.core.api.Assertions.assertThat;
```
malaffoon: Trying to be consistent with other tests. Probably just made things worse.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
```diff
@@ -1,149 +0,0 @@
-package org.opentestsystem.rdw.ingest.script
```
malaffoon: I was confused about the location of this file. First, it was in there twice; i'm assuming this was left over from earlier work so i deleted it. The other copy was under `test` resources but i'm pretty sure it needs to be under `main` resources. I realize this is subject to change as we set up the real pipelines but ...

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/ScriptDefinition.java
```diff
@@ -11,9 +11,9 @@
     private final Class<? extends PipelineScript> compiledScript;
     private final Map<String,Object> configuration;
 
-    ScriptDefinition(String name, String version, ResourceType type,
-                     Class<? extends PipelineScript> compiledScript,
-                     Map<String, Object> configuration)
+    private ScriptDefinition(final String name, final String version, final ResourceType type,
+                             final Class<? extends PipelineScript> compiledScript,
+                             final Map<String, Object> configuration)
```
esotrope: each param should each have their own line probly

malaffoon: oops, didn't notice that ... will fix

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/PipelineDefinitionTest.java
malaffoon: Do we think we need equals/hashCode implemented in the base class, can we just remove it? Are we loading into maps or anything where we want value-based equality?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptDefinitionTest.java
```diff
@@ -39,37 +38,52 @@ public void tearDown() throws Exception {
...
-    public PipelineScript enableXmlExtensions() {
-        return this;
+        assertThat(scriptDefinition.equals(anotherDefinition)).isTrue();
     }
 
-    @Override
-    public Object run() {
-        return "Test Results";
+
+    static class MockScript extends PipelineScript {
```
malaffoon: moved the helper class inside the test suite

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
gregcharles: Yes, I got that message -- finally. (heh) I'd argue that the cons (needless clutter) outweigh the pros (bit of extra communication of intent), but I'll follow the standard.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/ScriptDefinitionTest.java
gregcharles: Ugh, yes, I guess this is fine. I vastly prefer using the JUnit assertThat and Hamcrest, but if this is the standard I'll try to make it work.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/PipelineDefinitionTest.java
gregcharles: I just use it in the code compiler, where I have a map of ScriptDefinitions to compile errors, and I intend to rework that class eventually anyway. I really did intend these types (scripts and pipelines basically) to be uniquely determined by name, version, and type, so in that sense the equals test isn't broken. However, if you think that's dangerous or confusing, I'll implement equals and hash in the specific classes.

script-common/src/main/groovy/org/opentestsystem/rdw/ingest/script/DSLScriptBase.groovy
gregcharles: Yeah, that's confusing, right? It's naturally in test for the test cases, but then I wanted to be able to exercise the running application and I don't have the S3 script loader available yet, so I temporarily put it into main as well. I will delete it from test on my next checkin, and then just remember to restore it when I delete the one in main. 

Good catch on the package. IntelliJ really, really wants to put it in there. It doesn't hurt, but it's also unnecessary.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/PipelineDefinitionTest.java
malaffoon: Ok, in that case, i'll just rename the test ...

PR #583, opened 2019-04-18
RP-276, set default admin condition based on assessment type

malaffoon:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -148,8 +149,15 @@ public Exam parseExam(final TDSReport report, final int schoolId, final Assessme
...
 
-    private Integer toValidAdministrationCondition(final String value) {
-        return (Integer) new CheckNotNull<>().apply(adminConditionRepository.findIdByCode((value == null || value.isEmpty()) ? "SD" : value));
+    private Integer parseAdministrationCondition(final Opportunity opportunity,
+                                                 final Assessment assessment,
+                                                 final ExamParserHelper parserHelper) {
+
+        // the admin condition may be blank with a default that depends on the type of assessment
+        final String code = hasText(opportunity.getAdministrationCondition()) ? opportunity.getAdministrationCondition()
+                : assessment.getTypeId() == AssessmentType.SUMMATIVE.id() ? "Valid" : "SD";
```
malaffoon: Previously it always defaulted to "SD" which is fine for interims but not summatives.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessorTest.java
```diff
@@ -377,13 +379,29 @@ public void itShouldReturnErrorsForConfiguredRequiredElements() {
                     "{\"elementName\":\"ExamItems\",\"value\":\"none\",\"error\":\"must provide at least one value for ExamItems\"}," +
                     "{\"elementName\":\"OppId\",\"error\":\"value may not be blank\"}," +
                     "{\"elementName\":\"SessionId\",\"error\":\"value may not be blank\"}," +
-                    "{\"elementName\":\"Completeness\",\"error\":\"invalid value\"}" +
-                    ",{\"elementName\":\"AdministrationCondition\",\"error\":\"invalid value\"}," +
+                    "{\"elementName\":\"Completeness\",\"error\":\"invalid value\"}," +
+                    "{\"elementName\":\"AdministrationCondition\",\"value\":\"SD\",\"error\":\"invalid value\"}," +
```
malaffoon: my change (inadvertently) improved the error message

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
esotrope: still feels weird that we clump summative and interim admin conditions into one abstraction even when they don't apply to all assessment types. feels like they at least need an assessment type column or something so we can know which apply to which assessment type in the data

malaffoon: I don't disagree but i want a quick fix!  :)

PR #584, opened 2019-04-26
RP-283: create a system script to validate imported exams

gregcharles:
For the system scripts, I went with a "property resolver", which will expose to the system pipeline scripts any Spring beans that the ingest process chooses to let them see. For the exam processor, I made that most of the repositories and a couple of the services. I also exposed a couple of utility classes this way, which isn't necessary, but it allows the scripts to use them without needing import statements.

The system scripts use this property resolver via a "missing property" handler, so it's pretty easy to add or subtract what we expose, but would take a code change. I exposed more repo beans than the current validations use so that other kinds of validators will be possible.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/util/TrtUtils.java
```diff
@@ -1,7 +1,85 @@
...
+        }
+        return finalNames.isEmpty() ? initialNames : finalNames;
+    }
+
+    /**
+     * The Section504Status attribute has a legacy value that indicates "unknown" which maps to null.
+     * Otherwise it is a required boolean value.
+     */
+    public static final Function<String, Boolean> toSection504Status = value -> {
+        if ("Unknown/Cannot Provide".equalsIgnoreCase(value)) return null;
```
esotrope: magic string ✨

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/util/TestUtils.java
```diff
@@ -1,7 +1,44 @@
...
+        byte[] buffer = new byte[1024];
+        int len;
+
+        // read bytes from the input stream and store them in buffer
+        while ((len = in.read(buffer)) != -1) {
+            // write bytes from the buffer into output stream
+            os.write(buffer, 0, len);
+        }
+
+        return os.toByteArray();
```
esotrope: apache io utils didn't have this?

script-common/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -22,6 +26,130 @@ import javax.xml.transform.stream.StreamSource
...
+
+    /**
+     * Resolves missing properties by delegating to custom property resolver if one has been set.
+     *
+     * @param name the name of the property to resolve
+     * @return an object that the property resolver maps to the name.
+     * @throws MissingPropertyException if name cannot be resolved
+     */
+    def propertyMissing(String name) {
+        // Avoid endless recursion
```
malaffoon: I don't get it. How will it recurse?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/util/TestUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    public static String loadStringFromResourceFile(String name) throws Exception {
+        return new String(loadFromResourceFile(name), StandardCharsets.UTF_8);
+    }
+
+    public static ScriptInfo loadScriptFromFile(String filePath, Map<String,String> properties)  throws Exception {
+        final String scriptCode = loadStringFromResourceFile(filePath);
+        return new ScriptInfo(scriptCode, properties);
+    }
+
+    private static byte[] readFully(InputStream in) throws IOException {
```
malaffoon: `IOUtils.toByteArray`

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -91,16 +88,7 @@
 
     @Override
     public int parseSchoolId(final Examinee examinee) throws ImportException {
```
malaffoon: It feels like we could get rid of this method and just code it in the pipeline?

exam-processor/src/main/resources/scripts/ExamPipeline.json
```diff
@@ -4,9 +4,16 @@
...
+  "nodeDefinitions": [
+    {
+      "name": "trt_transform",
+      "version": "1.0",
+      "type": "USER_NODE"
+    },
+    {
+      "name": "trt_validate",
+      "version": "1.0",
+      "type": "SYSTEM_NODE"
```
malaffoon: After going over this i wonder if node type should be `transform` and `validate` and there should be a `system` flag. So the user node would be a non-system `transform` and the system node would be a system `validate`.

exam-processor/src/main/resources/scripts/trt_validate.groovy
```diff
@@ -0,0 +1,71 @@
...
+        errorImmediately 'UNKNOWN_SCHOOL',
+                "unable to find a school with natural id [${naturalId}]. School name: ${schoolName}, " +
+                        "district id: [${districtId}], district name: ${districtName}"
+    }
+}
+
+//
+// Validate the TRT test against the data store assessment.
+//
+def validateTest(def test, def assessment) {
```
malaffoon: Personally i would inline this and not have a helper method. I don't think encapsulating these two validations is particularly valuable.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/util/TrtUtils.java
esotrope: No, it is magical : J 💖

gregcharles: hmm, agree to disagree :) Overuse of constants is also an anti-pattern. 

in any case, Mark wrote this code. I just moved it so that the system scripts can invoke it. I disavow all knowledge and responsibility! Nyah, nyah!

script-common/src/main/resources/scripts/DSLScriptBase.groovy
gregcharles: The propertyResolver is only set onto system scripts, so if we have missing property in a user script, then it'll try to get resolve it with the propertyResolver, which will missing, so it will try to resolve that with propertyResolver ... 

I admit that I could just set it to null for user scripts, but this seemed a bit more fool proof.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/util/TestUtils.java
gregcharles: OK, OK. :) IOUtils has a number of readFully() methods that all require you to know the length up front in some way. I didn't notice that they had other readFully() methods under a different name. It literally took me less time to code my readFully than it did to type this comment. :)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
gregcharles: Hmm, it's true. Although schoolId used in the processing, not just validation, it is parsed out by the examinee processor only to be handed back to the examinee processor again. It could just be retrieved in a local method. It is passed to two different examinee processor methods though, so keeping it this way saves a db call. And also some code changes!

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/SystemScriptPropertyResolver.java
```diff
@@ -0,0 +1,66 @@
...
+            final GenderRepository genderRepository,
+            final SchoolRepository schoolRepository,
+            final StudentRepository studentRepository,
+            final SubjectRepository subjectRepository,
+            final ExamineeProcessor examineeProcessor,
+            final StudentExamProcessor studentExamProcessor) {
+
+        resolvers.put("assessmentRepository", assessmentRepository);
+        resolvers.put("ethnicityRepository", ethnicityRepository);
+        resolvers.put("examRepository", examRepository);
```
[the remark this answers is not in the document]
gregcharles: OK, I've split the ExamRepository into a read only base and full functionality subclass. I won't do it for the other repos for now, but maybe a maintenance ticket in case there's time later?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentService.java
```diff
@@ -25,23 +19,6 @@
...
-        if (!assessment.getSubjectCode().equalsIgnoreCase(test.getSubject()))
-            errorCollector.add(new DataElementError("Assessment natural id [" + assessmentNaturalId + "], Test subject", test.getSubject(), "exam subject does not match asmt subject"));
-
-        if (!assessment.getGradeCode().equalsIgnoreCase(parserHelper.validate("grade", test.getGrade(), toGrade)))
-            errorCollector.add(new DataElementError("Assessment natural id [" + assessmentNaturalId + "], Test grade", test.getGrade(), "exam grade does not match asmt grade"));
-
-        if (errorCollector.isEmpty()) return assessment;
-
-        throw new ImportException(ImportStatus.BAD_DATA, errorCollector.toJson());
+        return assessmentRepository.findOneByNaturalId(test.getName());
```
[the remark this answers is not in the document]
gregcharles: Well, I kept the service, but changed it to take the natural Id. Even though it just calls the repo method, it seems nicer to me to have the facade it's in work on a bunch of services than a bunch of services plus one repository.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -19,26 +26,43 @@ public PipelineDefinition getPipelineDefinition() {
...
         for (ScriptDefinition sd : scriptDefinitions) {
+            // For now, assume that user nodes transform the data and system nodes validate
+            boolean isTransform = sd.getType() == USER_NODE;
 
             try {
-                PipelineScript script = sd.getInstance(input);
-                if (pipelineDefinition.getOptions().contains("xml")) {
+                PipelineScript script = sd.getInstance(lastValidTransformation);
+
+                if (isTransform && pipelineDefinition.getOptions().contains("xml")) {
```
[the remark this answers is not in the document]
gregcharles: It is sort of an optimization. When the xml extensions are enabled, I parse the XML into an org.jdom.Document, because the extension methods need it to be in that form. However, all the validation logic (as written) is done on the Report object returned from XmlUtils.tdsReportFromXml(), so the validator doesn't need the other type of parsing or the extension methods. Will system scripts of other pipelines need this though? I don't know.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -19,26 +26,43 @@ public PipelineDefinition getPipelineDefinition() {
...
                     script.enableXmlExtensions();
                 }
 
-                input = script.run();
+                final Object results = script.run();
+
+                if (results == null) {
+                    // (A script always returns the value of its last expression, so this is unlikely to happen.)
+                    logger.warn("Script {} returned null results.", sd.toString());
```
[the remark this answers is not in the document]
gregcharles: I could break out of the loop. As I say, it's unlikely. It would take the script right literally putting in "return null" somewhere. I guess erroring out is the best way to handle that.

PR #585, opened 2019-04-28
use common assessment parser; moves some validation around

malaffoon:


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentParser.java
```diff
@@ -1,361 +1,198 @@
 package org.opentestsystem.rdw.ingest.processor.service.impl;
```
malaffoon: The meat of the changes are in these two files. The common library now handles the CSV parsing and some validation. That significantly changes how this loads and validates things.

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseAssessmentItemTest.java
```diff
@@ -8,11 +8,11 @@
 /**
```
malaffoon: Just renaming ...

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/model/WarehouseAssessmentTest.java
```diff
@@ -5,7 +5,7 @@
 /**
```
malaffoon: Just renaming ...

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcWarehouseAssessmentPackageRepositoryIT.java
```diff
@@ -5,9 +5,9 @@
 import org.junit.runner.RunWith;
```
malaffoon: Just renaming ...

package-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/service/impl/AssessmentParserIT.java
```diff
@@ -44,36 +60,51 @@ public void itShouldCatchAllErrorsAndReturnNoObjectsWhenParsingBadData() throws
 
         assertThat(assessmentParser.parse(toByteArray(this.getClass().getResourceAsStream("/IAB_ICA_Combined.bad-asmt.csv")), elementErrorCollector)).isEmpty();
         assertThat(elementErrorCollector.toJson()).isEqualTo("{\"messages\":[" +
-                "{\"elementName\":\"AssessmentId\",\"value\":\"\",\"error\":\"value may not be blank\"}," +
```
malaffoon: A lot of these error tests are now handled by the common parser, which has its own tests.

If you spend a few minutes thinking about it you may challenge that the feedback to the user will be less useful now: before, it would gather more errors and return them; now, basic "structural" errors will abort the validation process returning a single error. That means packages that are not well-formed may take more submissions to discover all the errors. So, yeah, that's true. But they will have the ability to check well-formedness using a separate utility.

PR #586, opened 2019-05-07
RP-321, reduce tenancy chain when archiving import payloads

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -212,4 +217,25 @@ public long resubmitImports(final RdwImportQuery query) {
...
+     * @param tenancyChain the full user tenancy chain
+     * @return a tenancy chain string with just import-related grants
+     */
+    static String reducedImportTenancyChainString(final TenancyChain tenancyChain) {
+        final List<String> grantStrings = newArrayList();
+        for (final Grant grant : tenancyChain.getGrants()) {
+            if (grant.getRole().equals(Authorities.DataLoadRole)
+             || grant.getRole().equals(Authorities.GroupAdminRole)) {
+                grantStrings.add(grant.toString());
+            }
```
ftdholbrook: should this logic be pushed up into TenancyChain?  I will probably have to have some similar business logic related to mult-tenancy and was going to add a rdwOnly() method to it.  Something like tenancyChain.importOnly().

malaffoon: Probably, but for the hotfix i didn't want to change RDW_Common.

PR #587, opened 2019-05-07
RP-283: fixes for Integer cast error. Affects tests only.

gregcharles:
I think this will fix the build issue. The ExamProcessorIT and ExamProcessorAuditIT  tests pass in TeamCity. I can't figure how, but they do.


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamineeProcessor.java
```diff
@@ -88,7 +88,15 @@
 
     @Override
     public int parseSchoolId(final Examinee examinee) throws ImportException {
-        return schoolRepository.findIdByNaturalId(getBestRelationshipValue(examinee, "SchoolId"));
+        final String naturalId = getBestRelationshipValue(examinee, "SchoolId");
+        Integer id = schoolRepository.findIdByNaturalId(naturalId);
```
malaffoon: nit: `final`

PR #588, opened 2019-05-09
RP-314: implement sandbox permissions for use pipeline scripts

gregcharles:


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorTransformationTest.java
```diff
@@ -112,7 +112,11 @@ private PipelineFactory setupPipelineFactory() throws Exception {
         ConfigurationCompiler configurationCompiler = new DefaultConfigurationCompiler();
         PropertyResolver propertyResolver = new DefaultPropertyResolver();
 
-        return config.pipelineFactory(scriptSource, scriptCompiler, configurationCompiler, propertyResolver);
+        return config.pipelineFactory(scriptSource,
```
esotrope: arg list is getting long and has optional props it looks like - should we have a "context" object we pass in instead?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -51,4 +76,68 @@ public PipelineResults run(final Object input)  {
 
         return new PipelineResults(lastValidTransformation);
     }
+
+    private Object runScript(final Script script, final ResourceType type) {
+        if (type == ResourceType.SYSTEM_NODE || sandboxSecurityManager == null) {
+            return script.run();
+        } else {
```
esotrope: this else isn't needed because the inside returns

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -51,4 +76,68 @@ public PipelineResults run(final Object input)  {
...
+
+            throw new RuntimeException(message, cause);
+        }
+    }
+
+    // Digs out script line number if possible
+    private Optional<Object> getScriptLineNumber(final Throwable t) {
+        Map<String,Object> map = ScriptUtils.extractScriptInfo(t);
+
+        if (map.get("lineNumber") != null) {
```
esotrope: why no hard type for the shape of "script info"?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -51,4 +76,68 @@ public PipelineResults run(final Object input)  {
...
+            return Optional.of(map.get("lineNumber"));
+        }
+
+        return Optional.empty();
```
esotrope: can this all be simplified with `Optional.ofNullable(map.get("lineNumber"))` ?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class DefaultSandboxPermissions implements SandboxPermissions {
+    private final PermissionCollection whitelist;
+    private final PermissionCollection blacklist;
+
+    public DefaultSandboxPermissions() {
+
+        // Define whitelist. Note: add new AllPermission() to make everything permitted except what
+        // is on the blacklist.
+        this.whitelist = new Permissions();
```
esotrope: +1 for whitelist

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,62 @@
...
+        // This one is probably only used by JUnit tests in dev to load certain classes from incremental builds.
+        // However, it should be safe to include everywhere.
+        whitelist.add(new FilePermission(System.getProperty("user.home") + "/.gradle/caches/-", "read"));
+
+        // Define blacklist. Note: leave this empty to allow everything permitted by the whitelist.
+        this.blacklist = new Permissions();
+    }
+
+    @Override
+    public void checkPermission(final Permission perm) throws AccessControlException {
```
esotrope: permissions are dangerous but perms are guaranteed damage : P

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/util/ScriptUtils.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        for (StackTraceElement element: stackTrace) {
+            if (StringUtils.endsWith(element.getFileName(), ".groovy")) {
+                final Map<String,Object> map = new HashMap<>();
+
+                map.put("className", element.getClassName());
+                map.put("methodName", element.getMethodName());
+                map.put("fileName", element.getFileName());
+                map.put("lineNumber", element.getLineNumber());
```
esotrope: may be expedient to use a map here but a typed class would help devs discover what fields are available easier

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -13,26 +23,41 @@
 public class Pipeline {
     private static final Logger logger = LoggerFactory.getLogger(Pipeline.class);
 
+    private static final int SANDBOX_RETRIES = 2;
+    private static final int SANDBOX_THREADS = 2;
+    private static final ExecutorService THREAD_POOL = Executors.newFixedThreadPool(SANDBOX_THREADS);
```
malaffoon: Can we make all this property-based?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
malaffoon: 👍  i would inline this as `return Optional.ofNullable(ScriptUtils.extractScriptInfo(t).get("lineNumber"));`

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
```diff
@@ -51,4 +76,68 @@ public PipelineResults run(final Object input)  {
...
+        } catch (InterruptedException e) {
+            if (retries > 0) {
+                return runInSandbox(script, retries - 1);
+            } else {
+                throw new RuntimeException("Too many interrupted exceptions", e);
+            }
+        } catch (ExecutionException e) {
+            // Unwrap the cause and rethrow.
+            Throwable cause = e.getCause();
+            if (cause == null) {
```
malaffoon: Can't the ExecutionException itself have a message?

Style thing probably but i think this would be more readable if there was a helper to extract the "bestCause". Use that and then go straight to line 119.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/util/ScriptUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Extract information about where a script error occurred by digging it out of the stack trace.
+     *
+     * @param stackTrace array of StackTraceElements such as can be retrieved from a Throwable or a Thread.
+     *
+     * @return a map of the script info if it can be found in the stack trace, otherwise an empty map.
+     */
+    @Nonnull
+    public static Map<String,Object> extractScriptInfo(final StackTraceElement [] stackTrace) {
```
malaffoon: Wow, if we have to resort to this i have to wonder if there is a better way?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/util/ScriptUtils.java
malaffoon: 👍

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,62 @@
...
+        if (whitelist == null || !whitelist.implies(perm)) {
```
malaffoon: seems like these can never be null

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,62 @@
...
+        whitelist.add(new RuntimePermission("createClassLoader"));
+        whitelist.add(new RuntimePermission("getProtectionDomain"));
+        whitelist.add(new SecurityPermission("getProperty.*"));
+        whitelist.add(new FilePermission(System.getProperty("java.home") + "/lib/-", "read"));
+
+        // This one is probably only used by JUnit tests in dev to load certain classes from incremental builds.
+        // However, it should be safe to include everywhere.
+        whitelist.add(new FilePermission(System.getProperty("user.home") + "/.gradle/caches/-", "read"));
+
+        // Define blacklist. Note: leave this empty to allow everything permitted by the whitelist.
```
malaffoon: Is this comment correct?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DisableSecurityManagerPermission.java
```diff
@@ -0,0 +1,36 @@
+package org.opentestsystem.rdw.ingest.script.security;
+
+import java.security.Permission;
+
+/**
+ * Special permission used to disable the SandboxSecurityManager. There is no current use case for this since
+ * sandbox security will only be enabled for code that shouldn't be able to disable it. However, a use case
+ * may appear in the future.
```
malaffoon: Not sure i grok this ... isn't it being used just a couple files down?

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/SandboxTest.java
```diff
@@ -0,0 +1,185 @@
...
+        pool.submit(callable).get();
+    }
+
+    @Test
+    public void itShouldRunSafeScriptInSandbox() throws Exception {
+        final String scriptCode =
+                "enable 'xml'" + LF +
+                "transform '//chapter' by {" + LF +
+                "    it.text = it.text + '!'" + LF +
+                "}" + LF +
```
malaffoon: Why does this one use `LF` but most of the other tests use `\n`? I'd favor `\n` if my opinion matters.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,62 @@
+package org.opentestsystem.rdw.ingest.script.security;
```
malaffoon: I know `sandbox` makes perfect sense but i'm worried about the conflict with "sandbox" initiative. Perhaps `DefaultUserScriptPermissions`?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/SandboxSecurityManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.security.Permission;
+
+/**
+ * Security manager that be enabled or disabled by thread to either check permission against a given set of sandbox
+ * permissions or not check them at all, which would be the default if no security manager is set on the system.
+ * This is more performant than assigning AllPermissions to the application domain and sandbox permissions
+ * to the untrusted domain because there is a big performance penalty for checking permissions at all.
+ */
+public class SandboxSecurityManager extends SecurityManager {
```
malaffoon: This is embarrassing but i just realized that you don't have to explicitly import `java.lang` things. I guess i knew that but didn't notice normally because of the IDE.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
esotrope: security "cage" perhaps?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
gregcharles: Fair point. I just haven't worked out what the return object should look like yet. I'll flesh this out later when dealing with the error reporting interface.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
gregcharles: You're right they work the same. I originally was check map for null too, but then I annotated the extractScript info as "NotNull". I didn't notice this extra simplification.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
gregcharles: Hmm, +1 you like the name or like the idea of having a whitelist?

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
gregcharles: I'm changing it to permission just in case that's what you meant. I generally don't like abbreviations either, but I got so used to seeing this one in the Java internal code.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/util/ScriptUtils.java
gregcharles: Agreed, but I intend this to be something of a proof of concept. (i.e., that I can get the script line number at all.) At some point, I'll integrate it into a script exception type or something like that.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
gregcharles: Yes, we could, but is it worth it? Is this something that would be useful to tune? We'd have to document it, and we'd have to validate the input. (What if they said threads should be 0, or -1?) We'd also need to do more testing to see if different number combinations would cause trouble, although I don't think they would.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/Pipeline.java
gregcharles: Yes, but it just says that an error was caught during execution. It's possible for ExecutionException to not wrap a "cause" exception, but I couldn't find a case where this happened. I like the idea of breaking out a best cause method. However, I think I'll also defer this to when I deal with error handling and reporting.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
esotrope: idea : j

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/util/ScriptUtils.java
gregcharles: Yes, I wondered too, and did spend some time looking for a better way. I thought about doing something clever like catching script exceptions somewhere in the base class and wrapping them up in a GroovyRuntimeException. I'd still need to populate the line number there though, and the only way to get that information would still involve digging it out of the stack trace. Also the best place to catch this would be in the Script.run() method, but that's abstract and implemented by the compiler. I'd need to introduce a PipelineScript.otherNameForRun():
`` try {
        return run();
    } catch (Throwable t) { 
        <digout line>
        throw new GroovyRuntimeException(<with line number and root cause>);
    }`` 

Then, I'd have to change the pipeline and tests to use otherNameForRun instead of run.
I still might do that!

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
gregcharles: I think so. As far as I can tell, we only load classes out of the .gradle/caches during JUnit tests. However, I'm not _sure_ it's impossible in other contexts. In any case, it doesn't seem dangerous to allow reading from .gradle/cache.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DisableSecurityManagerPermission.java
gregcharles: No, it's never used. I'm not sure where you mean. 

I was conflicted about whether to keep this class at all. I adapted this from an example that gave permission to code from a particular jar the ability to enable or disable the sandbox permissions. You'd need the path and exact name of the jar though, so that's not so useful in our case. Pretty much we want the sandbox permissions enabled in the worker threads only and never disabled. It would be OK to grant this permission to all other contexts, but not really necessary. Also I'm not quite sure how to do that.

script-common/src/test/java/org/opentestsystem/rdw/ingest/script/SandboxTest.java
gregcharles: OK. :) I sort of like LF, but there's no real reason for this to be OS independent. It's not like the script compiler cares.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/DefaultSandboxPermissions.java
gregcharles: Yeah, it's an overloaded term in this project. However, it's also the standard term for this kind of functionality. There are four classes using Sandbox in their names, not just this one. Maybe stick a pin in this for now? Later I'll change them all to UserScript or Cage or whatever we come up with.

script-common/src/main/java/org/opentestsystem/rdw/ingest/script/security/SandboxSecurityManager.java
gregcharles: Yes, that's true. Did I import java.lang somewhere? I'm not seeing it.

Groovy, btw, implicitly imports a bunch of Java and Groovy packages, like java.io, java.util, and some others. It makes writing scripts much easier. You can do a lot of stuff without dealing with imports at all.

malaffoon: Sorry, i wasn't clear. I was staring at `SecurityManager` trying to figure out if we had our own custom class or what. I looked for the import for it and it wasn't there. After scratching my head for a minute i realized it must be from java.lang.

gregcharles:
Updated several issues and documented others in a tech debt ticket: RP-326.

PR #590, opened 2019-05-13
reduce the tenancy chain that is stored with archived payloads

malaffoon:
Since the customer chose not to do the hotfix for this change, i've reworked it a tiny bit for the main develop branch ...


import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -209,6 +212,22 @@ public long resubmitImports(final RdwImportQuery query) {
...
+     * amount of user metadata that can be stored. And tenancy chains can be very large since they
+     * includes roles for services unrelated to RDW import. This method extracts just those grants
+     * that are related to RDW import.
+     *
+     * @param tenancyChain the full user tenancy chain
+     * @return a tenancy chain string with just import-related grants
+     */
+    static String reducedImportTenancyChainString(final TenancyChain tenancyChain) {
+        return tenancyChain.filter(grant ->
+                grant.getRole().equals(Authorities.DataLoadRole) || grant.getRole().equals(Authorities.GroupAdminRole)
```
malaffoon: @ftdholbrook i've left the role-awareness here while moving the guts of filtering into TenancyChain.

PR #593, opened 2019-05-20
switch import service to use tenant aware data source

malaffoon:
I'm trying to do this in byte-sized pieces.


common/src/test/java/org/opentestsystem/rdw/ingest/common/repository/CommonTestConfiguration.java
```diff
@@ -19,6 +20,9 @@
 public class CommonTestConfiguration {
 
     @Bean
+    // We need a data source for the repository IT's of course. But other modules bring in
+    // rdw-ingest-common:tests so we need to make this conditional on the common-test profile.
+    @Profile("common-test")
```
malaffoon: This data source was leaking into the other modules so make it load only if it is a common-test.

import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
```diff
@@ -11,6 +11,7 @@
         ElastiCacheAutoConfiguration.class,
         MailSenderAutoConfiguration.class,
         ContextStackAutoConfiguration.class
+//        DataSourceAutoConfiguration.class
```
malaffoon: oops, will remove

import-service/src/main/java/org/opentestsystem/rdw/ingest/ImportServiceApplication.java
```diff
@@ -19,5 +20,3 @@ public static void main(final String[] args) {
         SpringApplication.run(ImportServiceApplication.class, args);
     }
 }
-
-
```
malaffoon: I donated these blank lines to Father Joe's.

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/DefaultAuthorityService.java
```diff
@@ -46,6 +45,7 @@
     private final String client;
     private final String state;
 
+    // TODO - multitenant
```
malaffoon: i've put these TODO's in a couple places i noticed

import-service/src/main/java/org/opentestsystem/rdw/ingest/auth/SbacTokenConverter.java
```diff
@@ -60,14 +63,27 @@ public OAuth2Authentication extractAuthentication(final Map<String, ?> map) {
...
             authorities.addAll(authorityService.getAuthorities(tenancyChain));
 
+            // the tenantId is the first state that has an ingest role
+            // roles have already been filtered so just find the first state
+            final String tenantId = tenancyChain.getGrants().stream()
+                    .map(Grant::getStateId)
+                    .filter(Objects::nonNull)
+                    .findFirst()
+                    .orElse(TenantIdResolver.NO_TENANT_ID)
+                    .toUpperCase();
```
malaffoon: Dave, i've realized that tenancy chains for import users are (currently) very likely to not have state id's in them. Obviously, they'll have to change the credentials but we might want to discuss a good behavior for the app so that it is obvious to DevOps.

import-service/src/main/java/org/opentestsystem/rdw/ingest/multitenant/ImportTenantIdResolver.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * The import {@link TenantIdResolver} gets the tenant id from current user.
+ * TODO - multitenant; this doesn't really deal with actuator and status endpoints
+ */
+public class ImportTenantIdResolver implements TenantIdResolver {
+
+    @Override
+    public Optional<String> getTenantId() {
+        try {
+            return Optional.of(getCurrentUser().getTenantId());
```
malaffoon: Unlike reporting, this checks *just* the current user, there isn't a separate thread context. Dave, that okay? Am i missing something?

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -59,16 +57,15 @@
     }
 
     @Override
-    public RdwImport importContent(final SbacUserDetails user,
-                                   final byte[] payload,
```
malaffoon: This set of changes removes the user from being passed down the call chain ... the guy at the bottom can just use the current user from the thread context.

import-service/src/main/java/org/opentestsystem/rdw/ingest/service/DefaultImportService.java
```diff
@@ -212,22 +209,6 @@ public long resubmitImports(final RdwImportQuery query) {
...
-     * that are related to RDW import.
-     *
-     * @param tenancyChain the full user tenancy chain
-     * @return a tenancy chain string with just import-related grants
-     */
-    static String reducedImportTenancyChainString(final TenancyChain tenancyChain) {
-        return tenancyChain.filter(grant ->
-                grant.getRole().equals(Authorities.DataLoadRole) || grant.getRole().equals(Authorities.GroupAdminRole)
-        ).toString();
-    }
```
malaffoon: This isn't needed here anymore: we reduce the tenancy chain when we create the current user.

import-service/src/main/java/org/opentestsystem/rdw/ingest/multitenant/ImportTenantIdResolver.java
ftdholbrook: On the reporting side that works for web requests, but not for message based requests.  In message based requests we don't have a security context, so it is hand rolled.

malaffoon: Ah, so this will become an issue in the other services. Got it.

import-service/src/main/java/org/opentestsystem/rdw/ingest/multitenant/ImportTenantIdResolver.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Helper to get the current {@link SbacUser} from the security context.
+     *
+     * @return current SbacUser
+     * @throws IllegalStateException if there is no current auth
+     * @throws ClassCastException if current principal is not an SbacUser
+     */
+    public static SbacUser getCurrentUser() {
```
ftdholbrook: Seems a bit off that get user is inside the tenant id resolver as a public method?  I think there is a static helper for that over on the reporting side.

malaffoon: I admit to being a bit lazy until i get to sharing this with the other services.

PR #595, opened 2019-05-22
add multi-tenant datasource to migrate-olap, migrate-reporting, task-service

malaffoon:
More of the same. This is the last of the datasource changes. Now onto the hard stuff.

### DIFF
--- a/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
+++ b/exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
@@ -44,9 +44,9 @@ class JdbcStudentGroupRepository implements StudentGroupRepository {
     private String sqlExistsStudentInGroup;
 
     @Autowired
-    JdbcStudentGroupRepository(final NamedParameterJdbcTemplate namedParameterJdbcTemplate, final JdbcTemplate jdbcTemplate) {
+    JdbcStudentGroupRepository(final JdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
-        this.namedParameterJdbcTemplate = namedParameterJdbcTemplate;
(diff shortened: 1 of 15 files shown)


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/repository/impl/JdbcStudentGroupRepository.java
```diff
@@ -44,9 +44,9 @@
     private String sqlExistsStudentInGroup;
 
     @Autowired
-    JdbcStudentGroupRepository(final NamedParameterJdbcTemplate namedParameterJdbcTemplate, final JdbcTemplate jdbcTemplate) {
+    JdbcStudentGroupRepository(final JdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
-        this.namedParameterJdbcTemplate = namedParameterJdbcTemplate;
+        this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(jdbcTemplate);
```
malaffoon: I noticed a few places (in exam-processor) where we autowire both kinds of templates; i changed to derive one from the other as appropriate.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/repository/DataSourceConfiguration.java
```diff
@@ -47,56 +76,63 @@ public BatchConfigurer configurer(@Qualifier("migrateDataSource") final DataSour
...
+        return new DataSourceElementsProperties();
+    }
+
     @Bean
-    @ConfigurationProperties(prefix = "datasources.olap_rw")
-    public DataSource olapDataSource() {
-        return DataSourceBuilder
-                .create()
-                .build();
+    @RefreshScope
```
ftdholbrook: We need to do a full Config class audit and make sure all the correct @RefreshScope annotations are in place... I think they are easy to miss, and the application will run fine at startup, so very difficult to test.

PR #596, opened 2019-05-23
RP-386: save pipeline transformed document alongside original

gregcharles:
Notes: 
I'm saving transformed document with original MD5 digest plus extension "-transformed".  
This implementation requires each ingest processor that uses it to import the archive service. I couldn't think of a more clever way to do it.


common/src/main/java/org/opentestsystem/rdw/ingest/common/util/ByteArrayContentWriter.java
```diff
@@ -0,0 +1,11 @@
+package org.opentestsystem.rdw.ingest.common.util;
+
+import java.util.Properties;
+
+/**
+ * Functional interface used to expose the archive service's writeResource method as a lambda.
+ */
+@FunctionalInterface
```
esotrope: never used this before. interesting

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiver.java
```diff
@@ -0,0 +1,55 @@
...
+    protected static final String CONTENT_TYPE_HEADER = "originalContentType";
+    protected static final String SUFFIX = "-transformed";
+
+    public void archive(final RdwMessageHeaderAccessor accessor,
+                        final byte[] originalPayload,
+                        final byte[] transformedPayload,
+                        final ByteArrayContentWriter archiveService) {
+
+        if (!Arrays.equals(originalPayload, transformedPayload)) {
+            final String digest = DigestUtils.md5DigestAsHex(originalPayload).toUpperCase();
```
malaffoon: There is a lot of repeated calculation in here that i think we should avoid. The import record has the digest and the content type already. It feels better to just pull the import record if necessary. We should discuss because i might be missing something.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiver.java
```diff
@@ -0,0 +1,55 @@
...
+            final ImportContent content = ImportContent.valueOf(accessor.getContent());
+
+            final String location =
+                    new LocationStrategy.ImportContentLocationStrategy(content).location(digest) + SUFFIX;
```
malaffoon: We should really figure out how to share this; i know the current code is lazy and just repeats things.

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiverTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.Mockito.when;
+import static org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver.CONTENT_LENGTH_PROPERTY;
+import static org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver.CONTENT_TYPE_HEADER;
+import static org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver.CONTENT_TYPE_PROPERTY;
+import static org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver.SUFFIX;
+
+public class TransformedPayloadArchiverTest {
+
+    public static final String CONTENT = "EXAM";
```
malaffoon: nit: should probably be derived from ImportContent.EXAM

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiver.java
```diff
@@ -0,0 +1,55 @@
...
+    protected static final String CONTENT_LENGTH_PROPERTY = "Content-Length";
+    protected static final String CONTENT_TYPE_HEADER = "originalContentType";
+    protected static final String SUFFIX = "-transformed";
+
+    public void archive(final RdwMessageHeaderAccessor accessor,
+                        final byte[] originalPayload,
+                        final byte[] transformedPayload,
+                        final ByteArrayContentWriter archiveService) {
+
+        if (!Arrays.equals(originalPayload, transformedPayload)) {
```
malaffoon: Could this be returned as part of PipelineResults instead of calculating here?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -56,9 +61,11 @@ public void process(final Message<?> message) {
             if (pipeline == null) {
                 report = tdsReportFromXml(new ByteArrayInputStream(payload));
             } else {
-                final PipelineResults results = pipeline.run(payload);
-                logger.debug("Transformed TRT to: \n{}", results.asString());
-                report = tdsReportFromXml(new ByteArrayInputStream(results.asByteArray()));
+                // Run through pipeline and archive the transformation if any
+                final byte [] transformedPayload = pipeline.run(payload).asByteArray();
```
malaffoon: my comment about returning a "modified" flag with the pipeline results probably belongs here

malaffoon:
I think the only way to avoid each processor knowing about the archive service would be to make the pipeline a separate message handler. So import puts the original payload in the queue, the pipeline pulls it, does the transformation and put the transformed payload in the queue, and then the processor gets the transformed payload. If we had more time, it might be worth playing with that but i think it would be a bad approach if we couldn't optimize away the double message handling when the pipeline is not active.

tl;dr - i think this is fine for now. We should make sure the archive service configuration is as shared as possible.

common/src/main/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiver.java
```diff
@@ -0,0 +1,55 @@
...
+import org.opentestsystem.rdw.messaging.RdwMessageHeaderAccessor;
+
+/**
+ * Archiver for the transformed payload created during the pre-processor phase of one of the ingest processors.
+ * The original payload is already archived as part of the import service, and this service will save the transformed
+ * under the same name but with "-transform" added to it.
+ *
+ * If there is no transformation of the payload during pre-processing then no transformed file will be saved.
+ */
+public class TransformedPayloadArchiver {
```
malaffoon: I'm not sure i see the advantage of making this a class, isn't it just a utility method at this point?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -43,44 +55,55 @@
...
 
     @ServiceActivator(inputChannel = Sink.INPUT)
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         final byte[] payload = (byte[]) message.getPayload();
         final long importId = accessor.getImportId();
+        final TransformedPayloadArchiver archiver = new TransformedPayloadArchiver();
 
         // set tenant context from message
+        // TODO: breaks model of handling all exceptions internally.
```
malaffoon: But it was on purpose: this is a badly formed message. We should discuss what we want to do with these?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/ArchiveServiceConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.env.Environment;
+
+import org.opentestsystem.rdw.archive.ArchiveConfiguration;
+import org.opentestsystem.rdw.archive.ArchiveService;
+import org.opentestsystem.rdw.archive.ArchiveServiceFactory;
+
+@Configuration
+@Import(ArchiveConfiguration.class)
```
malaffoon: This is deprecated; there is an archiveService in `ArchiveServiceConfiguration`.

exam-processor/src/main/resources/application.yml
```diff
@@ -9,6 +9,23 @@ server:
...
+      static: us-west-2
+    # disable CloudFormation stuff
+    stack:
+      auto: false
+    credentials:
+      instance-profile: false
+
+# use LocalArchiveService by default
+archive:
+  root: file:///tmp/
```
malaffoon: This has changed to use the new archive service config (i.e. `archive.uri-root`)

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
malaffoon: @gregcharles, can you replace with something like this:
```
        final String tenantId = accessor.getTenantId();
        if (tenantId == null) {
            importRepository.updateStatusAndMessageById(importId, ImportStatus.INVALID, "no tenant id found in message header");
            return;
        }
```
Or i can give myself a TODO for it later.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -43,44 +55,55 @@
...
         this.pipelineFactory = pipelineFactory;
+        this.archiveService = archiveService;
     }
 
     @ServiceActivator(inputChannel = Sink.INPUT)
     public void process(final Message<?> message) {
         final RdwMessageHeaderAccessor accessor = RdwMessageHeaderAccessor.wrap(message);
         final byte[] payload = (byte[]) message.getPayload();
         final long importId = accessor.getImportId();
+        final TransformedPayloadArchiver archiver = new TransformedPayloadArchiver();
```
malaffoon: Sorry to nag but i really don't understand instantiating an object to execute a utility method. Either move the archiveService into the archiver and inject the archiver, or ... just call the method. What am i missing?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
gregcharles: But that's what TODO is for! Yes, I saw you already made a change to this anyway, but we should discuss it. With the tenant-aware changes, it becomes even more difficult to determine if the pipeline will fail before reading the message off the queue. I like the idea of logging a failure and stopping the processing, as you did with the no tenant ID error. Maybe a new BAD_PIPELINE error status?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
gregcharles: It took me a while to remember, but I did that so I wouldn't have to have a dependency on rdw-common-archive in rdw-ingest-common. That is, TransformedPayloadArchiver is in rdw-ingest-common and to autowire ArchiveService into it, I'd need to resolve that class. I think your "or" is saying you'd be happy with a static method, so I'll try that.

PR #602, opened 2019-05-25
deal with getting tenant from message in exam, group, package processors

malaffoon:
I am also trying to make all the tests run without the no-tenant exception showing up


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -48,10 +52,14 @@ public void process(final Message<?> message) {
         final byte[] payload = (byte[]) message.getPayload();
         final long importId = accessor.getImportId();
 
-        logger.debug("received {} import {}", accessor.getContent(), importId);
+        // set tenant context from message
+        final String tenantId = Optional.ofNullable(accessor.getTenantId()).orElseThrow(() ->
+                new IllegalArgumentException("no tenant id in " + accessor.getContent() + " message for import " + importId));
+        TenantContextHolder.setTenantId(tenantId);
```
malaffoon: In all three of these services, there is a single entry point for doing work. So, instead of using a global message interceptor i'm doing this directly. Thoughts @ftdholbrook?

exam-processor/src/test/resources/application-test.yml
```diff
@@ -1,3 +1,9 @@
 datasources:
   warehouse_rw:
     url-schema: warehouse_test
+
+tenant-properties:
+  tenants:
+    - id: CA
+      key: CA
+      name: California
```
malaffoon: oops, i have to change this

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/repository/impl/TenantAwareIT.java
```diff
@@ -0,0 +1,39 @@
...
+        YamlPropertiesConfigurator.class
+})
+@CachingTest
+@Transactional
+@ActiveProfiles("test")
+public abstract class TenantAwareIT extends TenantCacheableIT {
+
+    @Before
+    public void setTenantContext() {
+        TenantContextHolder.setTenantId("CA");
```
malaffoon: the point of all the IT changes: make sure a tenant is set during the tests

malaffoon:
Going to commit to fix build (crossing fingers) ... will address comments.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
ftdholbrook: If there are only 3 that makes sense

PR #604, opened 2019-05-26
make migrate job multi-tenant aware

malaffoon:


common/src/main/java/org/opentestsystem/rdw/ingest/common/multitenant/TenantAwareCachingPipelineFactory.java
```diff
@@ -13,7 +13,7 @@
 public class TenantAwareCachingPipelineFactory implements PipelineFactory {
 
     private final PipelineFactory wrappedFactory;
-    private final Map<String, Pipeline> pipelines = new HashMap<>();
+    private final Map<String, Pipeline> pipelines = new ConcurrentHashMap<>();
```
malaffoon: From previous PR.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
```diff
@@ -24,31 +30,45 @@
 
     private final MigrateLifecycle jobRunner;
     private final MigrateRepository repository;
+    private final TenantProperties tenantProperties;
 
-    public MigrateEndpoint(final MigrateLifecycle jobRunner, final MigrateRepository repository) {
+    @Autowired
+    // TODO - does this need RefreshScope since tenant properties has it?
```
malaffoon: @ftdholbrook, do you know for sure how this works? Do i have to put RefreshScope on every bean that uses a bean that has RefreshScope, or does the framework figure that out?

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
```diff
@@ -24,31 +30,45 @@
...
-            builder.append(". Last migrate: ").append(lastMigrate.getStatus())
-                   .append(" to ").append(asOffsetDateTime(lastMigrate.getLastAt()));
-
-            if (lastMigrate.getStatus() != MigrateStatus.COMPLETED) {
-                final Instant lastMigratedAt = repository.findLastMigratedAt();
-                if (lastMigratedAt != null) {
-                    builder.append(". Last successful migrate to ").append(asOffsetDateTime(lastMigratedAt));
+        builder.append(jobRunner.isRunning() ? "running" : "paused").append("\n");
+
+        // get status of each tenant by setting thread context and asking job runner for status
```
malaffoon: Anything wrong with this approach? Basically, in the thread that is fulfilling the request, i just cycle through all the tenants registered in the system, setting the thread context before asking the tenant-aware services for info.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
```diff
@@ -83,23 +110,34 @@ public boolean migrate() {
...
+        logger.info("Marking migrate {} as ABANDONED for tenant {}", migrate.getId(), tenantId);
         repository.updateStatusById(migrate.getId(), MigrateStatus.ABANDONED);
         jobRunner.enable(true);
         return true;
     }
+
+    private void checkAndSetTenant(final String tenantId) {
+        final TenantProperties.Tenant tenant = tenantProperties.findTenantById(tenantId).orElseThrow(() ->
+                new IllegalArgumentException("unknown tenant id " + tenantId + " (they are case sensitive)")
+        );
```
malaffoon: We don't have exception handles in place for the actuator end-points so this results in a 500 error ... but the info is correct, e.g.:
```
{
  "timestamp": "2019-05-26T22:41:53.630+0000",
  "status": 500,
  "error": "Internal Server Error",
  "exception": "java.lang.IllegalArgumentException",
  "message": "unknown tenant id XX (they are case sensitive)",
  "path": "/migrate/enable"
}
```

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -1,25 +1,35 @@
...
  * It skips job execution while paused or disabled. It is paused using Spring's Lifecycle
  * start/stop (which can be set using the pause/resume actuator end-points). It is disabled
  * if there are any unresolved migrations in the database. An unresolved migration is any
  * that has a status of FAILED or STARTED. To enable, any such migrate records should be set
  * to ABANDONED with a message.
+ * </p>
+ * <p>
+ * The job runner is tenant aware wrt to the enabled flag; it requires a tenant id resolver
+ * and stores disabled state by tenant.
+ * TODO - multitenant; that breaks the MigrateLifecycle interface enable/isEnabled
```
malaffoon: i wouldn't mind a brainstorming chat for this ... although i have come up with a solution that isn't wrong.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobRunner.java
```diff
@@ -28,15 +38,18 @@
     private final JobLauncher jobLauncher;
     private final Job job;
     private boolean paused = false;
-    private boolean enabled = true;
+    private final TenantIdResolver tenantIdResolver;
+    private final Map<String, Boolean> tenantDisabled = new ConcurrentHashMap<>();
```
malaffoon: I guess i could've hidden this behind a TenantAwareDisabledJobResolver or something but ... well, i didn't.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
ftdholbrook: It works by wrapping the class marked by RefreshScope in a proxy.  So mostly, no you don't have to touch every consumer.  Some usages (like the dynamic datasources) cache the results, and don't ever check them again, so in that case I had to include the datasource router as well as a config could have changed under the hood.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
ftdholbrook: That looks correct to me.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -14,32 +11,47 @@
...
     @Scheduled(fixedDelayString = "${migrate.batch.delay}")
     @Override
     public void run() {
-        super.run();
+        // run the migration for every known tenant
+        for (final TenantProperties.Tenant tenant : tenantProperties.getTenants().values()) {
+            TenantContextHolder.setTenantId(tenant.getId());
+            super.run();
+        }
+        TenantContextHolder.clear();
```
malaffoon: The final question about this refactor: do i put this tenant-looping code in the base class? It is the same for both migrate services right now. But i could imagine in the future they handle threading differently.

Hmm, maybe i'll pull it up to the base class.

migrate-reporting/src/main/resources/application.yml
```diff
@@ -98,3 +98,14 @@ logging:
     org.springframework.transaction: INFO
     org.springframework.jdbc.core.JdbcTemplate: INFO
     org.springframework.jdbc.datasource.DataSourceTransactionManager: INFO
+
+tenantProperties:
```
malaffoon: I'll revert this, i was just testing running locally.

ftdholbrook:
This looks pretty solid to me, if you want to talk it through tomorrow I think it is probably worth the time.  There are a bunch of moving parts here.

migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateEndpoint.java
malaffoon: I fixed this so it doesn't do that any more.

PR #606, opened 2019-05-27
get tenant id into migrate job logging

malaffoon:


migrate-common/src/main/java/org/opentestsystem/rdw/migrate/common/MigrateJobExecutionListener.java
```diff
@@ -37,7 +40,8 @@ public void beforeJob(final JobExecution jobExecution) {
                         .size(jobParameters.getBatchSize())
                         .build()));
 
-        logger.info("Migrate Job id " + jobExecution.getJobId() + "] is launched]");
+        logger.info("Migrate Job id {} for tenant {} is launched",
+                jobExecution.getJobId(), tenantIdResolver.getTenantId().orElse("??"));
```
malaffoon: The `.orElse("??")` will never be hit because this listener is only invoked when a migrate job is run which can only happen if there is a tenant id in the thread context.

PR #608, opened 2019-05-28
get ingest services working without failing due to multi-tenant issues

malaffoon:
With these changes, all the ingest services run without complaining about no-tenant.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
     environment:
       - CONFIG_SERVICE_REPO
       - ENCRYPT_KEY
+      - CONFIG_SERVICE_SEARCH_PATH=tenant*
       - GIT_USER
       - GIT_PASSWORD
       - SPRING_RABBITMQ_ADDRESSES=rabbitmq:5672
@@ -34,6 +35,7 @@ services:
       test: ["CMD", "curl", "-f", "http://localhost:8888/health"]
       interval: 10s
       timeout: 2s
+      retries: 6
 
   import-service:
     image: smarterbalanced/rdw-ingest-import-service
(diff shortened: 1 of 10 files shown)


docker-compose.yml
```diff
@@ -34,6 +35,7 @@ services:
       test: ["CMD", "curl", "-f", "http://localhost:8888/health"]
       interval: 10s
       timeout: 2s
+      retries: 6
```
malaffoon: Not sure if it is the new version but i was consistently getting startup failures because it wasn't healthy soon enough. This just gives it a little more time.

docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
     environment:
       - CONFIG_SERVICE_REPO
       - ENCRYPT_KEY
+      - CONFIG_SERVICE_SEARCH_PATH=tenant*
```
malaffoon: Not sure if this is necessary with the latest version but it serves as documentation.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultExamItemProcessor.java
```diff
@@ -1,53 +1,50 @@
...
+                             final ItemTraitRepository itemTraitRepository,
                              final ResponseTypeRepository responseTypeRepository) {
-        this.responseTypeRepository = responseTypeRepository;
         this.scoreConfiguration = scoreConfiguration;
-        itemTraitIds = newHashMap();
-        for (final ItemTrait trait : repository.findAll()) {
-            itemTraitIds.put(trait.getDimension(), trait.getId());
-        }
+        this.itemTraitRepository = itemTraitRepository;
+        this.responseTypeRepository = responseTypeRepository;
```
malaffoon: Change this to be queried just in time with caching at the repo level.

PR #614, opened 2019-06-03
fix the RdwImportRepositoryIT; oh yeah, and move everything to a new package

malaffoon:
This PR isn't nearly as big as it looks: almost all of it is moving everything to a new package. Why did i do that? Originally, i thought i was having package scanning problems because we have the same package in different modules. After i figured out the real problems, it seemed like a good idea to leave the package change in place.


import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/ApplicationConfiguration.java
```diff
@@ -1,4 +1,4 @@
-package org.opentestsystem.rdw.ingest;
+package org.opentestsystem.rdw.ingest.impsvc;
```
malaffoon: @esotrope, i know you won't like this but `import` is not a valid package name so it was either this or something like `import_service` and i don't like underscores in package names. I give you advance permission to smack me during a Thirsty Thursday.

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/multitenant/ImportTenantIdResolver.java
```diff
@@ -30,7 +31,8 @@
      * @throws ClassCastException if current principal is not an SbacUser
      */
     public static SbacUser getCurrentUser() {
-        final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        final SecurityContext context = SecurityContextHolder.getContext();
+        final Authentication authentication = context.getAuthentication();
```
malaffoon: Oops, i should've reverted this. I separated it to help when i was debugging.

import-service/src/test/java/org/opentestsystem/rdw/ingest/impsvc/repository/RdwImportRepositoryIT.java
```diff
@@ -23,28 +15,26 @@
...
 import static java.lang.Thread.sleep;
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Ignore("won't run with multi-tenant changes; i'll revisit after")
 @RunWith(SpringRunner.class)
 @SpringBootTest
 @Transactional
 @ActiveProfiles("test")
+@TestExecutionListeners(value = TenantTestExecutionListener.class,
+    mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
```
malaffoon: The only real change (along with the next file).

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/ApplicationConfiguration.java
esotrope: x D

PR #615, opened 2019-06-03
pipeline-service

malaffoon:
@gregcharles, there is (at least) one more thing that needs to be done here: set up the exam-processor with the system script for validation. I'm going to add a test that asserts that exam-processor has a non-null pipeline, but obviously i'll have to @Ignore that until you make that change.


common/src/main/java/org/opentestsystem/rdw/ingest/common/script/DefaultPipelineService.java
```diff
@@ -0,0 +1,96 @@
...
+ * (and the cache is tenant-aware).
+ */
+public class DefaultPipelineService implements PipelineService {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultPipelineService.class);
+
+    private final TenantIdResolver tenantIdResolver;
+    private final PipelineFactory pipelineFactory;
+
+    private final LoadingCache<String, Optional<String>> versions;  // tenantId_pipelineCode -> active version
+    private final Map<String, Optional<Pipeline>> pipelines;        // tenantId_pipelineCode -> pipeline
```
malaffoon: We want to cache the active version with a short TTL, so we can detect publish changes. We want to cache the pipeline forever (well, until the active version changes).

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/DefaultPipelineService.java
```diff
@@ -0,0 +1,97 @@
...
+
+        versions = CacheBuilder.newBuilder()
+                .maximumSize(100)
+                .expireAfterWrite(5, TimeUnit.MINUTES)
+                .build(new CacheLoader<String, Optional<String>>() {
+                    @Override
+                    public Optional<String> load(final String key) {
+                        return Optional.ofNullable(pipelineRepository.findActiveVersion(key.split("_")[1]));
+                    }
+                });
```
esotrope: i think you can externalize everything up to the build method with a method called "fromSpec" or something.

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorValidationTest.java
```diff
@@ -251,6 +221,37 @@ private void setReposToSunnyDay() {
...
+        // the pipeline has just the system post script
+
+        // this won't get called
+        final PublishedPipelineRepository repository = mock(PublishedPipelineRepository.class);
+
+        // to avoid interfering with other tests we have to have a custom script source
+        final ScriptSource scriptSource = new TestScriptSource();
+
+        // we need a system script property resolver
+        final PropertyResolver propertyResolver = new SystemScriptPropertyResolver(
```
esotrope: wow big constructor x J

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/DefaultPipelineService.java
malaffoon: Oh. Yeah. That looks handy.

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/DefaultPipelineService.java
gregcharles: Makes sense, but you may be burying the lede here. Why did you pull this out of RDW-Common? Is that temporary -- or are the needs of ingest and reporting irreconcilable?

malaffoon: I think they are different needs. Ingest needs to cache versions because it is receiving many requests a second, and it only needs the active pipelines. Reporting wants to know everything about all pipelines and is in user-time, so caching is not good. I suppose the system would be more responsive if we cached pipelines for running tests and stuff, but that caching would be for any/all pipelines (inc. the development pipeline), not just the active published one.

gregcharles: Oh, wait, I got confused with names. This isn't the factory. :)

exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExamProcessorValidationTest.java
gregcharles: Hey now! The property resolver is autowired with the repos needed to do the validations, plus a few more that seemed reasonable for future validations. There are just a lot of repos.

esotrope: haha. constructors like this are why dependency injection was invented. long live DI x D

PR #616, opened 2019-06-05
Externalize pipeline functionality

gregcharles:
This will make it easier to integrate pipelines into other processors.

### DIFF
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
@@ -1,0 +1,85 @@
+package org.opentestsystem.rdw.ingest.common.script;
+
+import java.util.Optional;
+import org.opentestsystem.rdw.common.model.ImportException;
+import org.opentestsystem.rdw.common.model.ImportStatus;
+import org.opentestsystem.rdw.ingest.common.repository.ImportRepository;
+import org.opentestsystem.rdw.ingest.common.util.ByteArrayContentWriter;
+import org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver;
+import org.opentestsystem.rdw.script.Pipeline;
+import org.opentestsystem.rdw.script.PipelineBuildException;
+import org.opentestsystem.rdw.script.PipelineResults;
+import org.opentestsystem.rdw.script.ScriptRuntimeException;
+
+/**
+ * Base class for processors that incorporate a pipeline for pre-processing their incoming messages.
+ */
+public class PipelineProcessor {
+    protected final ImportRepository importRepository;
+    private final PipelineService pipelineService;
+
+    public PipelineProcessor(final ImportRepository importRepository, final PipelineService pipelineService) {
+        this.importRepository = importRepository;
+        this.pipelineService = pipelineService;
+    }
+
+    /**
+     * Get the pipeline by creating it or loading from cache and use it to pre-process the input.
+     *
+     * @param optionalPipelineType the type of the pipeline to use. If empty, the input will be returned unchanged.
+     * @param input the input data to preprocess
+     * @param importId the import id used for updating status in case of a failure
+     * @param archiver the archiver to use for saving successfully preprocessed data if it varies from the input
+     * @return the input altered by the pre-processing or the original input if there is no pipeline
+     * @throws PipelineException
+     */
+    protected byte[] processByPipeline(
+            final Optional<PipelineType> optionalPipelineType,
+            final byte[] input,
+            final long importId,
+            final ByteArrayContentWriter archiver) throws PipelineException {
+
+        if (!optionalPipelineType.isPresent()) {
+            return input;
+        }
+
+        final PipelineType pipelineType = optionalPipelineType.get();
+
+        // get the pipeline dynamically so it can use the current tenant
+        // this expects the factory to deal with tenant and any caching
+        Pipeline pipeline;
+        try {
+            pipeline = pipelineService.getPipeline(pipelineType.code()).orElse(null);
+        } catch (final PipelineBuildException e) {
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
+            throw new PipelineException("Failed to load pipeline " + pipelineType.code(), e);
+        }
+
+        try {
+            if (pipeline == null) {
+                return input;
+            } else {
+                // Run through pipeline and archive the transformation if any
+                final PipelineResults results = pipeline.run(input);
+                final byte [] transformedPayload = results.asByteArray();
+                if (results.isModified()) {
+                    TransformedPayloadArchiver.archive(input, transformedPayload, pipelineType, archiver);
+                }
+                return transformedPayload;
+            }
+        } catch (final ScriptRuntimeException e) {
+            if (e.getCause() instanceof ImportException) {
+                final ImportException ie = (ImportException)e.getCause();
+                importRepository.updateStatusAndMessageById(importId, ie.getStatus(), ie.getMessage());
+                throw new PipelineException("Input failed validation: {}" + ie.getMessage());
+            } else {
+                importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
+                throw new PipelineException("Pipeline script failure: " + e.getMessage());
+            }
+        } catch (final Exception e) {
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
+            throw new PipelineException(
+                    pipelineType.code() + " pipeline failed to process input: " + e.getMessage(), e);
+        }
+    }
+}
--- a/common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
+++ b/common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
@@ -1,6 +1,5 @@
 package org.opentestsystem.rdw.ingest.common.script;
 
-import java.util.Optional;
 import org.opentestsystem.rdw.common.model.ImportException;
 import org.opentestsystem.rdw.common.model.ImportStatus;
 import org.opentestsystem.rdw.ingest.common.repository.ImportRepository;
@@ -14,7 +13,7 @@ import org.opentestsystem.rdw.script.ScriptRuntimeException;
 /**
  * Base class for processors that incorporate a pipeline for pre-processing their incoming messages.
  */
-public class PipelineProcessor {
+public abstract class PipelineProcessor {
     protected final ImportRepository importRepository;
     private final PipelineService pipelineService;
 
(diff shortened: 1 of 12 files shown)


common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+import org.opentestsystem.rdw.ingest.common.util.TransformedPayloadArchiver;
+import org.opentestsystem.rdw.script.Pipeline;
+import org.opentestsystem.rdw.script.PipelineBuildException;
+import org.opentestsystem.rdw.script.PipelineResults;
+import org.opentestsystem.rdw.script.ScriptRuntimeException;
+
+/**
+ * Base class for processors that incorporate a pipeline for pre-processing their incoming messages.
+ */
+public class PipelineProcessor {
```
malaffoon: abstract?

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+
+        final PipelineType pipelineType = optionalPipelineType.get();
+
+        // get the pipeline dynamically so it can use the current tenant
+        // this expects the factory to deal with tenant and any caching
+        Pipeline pipeline;
+        try {
+            pipeline = pipelineService.getPipeline(pipelineType.code()).orElse(null);
+        } catch (final PipelineBuildException e) {
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
```
malaffoon: will e.getMessage() be human interpretable?

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param optionalPipelineType the type of the pipeline to use. If empty, the input will be returned unchanged.
+     * @param input the input data to preprocess
+     * @param importId the import id used for updating status in case of a failure
+     * @param archiver the archiver to use for saving successfully preprocessed data if it varies from the input
+     * @return the input altered by the pre-processing or the original input if there is no pipeline
+     * @throws PipelineException
+     */
+    protected byte[] processByPipeline(
+            final Optional<PipelineType> optionalPipelineType,
```
malaffoon: Optional IN params are a red-flag for me; can we avoid it?

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiverTest.java
```diff
@@ -35,20 +30,20 @@ public void setUp() {
...
+        TransformedPayloadArchiver.archive(originalPayload, transformedPayload, Exam, archiveService);
 
         final ArgumentCaptor<String> locationCaptor = ArgumentCaptor.forClass(String.class);
         final ArgumentCaptor<Properties> propsCaptor = ArgumentCaptor.forClass(Properties.class);
 
         verify(archiveService).writeResource(locationCaptor.capture(), eq(transformedPayload), propsCaptor.capture());
 
         final String location = locationCaptor.getValue();
-        assertThat(location).contains(Content.name());
+        assertThat(location).contains(Exam.importContent().toString());
```
malaffoon: It shouldn't matter here but, in general, the enum `name` should be used for programmatic things. I guess this is a philosophy: is it more likely that the enum consts will be renamed or that somebody will override toString to make changes.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -1,64 +1,54 @@
...
 
-    static final String PipelineName = "exam";
-
-    // Expected type of document after pipeline processing.
-    private static final MimeType ContentType = APPLICATION_XML;
-
-    // The expected Import Content for this processor.
-    private static final ImportContent Content = ImportContent.EXAM;
+    @VisibleForTesting
+    static final PipelineType pipelineType = PipelineType.Exam;
```
malaffoon: nit: `PipelineType`

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -82,45 +72,22 @@ public void process(final Message<?> message) {
...
+                logger.error("Caused by:", e.getCause());
             }
             return;
-        } catch (final Exception e) {
-            logger.warn("pipeline failed to convert xml to tds report with error: " + e.getMessage());
-            importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
-            return;
         }
 
+        final TDSReport report = tdsReportFromXml(new ByteArrayInputStream(preProcessedPayload));
```
malaffoon: this needs to be in a try/catch ... absolutely no exceptions can escape this method

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
```diff
@@ -82,45 +72,22 @@ public void process(final Message<?> message) {
...
-        } catch (final ScriptRuntimeException e) {
-            logger.warn("pipeline failed to convert xml to tds report with error: " + e.getMessage());
-            if (e.getCause() instanceof ImportException) {
-                final ImportException ie = (ImportException)e.getCause();
-                importRepository.updateStatusAndMessageById(importId, ie.getStatus(), ie.getMessage());
-            } else {
-                importRepository.updateStatusAndMessageById(importId, ImportStatus.BAD_DATA, e.getMessage());
+            preProcessedPayload = processByPipeline(
+                    Optional.of(pipelineType), payload, importId, archiveService::writeResource);
+        } catch (PipelineException e) {
```
malaffoon: How does knowing this is a PipelineException help? Shouldn't this be catching any exception since we don't want to allow exceptions to escape? If the answer is, the helper converts all exceptions to PipelineExceptions, then i don't see the value in that. And i'm never a fan of creating a custom Exception that serves no purpose.

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+                importRepository.updateStatusAndMessageById(importId, ie.getStatus(), ie.getMessage());
+                throw new PipelineException("Input failed validation: {}" + ie.getMessage());
+            } else {
+                importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
+                throw new PipelineException("Pipeline script failure: " + e.getMessage());
+            }
+        } catch (final Exception e) {
+            importRepository.updateStatusAndMessageById(importId, ImportStatus.PIPELINE_FAILURE, e.getMessage());
+            throw new PipelineException(
+                    pipelineType.code() + " pipeline failed to process input: " + e.getMessage(), e);
```
esotrope: can we use `String.format()` makes for easier reading and editing later imo - willing to accept challenges tho

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineType.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static org.springframework.util.MimeTypeUtils.APPLICATION_XML;
+
+/**
+ * Data about the pipelines types that exist.
+ *
+ * CSV MIME type from RFC-7111
+ */
+public enum PipelineType {
+    Exam(APPLICATION_XML, ImportContent.EXAM),
```
esotrope: can this not be DB driven?

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
gregcharles: For the field name? That's not actually an option because of the name conflict. I could use PIPELINE_TYPE though, or ... hmm, just Type? I don't really like that.

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
gregcharles: I was just checking for null, but changed because I thought you'd like Optional more. I don't really need either one, but it integrates a bit more smoothly into PackageProcessor if I can check for null or empty. That's because the PackageProcessor uses a pipeline for assessments, but not for other types. It's just slightly smoother though. I could definitely put the extra logic in PackageProcessor.

common/src/test/java/org/opentestsystem/rdw/ingest/common/util/TransformedPayloadArchiverTest.java
gregcharles: Hmm, yeah. I'll change it because the LocationStrategy is using name(), so the test should too. I'm actually not totally comfortable with that. What if someone refactored to make the names Pascal case instead of all caps? I think it would be better if the enum had a specific method for getting a path component and the LocationStrategy used that.

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
gregcharles: Hmm, yes, in most cases at least, and subject to what you mean by human readable. It will mostly be ImportExceptions, which are sometimes strings and other times json, but basically human readable. It can also be script runtime errors, which would indicate the script line and the problem. Those would be program-y type messages, but give information for finding the problem. Hopefully those will be rare because the scripts are tested before they are published, but could happen if some condition in the input wasn't anticipated. In rare cases, there could be a db connection error, or something like that, but then we'd get a stack trace too.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
gregcharles: I could also change PipelineType to something else. PipelineEnum? PipelineConfig?

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
malaffoon: It's just checking isPresent once at the beginning so i'd put that logic outside this method.

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
malaffoon: Okay, sounds good. I just wanted to make sure we are going to get a message. Sometimes the exception getMessage() returns nothing and wanted to know that wasn't the case here. Thanks.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ExamProcessor.java
malaffoon: Doh, didn't notice that. Nah, just leave it (or do PIPELINE_TYPE if you must).

PR #617, opened 2019-06-05
Custom script to transform TRTs from external vendor

gregcharles:
I'm assuming schooIId is in the "schoolId" examinee relationship. Is that a safe assumption? I could duplicate the "best relationship" match if needed, but it would get a bit messy.

### DIFF
--- a/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExternalVendorScriptTest.java
+++ b/exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExternalVendorScriptTest.java
@@ -1,0 +1,105 @@
+package org.opentestsystem.rdw.ingest.processor;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.Before;
+import org.junit.Test;
+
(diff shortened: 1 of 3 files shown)


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/ExternalVendorScriptTest.java
```diff
@@ -0,0 +1,105 @@
...
+        assertThat(before.getTest().getBankKey()).isEqualTo(10200L);
+        assertThat(after.getTest().getBankKey()).isEqualTo(200L);
+
+        // Find all "bad" bank keys among the items in the before data
+        final List<Long> keyList = before.getOpportunity().getItem()
+                .stream()
+                .filter(item -> item.getBankKey() == 10200L)
+                .map(Opportunity.Item::getKey)
+                .collect(Collectors.toList());
+        assertThat(keyList.size()).isNotZero();
```
malaffoon: you can do `assertThat(keyList).isNotEmpty()`

malaffoon:
Awesome possum, thanks Greg.

PR #619, opened 2019-06-06
Integrate pipeline to Assessment Packages

gregcharles:
NOTES: I kept PipelineType out of the PipelineProcessor constructor because PackageProcessor may potentially handle several different pipeline types. I kept ArchiveService out because I didn't want ingest-common to need a dependency on common-archive just for that. However, the dependency snuck in anyway via common-script, so there was no point. I did make that dependency explicit in ingest-common's build.gradle though, so if it is every removed from script-common, it won't break ingest.


exam-processor/src/main/resources/application.yml
```diff
@@ -90,3 +90,10 @@ validation:
     - IDEAIndicator
     - Completeness
     - AdministrationCondition
+
+tenantProperties:
```
malaffoon: i assume you'll revert this before committing?

common/src/main/java/org/opentestsystem/rdw/ingest/common/script/PipelineProcessor.java
```diff
@@ -54,7 +49,7 @@ protected PipelineProcessor(final PipelineType pipelineType,
      * @param importId the import id used for updating status in case of a failure
      * @return the input altered by the pre-processing or the original input if there is no pipeline
      */
-    protected byte[] processByPipeline(final byte[] input, final long importId) {
+    protected byte[] processByPipeline(final PipelineType pipelineType, final byte[] input, final long importId) {
```
malaffoon: Ah, thanks for the explanation, makes sense.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ApplicationConfiguration.java
```diff
@@ -27,4 +36,13 @@ public SubjectSerializationService subjectSerializationService() {
...
         return new TabulatorService();
     }
+
+    @Bean
+    public PipelineService pipelineService(final TenantIdResolver tenantIdResolver,
+                                           final PipelineFactory pipelineFactory,
+                                           final PipelineRepository pipelineRepository,
+                                           @Value("${pipeline.cache.versions:maximumSize=100,expireAfterWrite=300s}")
+                                           final String cacheConfig) {
+        return new DefaultPipelineService(tenantIdResolver, pipelineFactory, pipelineRepository, cacheConfig);
```
esotrope: are we able to use spring `@Cacheable` annotation for the caching? then we can externalize the cache setup entirely to the yml file where you can choose what kind of cache, max size etc.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessor.java
```diff
@@ -20,10 +24,9 @@
 import org.opentestsystem.rdw.multitenant.TenantContextHolder;
 
 @EnableBinding(ImportPayloadSink.class)
-public class PackageProcessor {
+public class PackageProcessor extends PipelineProcessor {
```
esotrope: what is the benefit of extending here vs composing?

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessor.java
```diff
@@ -90,8 +104,23 @@ private void handleIncomingMessage(final Message<?> message, final ImportPayload
         // package processing is rarer so log at INFO level
         logger.info("received {} import {} for tenant id {}", accessor.getContent(), importId, tenantId);
 
+        // Process payload by pipeline if one exists
+        byte[] preProcessedPayload;
+        if (pipelineType == null) {
+            preProcessedPayload = payload;
+        } else {
```
malaffoon: you could set to payload and then avoid the extra block, i.e.
```
byte[] preProcessedPayload = payload;
if (pipelineType != null) {
  ...
}
```

exam-processor/src/main/resources/application.yml
gregcharles: Dang it. Yes, I'll fix that.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ApplicationConfiguration.java
malaffoon: The caching here is specialized. The active version is cached with a short TTL, not sure Spring's simple caches allow for that. And the pipeline cache enforces having a single instance of each pipeline per tenant (probably we could make that a simple Spring cache but it didn't wire up right when i tried it the first time).

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/PackageProcessor.java
malaffoon: The obvious one is that the import repository is shared but you're right, it could be hasA instead of isA.

gregcharles: I figured that question would come up. Mostly, it was just easier. Also, the perfect composition would be PipelineService, but that name is already being used, so I'd have to change that to PipelineCache or something. There's also the potential to override some of the behavior this way.

package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/ApplicationConfiguration.java
gregcharles: That makes sense, but I'll have to defer to Mark for that.

esotrope: hmm i was under the impression you could break out per-cache configs in the yml but i suddenly cant find an example of it on google.
I was thinking we could use a guava cache with that spec config but so far no luck finding an example..

PR #623, opened 2019-07-06
need to temporarily set tenant when resolving permissions for user

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/DefaultAuthorityService.java
```diff
@@ -134,14 +138,28 @@ public DefaultAuthorityService(final PermissionService permissionService,
...
+        final String stateId = grants.stream().map(Grant::getStateId).findFirst().orElse(null);
+
+        if (stateId == null) {
+            // to simplify the calling code, return a collection of empty sets
+            orgMap.put(DISTRICT_GROUP, Collections.emptyMap());
+            orgMap.put(DISTRICT, Collections.emptyMap());
+            orgMap.put(INSTITUTION_GROUP, Collections.emptyMap());
+            orgMap.put(INSTITUTION, Collections.emptyMap());
+        } else {
+            TenantContextHolder.setTenantId(stateId);
```
malaffoon: This is the change. Not dissimilar to how reporting does it.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingConfiguration.java
```diff
@@ -70,7 +70,7 @@ public Job migrateReportingJob(final JobExecutionListener listener,
                                    final Step finalizeStep) {
         try {
 
-            return jobBuilderFactory.get("Migrate OlAP Reporting Job")
+            return jobBuilderFactory.get("Migrate OLAP Reporting Job")
```
malaffoon: unrelated typo that's been driving me nuts for a couple years now

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/DefaultAuthorityService.java
esotrope: this feels a little awkward. i think these conditions go hand in hand no?:

IF a user has a district|school level grant THEN the user has a state ID in that grant.
If that's the case we shouldn't need an if/else block and we can just always set tenant ID to `grants.stream().map(Grant::getStateId).findFirst().orElse(null);`
I feel like doing that and maybe including a code comment to the effect that the map will be empty when the user has only client-level grants would be more streamline

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingConfiguration.java
esotrope: haha. the ol'ap

PR #624, opened 2019-07-15
make sure resubmit has a limit set

malaffoon:


import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/web/ScopedImportController.java
```diff
@@ -109,13 +109,23 @@ protected RdwImportResource uploadImport(final MultipartHttpServletRequest reque
...
+                    .build();
+        } else {
+            // if it isn't empty, make sure the content is correct and there is some limit
+            final RdwImportQuery.Builder builder = inputQuery.copy().content(this.content);
+            if (inputQuery.getLimit() == null) {
+                builder.limit(100);
+            }
+            query = builder.build();
+        }
+
```
malaffoon: Slightly less efficient since it always makes a copy of the query (before it only copied if the content wasn't set properly).

PR #629, opened 2019-07-27
RP-508: Creating a Sandbox/Tenant must migrate and initialize cached values

gregcharles:
Adds event listener to OLAP migration job and serializes calls to this method (scheduled, event triggered, or otherwise called via proxy) so that one must complete before another starts.

### DIFF
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
@@ -4,7 +4,9 @@ import org.springframework.batch.core.Job;
 import org.springframework.batch.core.launch.JobLauncher;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cloud.context.environment.EnvironmentChangeEvent;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.event.EventListener;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.scheduling.annotation.EnableAsync;
 import org.springframework.scheduling.annotation.EnableScheduling;
@@ -40,8 +42,9 @@ class MigrateOlapReportingJobRunner extends TenantAwareMigrateJobRunner {
         super(migrateRepository, importRepository, tenantIdResolver, jobLauncher, job, batchSize, tenantProperties);
     }
 
-    @Async
+    @Async("serialExecutor")
     @Scheduled(cron = "${migrate.olap-batch.run-cron}", zone = "GMT")
+    @EventListener(classes = EnvironmentChangeEvent.class)
     @Override
     public void run() {
         super.run();
--- a/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
+++ b/migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
@@ -42,7 +42,15 @@ class MigrateOlapReportingJobRunner extends TenantAwareMigrateJobRunner {
         super(migrateRepository, importRepository, tenantIdResolver, jobLauncher, job, batchSize, tenantProperties);
     }
 
-    @Async("serialExecutor")
+    /**
+     * Runs the migrate job. Each call will be serialized so that one must finish before the next starts. This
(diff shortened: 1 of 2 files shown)


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
```diff
@@ -40,8 +42,9 @@ public MigrateOlapReportingJobRunner(final MigrateRepository migrateRepository,
         super(migrateRepository, importRepository, tenantIdResolver, jobLauncher, job, batchSize, tenantProperties);
     }
 
-    @Async
+    @Async("serialExecutor")
     @Scheduled(cron = "${migrate.olap-batch.run-cron}", zone = "GMT")
+    @EventListener(classes = EnvironmentChangeEvent.class)
```
ftdholbrook: Elegant solution @gregcharles I like it.  
Reading: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/event/EventListener.html 

I think it may be slightly better to add a second "eventRun" method that takes the event type as and argument and also is surrounding by a try/catch block that only logs.  In my experience spring becomes very unstable if and event driven method throws.  That said, I don't think this method is likely to throw here at "run".

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/SpringAsyncConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @return an Executor that will enforce that one job finishes before the next can begin
+     */
+    @Bean(name = "serialExecutor")
+    public Executor serialExecutor() {
+        return new SerialExecutor(new SimpleAsyncTaskExecutor());
+    }
+
+    // Serial Executor adapted from java.util.Executor java docs.
+    private static class SerialExecutor implements Executor {
```
ftdholbrook: did you consider https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor--
I believe it accomplishes the same thing and is a tested part of the JDK.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
malaffoon: The run method, as written, won't throw (well, except for extraordinary situations in which case this is the least of our problems). However, we should add a comment that, because of the event listener, it shouldn't be changed in a way that would allow it to throw.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/SpringAsyncConfiguration.java
malaffoon: 👍

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
gregcharles: I put in the comment as Mark suggested. The separate method could be a problem since I think if run() were called by another method within the class, it wouldn't run with the proxy and so wouldn't honor the @Async annotation. However, as far as I can tell from my testing, if we annotated both methods with @Async("singleThreadExecutor"), they would both use the same executor and so still be serialized.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/SpringAsyncConfiguration.java
gregcharles: Yes, this seems to work the same way, so I changed to it. The only problem I could think of is that we might want to drop duplicate jobs (e.g., if we get five new calls to migrate while a previous migrate was still running, we would really only need to run one of them.) That would be easier to implement if we had access to the queue.

migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/MigrateOlapReportingJobRunner.java
gregcharles: I also checked against the MVC endpoint-triggered call to the run method and it gets serialized with the scheduled and event listener runs, so we should be OK there too.

PR #631, opened 2019-08-01
attempted fix for java.lang.IllegalStateException

ftdholbrook:
attempted fix for java.lang.IllegalStateException: No value for key [..TenantDynamicRoutingDataSource...] bound to thread

- unable to test this locally, but I don't think it can make things any worse
- plan to deploy and evaluate, if it is worse I will quickly back it out


ftdholbrook:
for RP-645

PR #637, opened 2019-08-02
push end-point migrate calls through the same executor

malaffoon:


migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/SpringAsyncConfiguration.java
```diff
@@ -0,0 +1,26 @@
+package org.opentestsystem.rdw.ingest.migrate.reporting;
```
malaffoon: For now i just copied this in both modules.

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -39,9 +41,26 @@ public MigrateReportingJobRunner(final MigrateRepository migrateRepository,
         super(migrateRepository, importRepository, tenantIdResolver, jobLauncher, job, batchSize, tenantProperties);
     }
 
+    /**
+     * Override to use the same executor to serialize work and avoid duplicating datasources
+     * and other resources.
+     */
+    @Async("singleThreadExecutor")
     @Scheduled(fixedDelayString = "${migrate.batch.delay}")
```
malaffoon: So one thing that bothers me here. If this is executed async with a fixed delay ... when does the delay counter start? @ftdholbrook , thoughts?

ftdholbrook: might want to add an initial delay to give it a pause before it starts up and starts running on the fixed delay.  Also might want to make sure it can run consistently in the amount of delay.

malaffoon: Reading about it doesn't sound promising. The interaction of Async and Scheduled is ... funkky.

ftdholbrook: yes poking around stack overflow looks like there are a fair number of problems with using them at the same time.

malaffoon: It absolutely does NOT run in the amount of the delay.

ftdholbrook: I guess because it is on the same method as async it isn't async?  because if it was async it would return immediately and delay and rate would be effectively the same thing?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -19,29 +23,59 @@
 
 /**
  * Scheduled execution of the {@link MigrateReportingConfiguration#jobBuilderFactory}'s job.
```
malaffoon: @ftdholbrook can you look at this approach and see what you think?

migrate-reporting/src/main/java/org/opentestsystem/rdw/ingest/migrate/reporting/MigrateReportingJobRunner.java
```diff
@@ -19,29 +23,59 @@
...
+     * Override to use the same executor to serialize work and avoid duplicating datasources
+     * and other resources.
+     */
+    @Scheduled(initialDelayString = "${migrate.batch.delay}", fixedDelayString = "${migrate.batch.delay}")
     @Override
     public void run() {
-        super.run();
+        try {
+            // get the results before returning so the scheduled delay is correct
+            executorService.submit((Runnable) super::run).get();
```
ftdholbrook: I think this should work, get will block until it completes

PR #638, opened 2019-08-03
remove any hint of query timeout

malaffoon:
setQueryTimeout is not good, remove it. Increase validationInterval from default (3s) to something bigger.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ apply plugin: 'io.spring.dependency-management'
 ext['groovy.version'] = '2.5.6'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
-String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-176"
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-SNAPSHOT"
 String rdwSchemaVersion = project.hasProperty("schema") ? project.property("schema") : "1.4.0-428"
 String dockerPrefix = "smarterbalanced"
 
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ apply plugin: 'io.spring.dependency-management'
 ext['groovy.version'] = '2.5.6'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -23,7 +23,7 @@ apply plugin: 'io.spring.dependency-management'
 ext['groovy.version'] = '2.5.6'
 
 // Set version dependencies. This allows the cli to force version, e.g. `-Pcommon=1.2.0-SNAPSHOT`
-String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-176"
+String rdwCommonVersion = project.hasProperty("common") ? project.property("common") : "1.4.0-SNAPSHOT"
```
malaffoon: waiting for common build

docker-compose.yml
```diff
@@ -42,6 +42,7 @@ services:
     ports:
       - 8080:8080
       - 8081:8008
+      - 9090:9090
```
malaffoon: oops, this shouldn't be in here dang it

docker-compose.yml
```diff
@@ -55,22 +56,22 @@ services:
...
+#    links:
+#      - config-server
+#      - rabbitmq
+#    depends_on:
+#      config-server:
+#        condition: service_healthy
+#    environment:
+#      - SPRING_PROFILES_ACTIVE=local-docker
+#      - CONFIG_SERVICE_ENABLED=true
+#      - CONFIG_SERVICE_URL=http://config-server:8888
```
gregcharles: Did you mean to check in this change to docker-compose?

common/src/test/resources/application-common-test.yml
```diff
@@ -15,12 +15,11 @@ datasources:
       &connectTimeout=${datasources.warehouse_rw.connect-timeout:10000}\
       &socketTimeout=${datasources.warehouse_rw.socket-timeout:40000}\
       "
-    jdbcInterceptors: "QueryTimeoutInterceptor(queryTimeout=${datasources.warehouse_rw.query-timeout:30})"
     username: root
     password:
     testWhileIdle: true
     validationQuery: SELECT 1
-    validationQueryTimeout: 10000
+    validationInterval: 35000
```
malaffoon: Why 35000?  Because moose.

docker-compose.yml
malaffoon: Nope, reverted.

PR #645, opened 2019-08-16
removes school year config from app ymls

esotrope:


migrate-olap/src/main/resources/application.yml
```diff
@@ -40,9 +40,6 @@ migrate:
     # every day at 1:00 am
     run-cron: "0 0 1 * * *"
 
-reporting:
-  school-year: 2019
-
```
malaffoon: tests run without any other changes?

esotrope: ran all the tests in this module and they passed (minus RST b/c i'm not setup for that locally)

PR #648, opened 2019-08-23
RP-721: integrate import service OAuth with Okta.

gregcharles:
Enables validating Okta access tokens based on configuration, while still supporting OpenAM during transition period. Per Jeff Khoury, integration to ART will still use OpenAM for the forseeable future.

### DIFF
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
@@ -100,24 +100,24 @@ class OktaTokenServices implements ResourceServerTokenServices {
         Map<String,Object> sanitized = new HashMap<>(claims);
 
         // The EXPIRES_IN seconds can be computed from EXP
-        if (claims.containsKey(EXP) && !claims.containsKey(EXPIRES_IN)) {
+        if (sanitized.containsKey(EXP) && !sanitized.containsKey(EXPIRES_IN)) {
             try {
                 long now = System.currentTimeMillis() / 1000L;
-                long exp = ((Number)claims.get(EXP)).longValue();
+                long exp = ((Number)sanitized.get(EXP)).longValue();
                 sanitized.put(EXPIRES_IN, (int)(exp - now));
             } catch (Exception e) {
-                logger.info("Cannot compute expires-in value from exp: " + claims.get(EXP));
+                logger.info("Cannot compute expires-in value from exp: " + sanitized.get(EXP));
             }
         }
 
         // Tenancy chain may be in a list, but must be joined into a string.
-        final Object tenancyChain = claims.get(SBAC_TENANCY_CHAIN);
+        final Object tenancyChain = sanitized.get(SBAC_TENANCY_CHAIN);
         if (tenancyChain instanceof Collection) {
             String tenancyChainString = StringUtils.join(((Collection) tenancyChain), ",");
             sanitized.put(SBAC_TENANCY_CHAIN, tenancyChainString);
         }
 
-        if (!claims.containsKey(GRANT_TYPE)) {
+        if (!sanitized.containsKey(GRANT_TYPE)) {
             sanitized.put(GRANT_TYPE, "password");
         }
 
--- a/import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
+++ b/import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
@@ -1,0 +1,127 @@
+package org.opentestsystem.rdw.ingest.impsvc.auth;
+
+import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.UncheckedExecutionException;
+import com.okta.jwt.AccessTokenVerifier;
+import com.okta.jwt.JwtVerificationException;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.authentication.AuthenticationServiceException;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.security.oauth2.common.exceptions.InvalidTokenException;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
+import org.springframework.security.oauth2.provider.token.AccessTokenConverter;
+import org.springframework.security.oauth2.provider.token.ResourceServerTokenServices;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import static org.springframework.security.oauth2.common.OAuth2AccessToken.EXPIRES_IN;
+import static org.springframework.security.oauth2.common.util.OAuth2Utils.GRANT_TYPE;
+
+/**
+ * An implementation of {@link ResourceServerTokenServices} backed by Okta. It expects the access token to be
+ * a JWT containing all the necessary OAuth information in its claims.
+ *
+ * @see SecurityConfigurer
+ * @see SbacTokenConverter
+ */
+class OktaTokenServices implements ResourceServerTokenServices {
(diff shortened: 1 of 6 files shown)


import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/SecurityConfiguration.java
```diff
@@ -1,26 +1,36 @@
...
     @ConfigurationProperties(prefix = "security.oauth2")
-    public ForgeRockTokenServices forgeRockTokenServices(final SbacTokenConverter tokenConverter) {
+    public ResourceServerTokenServices tokenServices(
+            final SbacTokenConverter tokenConverter,
+            @Value("${security.oauth2.provider:}") final String provider) {
+
+        // For transition period, support both OpenAM and Okta, depending on configured provider.
+        if (OKTA_PROVIDER.equalsIgnoreCase(provider)) {
+            return new OktaTokenServices(tokenConverter);
+        }
```
esotrope: i'm interested in the security flow here. is rdw ingest using this new code just to verify the token and extract permissions issued by okta? or is it involved in the token issuing? Also what makes an okta jwt different from a forgerock jwt if they have the same secret key?

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
```diff
@@ -0,0 +1,127 @@
...
+                long exp = ((Number)claims.get(EXP)).longValue();
+                sanitized.put(EXPIRES_IN, (int)(exp - now));
+            } catch (Exception e) {
+                logger.info("Cannot compute expires-in value from exp: " + claims.get(EXP));
+            }
+        }
+
+        // Tenancy chain may be in a list, but must be joined into a string.
+        final Object tenancyChain = claims.get(SBAC_TENANCY_CHAIN);
+        if (tenancyChain instanceof Collection) {
```
malaffoon: What if the tenancy chain isn't a collection? 

If it is always a collection the comment is a bit misleading, perhaps tweak it.

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
```diff
@@ -0,0 +1,127 @@
...
+            throw new InvalidTokenException(token + ": " + e.getMessage());
+        }
+    }
+
+    // Adjusts contents of the JWT claims so they synch with what the token convert expects.
+    private Map<String, Object> sanitize(final Map<String, Object> claims) {
+        Map<String,Object> sanitized = new HashMap<>(claims);
+
+        // The EXPIRES_IN seconds can be computed from EXP
+        if (claims.containsKey(EXP) && !claims.containsKey(EXPIRES_IN)) {
```
malaffoon: It is fairly confusing that this logic is querying `claims`, could you change it to `sanitized`?

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/SecurityConfiguration.java
```diff
@@ -1,26 +1,36 @@
...
 
     @Bean
     @ConfigurationProperties(prefix = "security.oauth2")
-    public ForgeRockTokenServices forgeRockTokenServices(final SbacTokenConverter tokenConverter) {
+    public ResourceServerTokenServices tokenServices(
+            final SbacTokenConverter tokenConverter,
+            @Value("${security.oauth2.provider:}") final String provider) {
+
+        // For transition period, support both OpenAM and Okta, depending on configured provider.
+        if (OKTA_PROVIDER.equalsIgnoreCase(provider)) {
```
malaffoon: In general for this kind of thing you can use @Conditional annotations. That is, two separate bean methods, conditional on the provider.

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import static org.springframework.security.oauth2.common.OAuth2AccessToken.EXPIRES_IN;
+import static org.springframework.security.oauth2.common.util.OAuth2Utils.GRANT_TYPE;
+
+/**
+ * An implementation of {@link ResourceServerTokenServices} backed by Okta. It expects the access token to be
+ * a JWT containing all the necessary OAuth information in its claims.
```
malaffoon: I'm not sure this is accurate: this implementation doesn't expect the accessToken to be anything, it just passes it to the helper to deal with.

import-service/src/test/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServicesTest.java
```diff
@@ -0,0 +1,106 @@
...
+    private OktaTokenServices services;
+
+    @Before
+    public void createServices() throws Exception {
+        jwtHelper = mock(OktaJwtHelper.class);
+        jwtVerifier = mock(AccessTokenVerifier.class);
+        jwt = mock(Jwt.class);
+
+        when(jwtHelper.getJwtVerifier()).thenReturn(jwtVerifier);
+        when(jwtVerifier.decode(anyString())).thenReturn(jwt);
```
malaffoon: could make this a teeny bit more robust (and reduce code) if you had a class-level accessToken and made this `when(jwtVerifier.decode(accessToken)).thenReturn(jwt);`

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/SecurityConfiguration.java
gregcharles: RDW ingest is just verifying the access token that the third-party retrieved by a direct call, with their own credentials,  to either the OpenAM/Forgerock or Okta server. In the OpenAM case, the access token is a UUID, which RDW uses to download the tenancy chain and other data in a separate call to the OpenAM server. In the Okta case, RDW decodes the access token, which is a JWT, and it contains all the necessary data, so _usually_ no additional network call is needed. However, part of the decoding checks the JWT signature to see that it's valid and the JWT hasn't been modified in-flight. This requires a behind-the-scenes call to the Okta server to retrieve the public key used for the verification, but that's done by Okta libraries, and it's cached for (I think) a day or so.

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
gregcharles: Done

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/SecurityConfiguration.java
gregcharles: Yes, I thought about redoing that, but I wasn't sure how to modify @ConditionalOnMissingBean(StubSecurityConfiguration.class) to deal with the now three possible cases, and figured I could be lazy since this is only in place to support the transitional period.

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/SecurityConfiguration.java
esotrope: cool thanks for splainin

import-service/src/main/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServices.java
gregcharles: I don't think so. The helper really just builds a JwtVerifier. If the JWT in the access token doesn't have the right information -- say it's missing the tenancy chain -- the verifier could still pass, but the auth will fail.

import-service/src/test/java/org/opentestsystem/rdw/ingest/impsvc/auth/OktaTokenServicesTest.java
gregcharles: Done

PR #649, opened 2019-08-23
redo query to improve performance

malaffoon:
The optimizer just wasn't dealing with this query. In dev, it went from 1.5h to 12s with this change.

### DIFF
--- a/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
+++ b/migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
@@ -1228,28 +1228,30 @@ sql:
         # ------------ exam_target_score  -------------------------------------------------------------------
         # Only Summative target scores are included
         # For Smarter Balanced Math, only the Concepts and Procedures targets are included
+        # The exam_target_score table can be very large; the sub-select forces the optimizer to do the right thing
         exam_target_score:
           sql:
             warehouseRead: >-
               SELECT
-                wets.id,
-                wets.exam_id,
-                wets.target_id,
-                wets.student_relative_residual_score,
-                wets.standard_met_relative_residual_score,
-                :migrate_id as migrate_id
-              FROM exam_target_score wets
-                JOIN exam we ON wets.exam_id = we.id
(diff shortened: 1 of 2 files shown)


task-service/src/main/resources/application.yml
```diff
@@ -73,11 +73,9 @@ task:
 #    cron: 0 0 4 * * *
 
 taskSendReconciliationReport:
+  # default root settings
   query: status=PROCESSED&after=-PT24H
   log: false
-
-#taskSendReconciliationReport:
-#  # there is no root (non-tenant) configuration for this task
```
malaffoon: unrelated, clean up annotated configuration file

migrate-olap/src/main/resources/application.warehouse.to.staging.sql.yml
```diff
@@ -1228,28 +1228,30 @@ sql:
...
-              FROM exam_target_score wets
-                JOIN exam we ON wets.exam_id = we.id
-                JOIN target wt ON wt.id = wets.target_id
+                sub.id,
+                sub.exam_id,
+                sub.target_id,
+                sub.student_relative_residual_score,
+                sub.standard_met_relative_residual_score,
+                :migrate_id AS migrate_id
+              FROM (SELECT wets.id, wets.exam_id, wets.target_id, wets.student_relative_residual_score, wets.standard_met_relative_residual_score
```
esotrope: surprised this is faster since there is a sub select.

malaffoon: The optimizer completely borked it for some reason. It was joining the exam-target-score table with the three reference tables FIRST, getting 350m results, then joining the exam table. This just forces it to do it the other way around.

agorina: Just thinking loud: would it make sense to make the same change for the reporting migrate?

malaffoon: Probably. But no problems have been reported and i've heard that i don't cut enough corners ...

PR #651, opened 2019-09-16
fix assessment update detection bug

malaffoon:


package-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultAssessmentPackageProcessor.java
```diff
@@ -114,11 +114,11 @@ private WarehouseAssessment processAssessmentForUpdate(final WarehouseAssessment
...
-        if (!Objects.equals(o1.getName(), o2.getName())
-         && Objects.equals(o1.getLabel(), o2.getLabel())
-         && Objects.equals(o1.getVersion(), o2.getVersion())
-         && (o1.getTypeId() == o2.getTypeId())
-         && (o1.getSchoolYear() == o2.getSchoolYear())) {
+        // get out early if some simple metadata don't match ...
+        if (!(Objects.equals(o1.getName(), o2.getName())
+           && Objects.equals(o1.getLabel(), o2.getLabel())
+           && Objects.equals(o1.getVersion(), o2.getVersion())
+           && (o1.getSchoolYear() == o2.getSchoolYear()))) {
```
malaffoon: I scoured the code history on this one and i think it is a bug that has been there for a while. If the name matches, the other fields will not be checked and the new assessment will be considered equals even if those other fields mismatch. Not good. Clearly in the code history, there was a copy and paste event where the "not" logic wasn't scoped properly.

ftdholbrook: Well that's not good.

PR #653, opened 2019-09-26
RDW-753: migrate jobs sometimes fail after tenant creation

gregcharles:


migrate-reporting/src/main/resources/application.yml
```diff
@@ -52,6 +52,15 @@ spring:
...
           destination: ConfigpropsUpdateExchange
+        MigrateJobInput:
+          content-type: application/json
+          destination: MigrateReportingJobExchange
+          group: MigrateReportingJobQueue
+          producer:
+            requiredGroups: MigrateReportingJobQueue
+        MigrateJobOutput:
+          content-type: application/json
+          destination: MigrateReportingJobExchange
```
ftdholbrook: I was confused about this for a moment, so @gregcharles  can you verify:  because of the inheritance structure you are using the same binding names "MigrateJobInput" and "MigrateJobOutput" from the processor, but each service is using a different underlying exchange and queue name.   Did you take a look at rabbit and verify that's working how you expected?  I was concerned for a moment that sending a reporting migrate request would trigger an olap request and vice versa.  This seems like it would work.

gregcharles: I didn't pull up Rabbit's management console. I'll check on that now.

gregcharles: Yes, it's getting messages through both exchanges. (I made a local change to the cron expression for the OLAP migration to trigger every two minutes.) I thought since the MIgrateJobProcessor is in migrate-common and it defines "MigrateJobInput" and "MIgrateJobOutput" that those have to match in the reporting and olap properties. Let me know if I'm wrong about that though.

malaffoon:
GitHub presentation isn't making it obvious so i should probably pull down the patch and look at it but i can't answer my one main question ...

Before this change, the migrate would happen on one scheduled thread, while the configuration reset would happen on a separate thread (MQ listeners have their own thread). How does this change make it so that the listener for the migrate and the listener for the config reset can't happen at the same time?

gregcharles:
Merging to test.

PR #654, opened 2019-10-01
add refresh scope to all DataSourceElementResolvers

ftdholbrook:
make consistent with other usage in ingest and reporting


migrate-olap/src/main/java/org/opentestsystem/rdw/ingest/migrate/olap/repository/DataSourceConfiguration.java
```diff
@@ -19,6 +19,7 @@
 import org.opentestsystem.rdw.multitenant.datasource.DataSourceElementsProperties;
 import org.opentestsystem.rdw.multitenant.datasource.DataSourceElementsResolver;
 import org.opentestsystem.rdw.multitenant.datasource.TenantDynamicRoutingDataSource;
+import org.springframework.web.bind.annotation.RequestParam;
```
ftdholbrook: not sure why that got in there...

gregcharles: IntelliJ gets overeager with auto complete. You have to fix where you were typing and then go remove the import it's added. I've had that happen too.

PR #656, opened 2019-10-28
add check that assessment names are not duplicated across subjects

malaffoon:
This adds a validation that the assessment name, e.g. "SBAC-IAB-G8E-Perf-Fubar", is not duplicate across subjects. It is allowed and expected to be duplicated within a subject since the name indicates a year-over-year continuity for IABs.

### DIFF
--- a/validator/src/main/java/org/opentestsystem/rdw/ingest/validator/service/AssessmentChecker.java
+++ b/validator/src/main/java/org/opentestsystem/rdw/ingest/validator/service/AssessmentChecker.java
@@ -3,7 +3,6 @@ package org.opentestsystem.rdw.ingest.validator.service;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 import org.opentestsystem.rdw.common.model.AssessmentType;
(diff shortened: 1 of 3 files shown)


validator/src/main/java/org/opentestsystem/rdw/ingest/validator/service/AssessmentChecker.java
```diff
@@ -31,34 +32,47 @@ static void checkAssessments(final MessageSink messageSink,
...
+            );
+
+        // duplicate names across subjects are probably wrong and have adverse affects in the system
+        // (this was the instructional resource problem when Math and ELA had assessments with the same name)
+        assessments.stream()
+            .collect(groupingBy(Assessment::getName, mapping(Assessment::getSubject, toSet())))
+            .entrySet().stream().filter(entry -> entry.getValue().size() > 1)
+            .forEach(entry ->
+                messageSink.error("duplicate assessment name {0} across subjects {1}", entry.getKey(), entry.getValue().toString())
+            );
```
malaffoon: This is the new bit. The rest is me reformatting to avoid needless collection of things.

PR #659, opened 2019-12-10
use PostgreSql no-op instead of empty list

ftdholbrook:


migrate-olap/src/main/resources/application.staging.to.reporting.sql.yml
```diff
@@ -1,6 +1,8 @@
 sql:
   migrate:
+    # No-Op for now
     finalizeList:
+      - NULL
```
malaffoon: And that works?  `NULL` is a valid SQL command in redshift?

ftdholbrook: https://stackoverflow.com/questions/12587433/what-to-use-as-a-noop-in-postgresql
I tried it from the console and it worked

malaffoon: Cool.

PR #665, opened 2020-05-29
cleanup documentation; remove unused FtpReportSender

malaffoon:
I'm trying to "tighten up" the documentation and make it consistent between projects. There will be a corresponding PR for RDW_Reporting.


CONTRIBUTING.md
```diff
@@ -85,42 +137,3 @@ specified to avoid conflict.
 java -jar import-service/build/libs/rdw-ingest-import-service-1.1.0-SNAPSHOT.jar
 java -jar exam-processor/build/libs/rdw-ingest-exam-processor-1.1.0-SNAPSHOT.jar --server.port=8082
 ```
-
-#### FTP Server
```
malaffoon, line 89: This feature isn't in the system any more.

README.md
```diff
@@ -109,14 +101,46 @@ smarterbalanced/rdw-ingest-exam-processor        latest              cf83654e781
...
 Please refer to the [RDW Schema](https://github.com/SmarterApp/RDW_Schema) project for that.
 
 The OLAP applications require remote AWS data stores being configured properly. Care should be taken because these
 remote resources are often shared; also misconfiguration could result in actions being taken on the wrong database.
 
+**NOTE** - The configuration server requires a backing git repository with proper configuration files,
+including configuration settings for the `local-docker` profile. Without that, the system will not
+run properly. (It is beyond the scope of this document to describe how to set that up.)   
+
+**TODO** - change the project so it can run without a configuration server; see TIMS
```
vgonzalez9, line 142: Should we add this todo to the top level RDW? Or should each sub project have that TODO added?

malaffoon: I think it would be better to keep it separate: making ingest work that way will be a lot easier than reporting. But your sentiment is right: this should be a high level Smarter Balanced philosophy so all the projects are consistent at some point.

PR #671, opened 2020-06-12
add max-score to subject trait; simplify exam trait score

malaffoon:


malaffoon:
I'm going to commit this so it can get thru the RST build (crossing fingers). I'll address any questions and concerns.

malaffoon:
In theory, this provides flexibility for whatever they do. In practice, all the purpose/category metadata for ELA are consistent across purposes. In your example ORG is always max points 4 for all five purposes.  

I think, in the UI, we should allow for each purpose having different max points for each category. That is, set the metadata for the table for each purpose separately. Again, in practice, they'll all have the same values and will end up looking the same but, in theory, they don't have to. I don't think that is a big lift but, if it is, let me know and we can discuss.

PR #673, opened 2020-07-14
cleanup slf4j bindings; include filename for parsing error

malaffoon:


validator/build.gradle
```diff
@@ -9,16 +9,11 @@ import org.gradle.internal.jvm.Jvm
...
+// create an executable jar
 jar {
     manifest {
         attributes 'Main-Class': 'org.opentestsystem.rdw.ingest.validator.ValidatorApplication'
     }
-
-    from {
-        configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }
-        configurations.runtime.collect { it.isDirectory() ? it : zipTree(it) }
-    }
```
malaffoon, line 21: i was following conversion instructions (from gradle 4 to 6) but this isn't needed, it builds the jar fine without it. With it, it adds two copies of all the runtime libraries, including slf4j, which was causing warnings to be emitted.

validator/src/main/java/org/opentestsystem/rdw/ingest/validator/service/DefaultValidator.java
```diff
@@ -43,7 +43,7 @@ public void validate(final MessageSink messageSink, final List<String> subjectFi
             } catch (final FileNotFoundException e) {
                 messageSink.error("file not found: {0}", file);
             } catch (final IOException | IllegalArgumentException e) {
-                messageSink.error("problem reading file: " + e.getMessage());
+                messageSink.error("problem reading file {0}: {1}", file, e.getMessage());
```
malaffoon, line 46: It wasn't including the filename in the error message. Since it can be processing multiple files, that was an oversight.

PR #675, opened 2020-08-19
validate trait scores

malaffoon:


exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/Assessment.java
```diff
@@ -17,9 +17,9 @@
     private int subjectId;
     private String subjectCode;
     private List<AssessmentScore> scores;
+    private List<AssessmentTrait> traits;
     private Map<String, Integer> targetCodeToId;
     private Map<String, Integer> items;
-    private Map<String, Integer> traits;
```
malaffoon, line 22: Need more info than just the trait code and id.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/model/ExamTrait.java
```diff
@@ -6,7 +6,7 @@
 public class ExamTrait {
     private Long id;
     private int traitId;
-    private int score;
+    private Integer score;
```
malaffoon, line 9: The way parsing and validating works, we have to allow a null value to be set, even though that is logically invalid.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -1,54 +1,55 @@
 package org.opentestsystem.rdw.ingest.processor.service.impl;
 
 import com.google.common.collect.ImmutableList;
-import org.opentestsystem.rdw.ingest.processor.model.ExamTrait;
```
malaffoon, line 4: Hmm, i did "optimize imports" but import settings must be different than the old days.

exam-processor/src/main/java/org/opentestsystem/rdw/ingest/processor/service/impl/DefaultStudentExamProcessor.java
```diff
@@ -313,4 +314,17 @@ private Integer toValidCompleteness(final String value) {
 
         throw new IllegalArgumentException("invalid value, must be a positive number");
     };
+
+    private static Function<String, Integer> traitScoreChecker(final Integer maxScore) {
```
malaffoon, line 318: The point of the PR: trait scores must be integers between 0 and the max score. If the max score isn't defined, then  just check that it is a non-negative integer.

gregcharles: 👍

PR #680, opened 2020-10-02
Hotfix/2.3.1 

jchen1005:
this fix is to combine the validation error response for exam and student trt attribute. what's happening now is that we do the exam validation first, and ignore the student attribute validation if there is an exam validation failure. smarter balanced wants to combine the validation so that the user will know the exact attribute that failed the validation instead of the partial error response.


gregcharles (approved):
Looks good!

PR #709, opened 2021-07-22
SRS-429: Change to format of input file and other tweaks.

gregcharles:


exam-processor/src/test/java/org/opentestsystem/rdw/ingest/processor/multithreaded/MultiThreadedStudentExamWriter.java
```diff
@@ -96,9 +97,10 @@ protected void before() {
                         .build());
     }
 
+    // Note: synchronized because Java randoms are not thread-safe.
     @Override
-    protected void work() {
-        examWriter.upsert(student, examBuilder.oppId(new Float(Math.random()).toString()).build(), 1);
+    protected synchronized void work() {
+        examWriter.upsert(student, examBuilder.oppId(UUID.randomUUID().toString()).build(), 1);
```
norway77, line 103: Greg can you go over this.

gregcharles: Sure! The basic problem is that this is part of a multi-threaded test, and the Java random generation libraries are not thread-safe. That is, if you ask for a random number simultaneously in two or more threads, you probably will get duplicates. Before this rev, the duplicates were just ignored, which made the tests less effective, but didn't make them fail. Now duplicates cause an exception, which, in the real application, is handled by the exam processor in such a way as to allow the recon job to include them in its report. However, in this test, the exception is interpreted as a test failure. Synchronizing the method fixes that problem.

The change from a random float to a UUID was just for style. Generating random unique identifiers is why UUIDs exist in the first place. Now, they are supposed to be globally unique (for every application in the whole world), and we only need uniqueness within this single test, so it's overkill. Still, I think using a UUID makes it clearer what the intention is.

PR #713, opened 2022-02-15
Fixed the import_id error in the update-inferred-school query for gro…

xuping-xu:
Fixed a wrong import_id parameter in the update-inferred-school query for group-processor.


group-processor/src/main/resources/application.sql.yml
```diff
@@ -59,7 +59,7 @@ sql:
             UPDATE student st
              JOIN upload_student_group sgl ON st.ssid = sgl.student_ssid
              JOIN import imp on imp.id = :import_id
-            SET st.inferred_school_id = sgl.school_id, st.update_import_id = 17410,
+            SET st.inferred_school_id = sgl.school_id, st.update_import_id = :import_id,
```
malaffoon, line 62: Okay, ignore the comment in the other commit

